PR #39, opened 2013-10-22
PDI-10389 Text File Output Step throws a null pointer error when "Do not...

stepanovdg:
... create file at start" is checked, "Add Ending line of file" is valued, and transformation does not pass any rows to the file input step.

 Reopened. Fixed regression`s  situation. Added unit test.


lucboudreau:
http://10.100.11.222:8080/job/wingman-ant-cmd/8/

PR #208, opened 2014-01-17
PDI-10749 - Table Output step ignores Time zone for DATE datatype

dbogdanovich:
New system variable KETTLE_COMPATIBILITY_DB_IGNORE_TIMEZONE has been added
for backward compatibility.


rajagopalnaidu:
I set KETTLE_COMPATIBILITY_DB_IGNORE_TIMEZONE to N but still getting unexpected results. I am reading from MongoDB input whose timezone is UTC from PDI(time zone is CET). When I write the date/time fields they were converted to CET and expected to have them in UTC.

PR #266, opened 2014-02-03
[PDI-8019] Updated ElasicSeach support for version 0.90.10 (curremt stable)

nitram509:
Hi,

I've updated the ElasticSearch APIs to current stable version 0.90.10.
Additionally, I've updated ElasticSearchBulk.java so that it handles data types (String, Date, BigNumber,Number and Boolean) correctly.
This is important, because only upgrading the API versions leads to wrong data,
when the bulk insert runs.

Regards
Martin


pamval:
Hi. thanks for your contribution and apologies for the very late reply. 

I'm closing this since there are too many merge conflicts. 
However, if you take a look at the code on the master branch, you'll notice that we've  updated Elastic Search to version 1.5.2.
This will probably be available in the next Pentaho release.

PR #293, opened 2014-02-14
PDI-11431 NPE when using "Pass through.." and "Skip erorr.."

daschg:


mbatchelor:
No response nor fix - closing PR.

PR #420, opened 2014-03-27
Performance enhancement

dbahdano:
1) Use quicker StringBuilder instead of StringBuffer for local (and, hence, thread-safe by origin) variables
2) Remove unnecessary synchronized sections from environmentSubstitute()


mbatchelor:
In the absence of test cases and proof that it wouldn't regress PDI-4587, closing this PR.

PR #434, opened 2014-04-01
Fixes for PDI-3813 - reference fields with null values caused an exception

malsmith:
Handle null reference field correctly.


pamval:
Closing this, since the corresponding jira issue was also closed

PR #435, opened 2014-04-01
Fix for pdi-11865 - select editable and updateble fields and reference fields

malsmith:


rpbouman:
Closing since this was closed in Tatsiana's pull request.

PR #438, opened 2014-04-02
PDI-11865 Improved Field match guess algorithm handles SFDC reference fields

malsmith:
Matches longest, best matches first


mbatchelor:
@malsmith - thank-you much for the contribution. I was able to merge it in this afternoon.

PR #439, opened 2014-04-02
PDI-5491 Enable setting Rowset prefetch size in the Table input step

malsmith:
Fixes "runaway" query prefetch when rows are wide or contain blob columns.  But required UI changes that are only localized to us_EN.


mbatchelor:
@malsmith  - any chance you can clean up the checkstyle issues so I can process the pull request? Thanks!

mbatchelor:
Unfortunately, this isn't the way the PM wants this implemented. Instead of only being implemented at the Table Input step, PM wants it implemented as a database configuration option as well. Please see http://jira.pentaho.com/browse/PDI-2458

PR #494, opened 2014-04-28
[PDI-11190]: Do not create file at start is still creating zero byte files

TatsianaKasiankova:
There are fix and unit test.


mattyb149:
This has since been fixed to use Kettle's Const.isEmpty()

PR #510, opened 2014-05-13
PDI-11595 - DatabaseMeta addOptions method has MySQL/Infobright specific code

dbahdano:
Moved the database specific code to particular *DatabaseMeta classes


mbatchelor:
@dbahdano - now is the time we could get in this type of change. Could you re-base and squash please?

mbatchelor:
Closing in favor of #3336

PR #517, opened 2014-05-19
[PDI-11211] If Job Executed by transform Job Executor fails the transformation does not fail

stepanovdg:
From correspondence with Jens was realized that misunderstanding about expected behavior : if child transformation fails - execution result should be placed to specified step/field.


jbleuel:
I don't see an obvious relation between this pull request and PDI-11211.
Propose to create a JIRA case that illustrates the issue with a test case or close it out.

pamval:
Following Jens recommendation and closing this.

PR #654, opened 2014-07-17
Launch Java using exec so that it inherits the PID

chewi:
This makes it much easier to write init scripts for Carte. It also ensures that issuing Ctrl+C actually terminates Java and doesn't leave it running in the background. Jsvc support would be even better but I'll leave that to someone else.


dkincade:
@mbatchelor please review - we should either take this or close it

mbatchelor:
Cannot merge this as the spoon.sh script has evolved since this PR. I've created PDI-15435 (where this should have started really) and included a code sample there.

PR #662, opened 2014-07-28
PDI-12648 Line breaks removed from xml tag values in job/transformation files

falling-down:
This goes forward with normal XML encoding but then translates CRLF, CR, and/or LF back to an actual line break.

Probably not the most elegant solution, but it works.

### DIFF
--- a/core/src/org/pentaho/di/core/xml/XMLHandler.java
+++ b/core/src/org/pentaho/di/core/xml/XMLHandler.java
@@ -768,7 +768,7 @@ public class XMLHandler {
 
     if ( val != null && val.length() > 0 ) {
       value.append( '>' );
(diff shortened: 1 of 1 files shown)


prakapenka:
This is a train. This train carries the the best way of xml line endings format.
__________ . .____________ . ____________ . _____________ . _____| |___ 
.replaceAll() H|H .replaceAll()  H|H  .replaceAll()  H|H  .replaceAll()  H|=H=======`)
-------------oo-^-oo------------oo-^-oo------------oo-^-oo--------------oo-^o-o------o-o=

mbatchelor:
We wouldn't want a string search/replace solution actually. More interested in having the text be CDATA'd.

PR #692, opened 2014-08-07
PDI-11884 Get IMAP emails: unable to load BODYSTRUCTURE for sent emails

MikhailGorbachev:


pamval:
this needs to be rebased and reissued, when applicable

PR #802, opened 2014-10-02
[BACKLOG-838] Fix for repository name and description - Cleanup var name...

pminutillo:
...s to alleviate confusion


pamval:
closing since the corresponding jira case was closed back in 5.2. 

Resubmit using a [CLEANUP] tag if you fell this is relevant.

PR #815, opened 2014-10-16
[PDI-12524]: Fixed UDJC on Java 7 for Mac

mattyb149:


mbatchelor:
@mattyb149 - can you add a test case for this ?

mattyb149:
@mbatchelor the code is insulated by the method; it's more of an integration test, which we don't run on Mac. What to do?

PR #817, opened 2014-10-17
[PDI-13072]: added matrix parameters support in REST

vladimirdolzhenko:
http://jira.pentaho.com/browse/PDI-13072

PR #820, opened 2014-10-21
[PDI-13078]: REST doesn't work w/o any input

vladimirdolzhenko:
http://jira.pentaho.com/browse/PDI-13078

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/rest/Rest.java
+++ b/engine/src/org/pentaho/di/trans/steps/rest/Rest.java
@@ -44,6 +44,7 @@ import org.pentaho.di.core.Const;
 import org.pentaho.di.core.encryption.Encr;
 import org.pentaho.di.core.exception.KettleException;
 import org.pentaho.di.core.row.RowDataUtil;
+import org.pentaho.di.core.row.RowMeta;
 import org.pentaho.di.i18n.BaseMessages;
 import org.pentaho.di.trans.Trans;
 import org.pentaho.di.trans.TransMeta;
@@ -320,7 +321,7 @@ public class Rest extends BaseStep implements StepInterface {
     Object[] r = getRow(); // Get row from input rowset & set row busy!
 
     logBasic(Arrays.toString(r));
-    if ( r == null ) {
+    if ( r == null && !first ) {
       // no more input to be expected...
       setOutputDone();
       return false;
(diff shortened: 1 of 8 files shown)


engine/src/org/pentaho/di/trans/steps/rest/Rest.java
```diff
@@ -302,7 +320,8 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
 
     Object[] r = getRow(); // Get row from input rowset & set row busy!
 
-    if ( r == null ) {
+    logBasic(Arrays.toString(r));
+    if ( r == null && !first ) {
```
mattyb149: Not sure this should be at the Basic level, nor should it be logged if the row is null. If we want to log that there were no rows, we should do that at the Detailed or Debug level.

mattyb149:
It is possible that Pentaho will choose to include the functionality specified in the Jira case above, and choose not to include the extra feature (matrix parameters) that is captured in #817.  For this reason I recommend using a clean branch off master and cherry-picking only the related commits in (not including the matrix parameter feature code)

engine/src/org/pentaho/di/trans/steps/rest/Rest.java
vladimirdolzhenko: it's just common logging among steps

vladimirdolzhenko:
@mattyb149 I agree that it is worth to make a pull request per feature. 
As for matrix parameters I disagree with you - there are some API that have it and it is worth to support it as a part of RESTful API standard.

engine/src/org/pentaho/di/trans/steps/rest/Rest.java
mattyb149: With a large number of rows with a large number of columns, this would cause a large decrease in performance. Logging information for each row should be done at RowLevel (which is DEBUG level in Log4J), or if it really is more important than other row-level logging, then Detailed or Debug would be fine.

If the user needs to log the row data at Basic level, they can follow the REST step with a Write To Log step and add all fields at level Basic.

mattyb149:
I'm not saying the feature isn't worth having, I'm saying the branch in this pull request contains commits that address a different Jira case.  Since both look like they're being reviewed for 6.0, then it will probably end up ok, as we can merge the other one first and then this one will merge cleanly on top of it

rpbouman:
@vladimirdolzhenko, thank you for your effort and sorry to keep you waiting so long. We have considered your requirements and inspected your patch. 

Unfortunately we do not see an opportunity to accept this particular patch. The main reason is that we found a number of scenarios that would break when accepting this code change. I will therefore close this pull request.

However, we do think the requirements are valid and actually apply to many more steps. So, we are currently considering how we can best make this happen in a future release, and in a general way so that it can be applied to other steps as well.

Thanks, and best regards.

PR #826, opened 2014-10-28
[PDI-629] - XML Input Stream Accept Filenames

matthewtckr:
Adds the ability to accept input from a previous step, in order to read multiple XML files.
Added blackbox tests for single- and multi-file transformations.


engine/src/org/pentaho/di/trans/steps/xmlinputstream/XMLInputStream.java
```diff
@@ -142,16 +108,86 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
...
+        new ResultFile( ResultFile.FILE_TYPE_GENERAL, data.fileObject, getTransMeta().getName(), getStepname() );
+      resultFile.setComment( BaseMessages.getString( PKG, "XMLInputStream.Log.ResultFileWasRead" ) );
+      addResultFile( resultFile );
+    }
     return true;
   }
 
+  private void closeFile() throws KettleException {
+    try {
+      if ( data.xmlEventReader != null ) {
```
mbatchelor: 1- I would like to see a catch around each of these
2- Usually, when we get an exception closing files, we simply log it as a warning and move on. There's nothing more to do if you cannot close the file or input stream.

matthewtckr:
Updated the closeFile() method to log separate messages for each object that is being closed, instead of throwing an exception.

PR #841, opened 2014-11-14
Fix several MonetDB Bulk Loader issues

dnedev:
[PDI-10169] MonetDB Bulk Loader Step Ignore Truncate Table checkbox
- Pass unquoted schema and table separately when getting a truncate statement 

[PDI-12014] Incorrect data type used for string columns when creating tables in MonetDB
- Use STRING if length is unknown

[PDI-12278] MonetDB Bulk Loader error with MonetDB Jan 2014-SP2 bugfix release
- Do not make a second call to to data.out.writeLine( "" ) (by Juan Carlos)


pamval:
Two of the three issues here are already solved.

Please rebase and reissue pull for remaining issue (PDI-10169) if still relevant.

Thanks for your contribution!

PR #851, opened 2014-11-22
[PDI-2343] Adding variables used in transformations and jobs

claidiere:
Improvements on the auto-documentation step :
  1 - cleaning the code to be in sync with Pentaho code style
  2 - changing the margin of the page and add comment as the page is LANDSCAPE
  3 - Add the generation date as page footer
  4 - Modify the ui to add a selector to display variables used by the trans / job
  5 - Modify the code to add the variables used by the trans / job inside the output


pamval:
Hi, there. Thanks for contributing.

There are a couple of issues with the current pull request:

1. There are a series of merge commits here. You should rebase your work and ensure we don't push the merge commits.
2. Commit messages should start with the case number ([PDI-2343] - ....) and it's probably better if the non-merge commits here are squashed into one.
3. This is a large change, so try and provide a set of unit tests that exercise it.

thanks!

PR #854, opened 2014-11-24
[PDI-6141] Add Metadata Injection Support to Insert/Update Step

cdeptula:


mbatchelor:
Need to close this PR for two reasons. First, this is the old way of doing Metadata Injection instead of the new way using annotations. Second, there are merge conflicts. I've asked both Matt Tucker and Chris if they want to knock this out using the old method.

PR #862, opened 2014-11-25
[PDI-6141] Add Metadata Injection to Update Step

cdeptula:


mbatchelor:
Closed PR because it's using the old style MI and not the new style.

PR #886, opened 2014-12-23
PDI-13204 - RowMeta indexOfValue method works in O(n) time

prakapenka:
Applied original MC path and additional fixes to this path.
This is ambiguous commits and possibly can bring some regressions.
See commits descriptions and junit tests results.

### DIFF
--- a/core/src/org/pentaho/di/core/row/RowMeta.java
+++ b/core/src/org/pentaho/di/core/row/RowMeta.java
@@ -127,6 +127,8 @@ public class RowMeta implements RowMetaInterface {
   @Override
   public void setValueMetaList( List<ValueMetaInterface> valueMetaList ) {
     this.valueMetaList = valueMetaList;
+    // we have new list and we need re-hash
+    indexValueMap();
   }
 
   /**
@@ -233,6 +235,8 @@ public class RowMeta implements RowMetaInterface {
     if ( valueMeta != null ) {
       valueMetaList.set( index, valueMeta );
     }
+    // array list got shifted - so we can't avoid re-hash!
+    indexValueMap();
   }
 
   /**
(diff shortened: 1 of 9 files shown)


mbradbury2014:
@mattyb149 please review...thanks.

dkincade:
@mattyb149 where are we on this?

mattyb149:
I asked for an explanation of the commits, as there are "optional fixes", "additional fixes", etc. Not sure if this just needs to be squashed or if I'm supposed to cherry-pick some commits, try it out, then cherry-pick more, etc.

prakapenka:
It is not squashed to have ability to track all changes.

1 commit - as-is tested patch provided in attachment to ESR - in case of implement it as-is this will be first commit cherry-pick only (integration test broken, original fix, original junits, one step broken)

2 junits that shows what is wrong with original ESR path (since junits for provided with patch originally were not provide real coverege) (integration test broken, original fix, extended junits, one step broken)

3 fixes most issues with original ESR patch (integration test broken, extended fix, extended junits, one step broken)

4 according to architectural issues with original ESR patch we have to fix some existed steps to work properly. This is a fix to one particular place. This issue found with integration tests. (integration test broken, extended fix, extended junits, steps ok)

5 fix to integration tests (not null RowMeta) (integration test ok, extended fix, extended junits, steps ok)

6 fix to prevent external List<RowMeta> modifications. See commit description.  (integration test ok, extended fix, extended junits, steps ok)

As I said before this is arguable fix - so to have ability to inspect all sides of this fix I did not squash commits - so apply them one-by-one and see what is happens is good way to carefully track all possible aspects.

If it were a one big commit from my point of view that would be difficult to track all changes why-and-where causes.

PR #889, opened 2014-12-23
[PDI-13315] Clear out result rows on first iteration of JobEntryJob

mattyb149:
This is only a proposed patch to fix the one case, it should be tested against the other execution paths (per-row, e.g.)

### DIFF
--- a/engine/src/org/pentaho/di/job/entries/job/JobEntryJob.java
+++ b/engine/src/org/pentaho/di/job/entries/job/JobEntryJob.java
@@ -997,9 +997,11 @@ public class JobEntryJob extends JobEntryBase implements Cloneable, JobEntryInte
 
         if ( iteration == 0 ) {
           result.clear();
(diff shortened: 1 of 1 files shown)


ivanpogodin:
@mattyb149

This pull-request actually fixes only not-per-row execution path
The way of fixing "per-row" is not clear because entries JobEntryJob, JobEntryTrans, JobEntryShell have different behavior.

JobEntryJob - bug
  Result rows of JobEntryJob are the result of processing the last input row (duplicated when input rows.size > 1)

JobEntryTrans - looks like correct
  Result rows of JobEntryTrans are the result of processing the last input row.  

JobEntryShell - might be correct but differs
  Result rows are always the same input rows

ivanpogodin:
@mattyb149 see also alternate solution
https://github.com/pentaho/pentaho-kettle/pull/1035

PR #904, opened 2015-01-09
Fix BaseMessages Ambiguous Method call errors

mikebski:
Eclipse was complaining about the BaseMessages.getString method calls being ambiguous.  There were about 125 errors, and creating a method with
no variable arguments fixes the errors.  This was eclipse 4.2.2 and java
1.7.  There was a JDK bug that caused this and an eclipse fix if you want to modify the ecliipse.ini, but this
fixes the code.


pamval:
thanks for the contribution, but I think this was already addressed.

PR #906, opened 2015-01-12
[BACKLOG-1865] - adding images for moving items among a dual list box ui

kurtwalker:


kurtwalker:
@bytekast @pamval please review

PR #913, opened 2015-01-14
[PDI-9289] Setting a null Boolean results in NPE

gretchiemoran:
Minor fix to prevent NPE when setting Boolean values ...


gretchiemoran:
@dkincade @mattyb149 , this is one last minor adjustment, noted after integration test changes.

PR #916, opened 2015-01-14
PDI-12278 monetdb bulk loader error

falling-down:
Apparently a bug fix to monetdb does not require the bulk loader to repeatedly flush the buffer. Doing so was creating a dropped connection error after loading the data.

Tested this fix on Windows.


deinspanjer:
This fix was incorporated into pull request 1146.

PR #918, opened 2015-01-14
PDI-13260 - Excel-Input Step does not save all properties

mbatchelor:


dkincade:
Code reviewed and approved

PR #927, opened 2015-01-15
[BACKLOG-974] Unit test fix for linux vs. windows (order)

mdamour1976:


mdamour1976:
@e-cuellar this test was working on windows due to order of operations and failing on linux.  I have tested on windows and linux now.

PR #928, opened 2015-01-15
[PDI-13369] - Quoting of column names breaks creating of ALTER script

aliakseihaidukou:


aliakseihaidukou:
@dkincade @mattyb149 please review
this is needed to fix CsvTransformGeneratorTest in data-access.
This issue is caused by https://github.com/pentaho/pentaho-kettle/pull/886 - RowMeta.valueIndexMap isn't updated when RowMeta.valueMetaList item changes it's name property.

dkincade:
@mattyb149 @brosander @kurtwalker @hudak

brosander:
@mattcasters mind checking this out? I'm not familiar enough with the databasemeta and valuemeta mechanics to feel comfortable pulling this in right before code freeze

core/test-src/org/pentaho/di/core/database/DatabaseMetaTest.java
```diff
@@ -49,4 +53,23 @@ public void testDatabaseAccessTypeCode() throws Exception {
...
+    doReturn( "\"" ).when( databaseMeta ).getEndQuote();
+    final DatabaseInterface databaseInterface = mock( DatabaseInterface.class );
+    doReturn( true ).when( databaseInterface ).isQuoteAllFields();
+    databaseMeta.setDatabaseInterface( databaseInterface );
+
+    final RowMeta fields = new RowMeta();
+    final ValueMeta valueMeta = new ValueMeta( "test" );
+    fields.addValueMeta( valueMeta );
+    databaseMeta.quoteReservedWords( fields );
+    assertEquals( valueMeta, fields.searchValueMeta( valueMeta.getName() ) );
```
mattyb149: We should also assert that the index of the ValueMeta is the same.

aliakseihaidukou:
@mattyb149 updated the test

mattyb149:
Is this likely the only regression due to the changes in #886 or can this happen any place RowMetaInterface.getValueMeta().setName() is called? Perhaps the index map should've mapped by Object ID (hashCode) instead of name?

hudak:
@mattyb149, this is a side effect of [PDI-13204](http://jira.pentaho.com/browse/PDI-13204)
This [RowMeta unit test](https://github.com/pentaho/pentaho-kettle/blob/master/core/test-src/org/pentaho/di/core/row/RowMetaTest.java#L216) suggests that this is the new expected behavior

Alternatively we could reindex when map lookup fails: https://github.com/hudak/pentaho-kettle/commit/a938a5cb3ba4e314c34bf508d4649d1e88ff8d2b but this would degrade performance

mattyb149:
The side effect is what I'm worried about, if we'll need an extra call every time anyone changes the name of a ValueMeta from a RowMeta. Also what happens if you change one's name and another to the first one's original name? Seems like searchValueMeta will behave differently depending on where you'd call it in the sequence of two renames...

aliakseihaidukou:
This can happen in any place where indexOfValue, searchValueMeta, getString( Object[] dataRow, String valueName, String defaultValue ) are called. Index map is mapped by name because we search by name in these methods, other key will break the performance gain.

hudak:
#951

mattyb149:
Closing in favor of #951

PR #935, opened 2015-01-16
[PDI-13379] Password is not secure in new "Named cluster" dialog

mdamour1976:


mdamour1976:
@e-cuellar for review

PR #938, opened 2015-01-16
[PDI-13378] Cannot set variable as port number in new "Named cluster"

e-cuellar:


e-cuellar:
@mdamour1976 Please Sir!

PR #941, opened 2015-01-19
[BACKLOG-1962] Named cluster - values left blank being filled with null after saving configuration

mdamour1976:


mdamour1976:
@e-cuellar for review

PR #947, opened 2015-01-19
[BACKLOG-1965] Creating a NamedCluster with empty name corrupts metastore

mdamour1976:


mdamour1976:
@e-cuellar For review

PR #949, opened 2015-01-19
[BACKLOG-1960] Named cluster drop-down is too narrow for HBase and Pig

mdamour1976:


mdamour1976:
@e-cuellar while we're at it might as well include this too, cherry-picked, for review

PR #957, opened 2015-01-20
[BACKLOG-1793] Send heartbeat information for long running jobs / transf...

pedrofvteixeira:
...ormations

	- 2 new extension point plugin ID's : TransformationHeartbeat and JobHeartbeat
	- Job and Trans classes now trigger a ScheduledExecutor service that calls the heartbeat extension point plugin at periodic intervals
	- the heartbeart ScheduledExecutor is started after the JobStarted / TransStarted and it is shutdown immediately before JobFinish / TransFinish

### DIFF
--- a/core/src/org/pentaho/di/core/Const.java
+++ b/core/src/org/pentaho/di/core/Const.java
@@ -143,6 +143,11 @@ public class Const {
   public static final int SORT_SIZE = 5000;
 
   /**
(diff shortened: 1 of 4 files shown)


pedrofvteixeira:
@rmansoor, @CodeOnCoffee, @tkafalas please review; to be merged *before* https://github.com/pentaho/pdi-monitoring-plugin/pull/29

PR #958, opened 2015-01-20
[BACKLOG-1959] Named Clusters should be Labeled Hadoop Clusters

mdamour1976:


mdamour1976:
@e-cuellar for review

PR #964, opened 2015-01-21
[PDI-13375] - StepMeta clone doesn t clone over step attributesMap

akhayrutdinov:


engine/test-src/org/pentaho/di/trans/step/StepMetaTest.java
```diff
@@ -0,0 +1,71 @@
+package org.pentaho.di.trans.step;
```
mattyb149: Needs a copyright header

akhayrutdinov: @mattyb149, have done it

mattyb149:
I can't see in the provided Unit Test where your changes to replaceMeta() have been tested. Can you explain (or add the appropriate tests)?  Thanks!

akhayrutdinov:
@mattyb149, oups, my fault. Fixed it by instantiating an attributes' map for a test instance. 

replaceMeta() is called from clone(), and the original ticket refers to problems during cloning routine, thus there is two steps in the test:
1. create a copy of a step by calling clone()
2. check whether its content is the same as the original has

PR #965, opened 2015-01-21
[PMD-994] - Add Redshift dialect to Pentaho Metadata

aliakseihaidukou:
add Redshift to kettle-database-types.xml

@mbatchelor @lucboudreau this fixes failing projects that use metadata(like PIR)


mattyb149:
Do we need this as a native plugin,r would it work as a PDI Database plugin? As a rule I'd like to kethe list of native types small if possible, as the default (built-in) list is quite long already.

If this needs to be a native plugin then that's fine, just wondering.

lucboudreau:
Matt, I've merged this one because it is a followup of something previously checked in.

PR #978, opened 2015-01-22
The dead code has been removed.

Maverick-Crank-GRey:
The '#extractAndSetArgument' method is dead and unused. Kill the dead!


mattyb149:
Because it is a public method, it is technically part of our API. We would likely have to deprecate this first, then delete it at a major release (6.0, e.g.).

PR #999, opened 2015-01-30
[PDI-13420]: Fixed bug in DataGrid to allow creation of NULL values

mattyb149:
Also related to the reproduction transformation in the Jira case, a fix for the Formula step was provided to cast null formula results to the correct data type


ivanpogodin:
@mattyb149  could you merge it?
backport SP-1686 is blocked by this issue

PR #1004, opened 2015-02-03
hive2-jdbc isSigned not supported

cwalet:
as hive2-jdbc still dont't support “ResultSetMetaData.isSigned”,calling such method will do nothing but: throw new SQLException("Method not supported");


mattyb149:
We take care of this in our Big Data Plugin by proxying the call to isSigned and returning a default value if the original throws an exception. Having said that, it's probably a good idea to have this in Kettle proper, do you mind writing up a Jira case and adding this pull request to it?  Thanks!

cwalet:
that's a good idea..but i don't have Pentaho jira account yet.BTW,my poor english "not supported" to do so^_^

mbatchelor:
@cwalet - thank-you for your contribution. Closing this PR in favor of #2626 which was re-based to the codeline and doesn't violate checkstyle requirements.

PR #1010, opened 2015-02-04
[BACKLOG-2051]: Implement support SSL for Carte

TatsianaKasiankova:
There are implementation and unit tests.


TatsianaKasiankova:
@deinspanjer  Hi, Daniel
Please take a look this one.

engine/src/org/pentaho/di/www/WebServer.java
```diff
@@ -77,9 +78,11 @@
 
   private String passwordFile;
 
+  private SslConfiguration sslConfig;
+
   public WebServer( LogChannelInterface log, TransformationMap transformationMap, JobMap jobMap,
       SocketRepository socketRepository, List<SlaveServerDetection> detections, String hostname, int port,
-      boolean join, String passwordFile ) throws Exception {
+      boolean join, String passwordFile, SslConfiguration sslConfig ) throws Exception {
```
mattyb149: As a public "API" (constructor), should we keep the original as deprecated and have it call a new constructor with a null SSL config?

pentaho-nbaker: Agree with Matty

pentaho-nbaker:
Aliaksandr_Drebenchuk is going to add back the original constructor in the next commit

PR #1011, opened 2015-02-05
[PDI-13447] - String Operations Reverse String Function

matthewtckr:
Add a calculator function to reverse a string


bkemper-pentaho:
Not sure I'm crazy about adding yet another function to the calculator step.   In some respects it would be a reasonable standalone utility step, or possibly as part of the String Operations.   But it's not clear to me this would make sense where it is...

mattyb149:
@bkemper-pentaho :+1: for String Operations

matthewtckr:
Thanks for the feedback, I've updated the PR to use the String Operations step instead of Calculator.

PR #1020, opened 2015-02-10
Get rid of psql in pgBulkLoader, and minor improvements

marco44:
Per discussion on PDI-12080 (http://jira.pentaho.com/browse/PDI-12080):

The call to psql binary isn't necessary to start a Copy operation in PostgreSQL: the JDBC driver can also do it.
This patch uses this API (CopyManager), and has various small improvements:
- Raise an error when pgBulkLoader doesn't know how to handle a datatype (instead of putting NULL in the output)
- Add handling of timestamp types (they were ignored)


marco44:
I added a correction: I forgot to test that the truncate mode worked ok. And it didn't...

There is still the problem that all sessions are doing the truncate, which is bad. I'll try to solve this in next commit.

mbatchelor:
@marco44 - could you please rebase? 7.0 is the time we could get this in. Thanks!

marco44:
Hi, I rebased. Though it seems quite a few things changed…

There have been some changes in ValueMeta (don't have the time to dig deeper into this, I just did the same conversion as has been done in the original pgbulkloader, to make this compile). I see there have been some other patches, such as:
[PDI-682] - Replace StringBuffer with StringBuilder for performance

As I don't know what it does, I didn't include it. There may be others too...

I also couldn't do the ant build:

[ivy:resolve]           ::::::::::::::::::::::::::::::::::::::::::::::
[ivy:resolve]           :: com.force.api#force-wsc;37.0.3: not found
[ivy:resolve]           :: com.force.api#force-partner-api;37.0.3: not found

It doesn't seem to come from my part though, it compiled correctly. So I didn't know if my code still works...

Sorry, I got interested in this a year and a half ago, but I'm not a Java programmer at all, so I didn't keep up with Kettle's source. I'd rather not meddle with things and make things worse.

marco44:
Closing as evis has taken over

PR #1022, opened 2015-02-11
PDI-13452 - Automatic Documentation Output Step fails when receiving a reference that doesn't have any steps in it

SergeyTravin:
@mbatchelor, @mattcasters could you review it?

It may be related to PDI-13455 - It is possible to save empty jobs and transformations into the repository.


SergeyTravin:
@mbatchelor, @mattcasters could you review it?
Backport for it https://github.com/pentaho/pentaho-kettle/pull/1111

PR #1023, opened 2015-02-11
[PDI-13365] - Spoon.sh: unable to run Spoon outside the spoon home dir

YuryBY:
@rmansoor, would you mind review/merge?

Additionally (beside the main fix): 
Some variables were changed to avoid complication while switching between catalogs.
Now at the end of spoon.sh script we are returning to initial catalog to ensure that user will be in the same directory as before running spoon.sh file.

### DIFF
--- a/assembly/package-res/spoon.sh
+++ b/assembly/package-res/spoon.sh
@@ -24,16 +24,15 @@ export LIBOVERLAY_SCROLLBAR=0
 # Bug in: https://bugs.launchpad.net/ubuntu/+source/unity-gtk-module/+bug/1208019
 export UBUNTU_MENUPROXY=0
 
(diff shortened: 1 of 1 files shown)


rmansoor:
@YuryBY when I try this change, I am getting this error.  
Error: Unable to access jarfile ./Pentaho/design-tools/data-integration/launcher/launcher.jar

Did you see this error on your side. ?

mtseu:
@rmansoor the way you tested this is incorrect.
during the assembly process all 'launcher.jar' are replaced to the real launcher version.
this is what you did not do in your test. you have to run Kettle build in order to test the fix.

rmansoor:
@mtseu you are correct. processing the pull request now. thanks for the clarification

PR #1026, opened 2015-02-12
[BACKLOG-1986] Implement the heart beat notification for long running jo...

pedrofvteixeira:
...bs and transformation

	- hardening of the heartbeat thread shutdown upon Job/Transformation finish
	- each time the heartbeat thread wakes up, it checks if this Job/Trans instance is finished, and shuts itself down if this is so;


pedrofvteixeira:
@pentaho-nbaker, @rmansoor, @tkafalas please review

PR #1029, opened 2015-02-13
[PDI-13210] - Create a map of extension point plugin for later use

rmansoor:
I have tested using a transformation with quite a few step.

### DIFF
--- a/core/src/org/pentaho/di/core/extension/ExtensionPointHandler.java
+++ b/core/src/org/pentaho/di/core/extension/ExtensionPointHandler.java
@@ -22,12 +22,8 @@
 
 package org.pentaho.di.core.extension;
 
-import java.util.List;
-
 import org.pentaho.di.core.exception.KettleException;
 import org.pentaho.di.core.logging.LogChannelInterface;
-import org.pentaho.di.core.plugins.PluginInterface;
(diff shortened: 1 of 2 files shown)


rmansoor:
@mattyb149 , @pentaho-nbaker please review

mattyb149:
Hmm I can't find my comments, I thought I'd posted or emailed them a while ago.

The code looks good, I'm just wondering if we should change the behavior of extension points in order to get the performance increase of having cached instances of the plugins (instead of caching the plugin classes). Having them mapped by ID is a great performance increase already, and having cached instances would be even better, but it would change the current behavior of how extension points are invoked.

@pentaho-nbaker @mattcasters What do you guys think about this? The change in behavior is that all  member variables (static and non-static) will retain their state between invocations of the extension point plugin's callback. Before, a new extension point plugin object would be created and then the callback would be invoked. This allowed non-static members to be used among methods in the class.

I'm thinking this is fine as a change, we would just want to document it. We had never made any claims as to how extension point plugins' member variables behaved before, so I would think we can just choose this to be the "official" way, and mention in the documentation that if you were using non-static member variables in earlier versions, the behavior is "undefined" or something like that.

PR #1035, opened 2015-02-19
[PDI-13315] [v2] Subjobs called from jobs are duplicating result rows

ivanpogodin:
@mattyb149 review it please.
This is suggested instead of https://github.com/pentaho/pentaho-kettle/pull/889
The behavior implemented is like in JobEntryTrans when execute transformation on the local machine:
  if subjob result contains rows, they are used as the JobEntryJob result rows
  if subjob result contains no rows, JobEntryJob result rows stay unchanged


pentahofan:
I hope this fixes this bug :)

ivanpogodin:
@mattyb149 
see also https://github.com/pentaho/pentaho-kettle/pull/1047 (SP-1680 - backport to 5.3)

PR #1036, opened 2015-02-19
BACKLOG-2041 -- Create encrypted data transfer flow

kolinus:
What was done:
1) added initial functionality: key generation, session key encryption \
sharing
2) added rmote step encryption support
3) added unit tests for encoding \ decoding keys and random keys
generation

### DIFF
--- a/core/src/org/pentaho/di/core/encryption/CertificateGenEncryptUtil.java
+++ b/core/src/org/pentaho/di/core/encryption/CertificateGenEncryptUtil.java
@@ -22,17 +22,21 @@
 
 package org.pentaho.di.core.encryption;
 
+import java.security.InvalidKeyException;
 import java.security.Key;
 import java.security.KeyFactory;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
 import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
 import java.security.spec.KeySpec;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 
 import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
 import javax.crypto.KeyGenerator;
+import javax.crypto.NoSuchPaddingException;
 import javax.crypto.spec.SecretKeySpec;
 
 import org.pentaho.di.core.logging.LogChannel;
@@ -43,8 +47,8 @@ import org.pentaho.di.core.logging.SimpleLoggingObject;
 public class CertificateGenEncryptUtil {
 
   public static final int KEY_SIZE = 1024;
-  public static final String PUBLIC_KEY_ALGORYTHM = "RSA";
-  public static final String SINGLE_KEY_ALGORYTHM = "AES";
+  public static final String PUBLIC_KEY_ALGORITHM = "RSA";
+  public static final String SINGLE_KEY_ALGORITHM = "AES";
   public static final String TRANSMISSION_CIPHER_PARAMS = "RSA/ECB/PKCS1Padding";
   private static final LoggingObjectInterface loggingObject = new SimpleLoggingObject(
       "Certificate Encryption Utility", LoggingObjectType.GENERAL, null );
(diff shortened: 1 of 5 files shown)


kolinus:
@deinspanjer  Daniel, could you please review. Please note that current;y encryption flag is set to false -- so nothing should be changed in behavior. Also I added the tests proving key encryption \ decryption works

core/src/org/pentaho/di/core/encryption/CertificateGenEncryptUtil.java
```diff
@@ -36,7 +43,9 @@
 public class CertificateGenEncryptUtil {
 
   public static final int KEY_SIZE = 1024;
-  public static final String ALGORYTHM = "RSA";
+  public static final String PUBLIC_KEY_ALGORYTHM = "RSA";
```
deinspanjer: It is a nit-pick, but please change the spelling to algorithm. :)

core/src/org/pentaho/di/core/encryption/CertificateGenEncryptUtil.java
```diff
@@ -53,10 +62,52 @@ public static KeyPair generateKeyPair() {
     return pair;
   }
 
+  public static Key generateSingleKey() throws NoSuchAlgorithmException {
+    Key key = KeyGenerator.getInstance( SINGLE_KEY_ALGORYTHM ).generateKey();
+    return key;
+  }
+
+  public static byte[] encodeKeyForTransmission( Key encodingKey, Key keyToEncode ) throws Exception {
```
deinspanjer: Let's throw the more specific exceptions here, then test for some of the more likely ones in the unit tests.

core/src/org/pentaho/di/core/encryption/CertificateGenEncryptUtil.java
```diff
@@ -53,10 +62,52 @@ public static KeyPair generateKeyPair() {
...
+
+  public static byte[] encodeKeyForTransmission( Key encodingKey, Key keyToEncode ) throws Exception {
+    Cipher cipher = Cipher.getInstance( TRANSMISSION_CIPHER_PARAMS );
+    cipher.init( Cipher.WRAP_MODE, encodingKey );
+    byte[] encodedKey = cipher.wrap( keyToEncode );
+    return encodedKey;
+  }
+
+  public static Key decodeTransmittedKey( byte[] sessionKey, byte[] transmittedKey, boolean privateKey )
+      throws Exception {
```
deinspanjer: Same thing with exceptions.

core/src/org/pentaho/di/core/encryption/CertificateGenEncryptUtil.java
```diff
@@ -53,10 +62,52 @@ public static KeyPair generateKeyPair() {
...
+    if ( unwrappedKey != null ) {
+      decryptionCip.init( Cipher.ENCRYPT_MODE, unwrappedKey );
+    } else {
+      SecretKeySpec sks = new SecretKeySpec( unencryptedKey, SINGLE_KEY_ALGORYTHM );
+      decryptionCip.init( Cipher.ENCRYPT_MODE, sks );
+    }
+    return decryptionCip;
+  }
+
   public static byte[] encryptUsingKey( byte[] data, Key key ) {
```
deinspanjer: We should catch the most likely exceptions and craft some readable error messages for them rather than just relying on call-stacks.  These errors are likely to confuse users and we don't want to frustrate them.

engine/src/org/pentaho/di/trans/TransMeta.java
```diff
@@ -282,6 +282,9 @@
   /** The log channel interface. */
   protected LogChannelInterface log;
 
+  protected byte[] key;
```
deinspanjer: Let's use a more descriptive name in the code and the XML so people know what kind of key they are looking at.

core/src/org/pentaho/di/core/encryption/CertificateGenEncryptUtil.java
kolinus: SO: throw specific Exception or catch them in method and create meaningful logging?

deinspanjer: In the methods that currently throw exception, throw the non-runtime error types explicitly.  In the callers, catch the most likely of those exceptions and log readable messages (then you can rethrow if you want the callstack as well).

kolinus:
@deinspanjer  Daniel, I added code based on you feedback

PR #1038, opened 2015-02-20
[PDI-13456] - Metadata Injector and Get File Names throws ConcurrentModi...

akhayrutdinov:
...ficationException

### DIFF
--- a/engine/src/org/pentaho/di/trans/step/BaseStep.java
+++ b/engine/src/org/pentaho/di/trans/step/BaseStep.java
@@ -39,6 +39,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.pentaho.di.core.BlockingRowSet;
 import org.pentaho.di.core.Const;
@@ -275,7 +276,8 @@ public class BaseStep implements VariableSpace, StepInterface, LoggingObjectInte
    * Map of files that are generated or used by this step. After execution, these can be added to result. The entry to
    * the map is the filename
    */
-  private Map<String, ResultFile> resultFiles;
+  private final Map<String, ResultFile> resultFiles;
+  private final ReentrantReadWriteLock resultFilesLock;
 
   /**
    * This contains the first row received and will be the reference row. We used it to perform extra checking: see if we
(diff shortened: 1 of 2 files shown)


brosander:
@akhayrutdinov Should we be worried about consistency in the map? This may hide the exception but if we've got people iterating over a map that's being actively changed, elements could be added or removed during iteration still.

lucboudreau:
A concurrent hash map does not protect against concurrent modifications. 

This problem happens because someone is making changes to the map while someone else is iterating on it. That's a no-no.

akhayrutdinov:
@brosander, I agree with your concern as CHM only prevents from exception when being modified at the same time with being iterated over (so does @lucboudreau said). However, I see no more feasible solutions.

Basically, each step can be accessed at every moment so far, despite of its state. This may lead to inconsistence, as the data can be retrieved before it is properly prepared. There is no way to guarantee the safe work now.

Just look at how MetaInject works: https://github.com/pentaho/pentaho-kettle/blob/9f1fa3783e37059bc70f0f3350c35dfc8e836a5d/engine/src/org/pentaho/di/trans/steps/metainject/MetaInject.java#L305 
That means it tries to obtain data from a running transformation, executed in separate threads.

My suggestion on how this should be implemented is based on a sort of "safepoints". Assume each step has a special state, called "safepoint", when it can share its current state (in other words - its state is observerable) and it is guaranteed this state is consistent. Unless the step reaches the state, it blocks every thread calling step's data accessors (like this one: https://github.com/pentaho/pentaho-kettle/blob/9f1fa3783e37059bc70f0f3350c35dfc8e836a5d/engine/src/org/pentaho/di/trans/Trans.java#L2583).

But, frankly speaking, I don't believe this approach can be implemented since it has a enormous architectural impact. Therefore I have selected the easiest way to hide the problem - just to handle simultaneous operations over the map.

During the process of execution it can reveal "strange" content snapshot, however at the very end, when no writers are alive, the data would not be corrupted.

Please,  share your thoughts with me

lucboudreau:
If this list is only a cache of things and it is ok for it to go out of sync momentarily (like a list of listeners for example), you should consider a CopyOnWriteArrayList, or some equivalent Map-style object.

brosander:
@lucboudreau the javadoc states that it won't throw a ConcurrentModificationException, we'll have temporary consistency issues as @akhayrutdinov suggested but this should prevent the exception (keySet, values, entrySet state they won't throw)
http://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html#keySet()

lucboudreau:
@mattcasters Any opinion on this?

akhayrutdinov:
@brosander, @lucboudreau, @mattcasters,
any updates on this, guys?

akhayrutdinov:
@brosander, I've updated the PR with lock-based implementation and rebased it from the current master state. Review it please.

/cc @mattcasters, @lucboudreau

PR #1040, opened 2015-02-20
BACKLOG-2041 - Create encrypted data transfer flow

kolinus:
What was done:
1) added dummy code to catch block of unit tests to make codestyle happy
2) removed unnecessary import after refactor


kolinus:
@deinspanjer  Daniel, some cosmetics

PR #1042, opened 2015-02-21
Update JobEntryWaitForSQL.java

MarceloMadruga:
I'm getting an executivo error because of executivo metido which is setting 1 error to the execution even it doesn't fail. I suggest to check the result at the end and remove the error... Or don't even set that error.


mbatchelor:
Thank-you so much for the contribution. I created http://jira.pentaho.com/browse/PDI-15437 in our JIRA system to describe the issue, and then brought your code over in a different PR which follows our checkstyle rules ( #2627 ). So I'm closing this PR, but not because it wasn't useful - it just didn't meet our current checkstyle rules. Thank-you again for your contribution!

PR #1046, opened 2015-02-24
[PDI-12521] - Well Formed XML Dialog Wildcard

matthewtckr:
Fix an issue where the Wildcard field is not properly set when opening the dialog window


mattyb149:
This is actually for http://jira.pentaho.com/browse/PDI-13521

PR #1048, opened 2015-02-25
[BACKLOG-2149] ConcurrentModificationException occurs during PDI PluginR...

tkafalas:
...egistry


mattyb149:
Was this while loop necessary? What about a synchronized block using the map, or something Iike that?

PR #1057, opened 2015-02-26
Google Analytics Input step

gkfabs:
Hello,

Here are 3 changes I am using since a while and are simplificating my life.
I hope they are done the right way.

Thanks


gkfabs:
This pull request is no longer valid

PR #1063, opened 2015-02-27
[BACKLOG-2253]: ExcelWriter: cell background goes completely black when editing a cell that has been formatted

TatsianaKasiankova:
There is a fix that consists of avoiding using CellUtil.setCellStyleProperty() method on POI because of known issue:
https://bz.apache.org/bugzilla/show_bug.cgi?id=56548.


TatsianaKasiankova:
@deinspanjer 
Hi, Please review the fix.

PR #1066, opened 2015-03-02
[PDI-13539]: Fixed the setting of origin step in Merge Join

mattyb149:
I deprecated one API method in RowMetaInterface and added a new one (delegating the old to the new), but since this is for 6.0 I figured it was ok.

This change allows the origin step for fields to be retained even if in info streams, and if they are renamed then the origin step is set as the one that is passed in. For this Jira case, I only updated Merge Join. In the future we can change other steps (Join Rows, e.g.) to use the new API method and pass its own name in as the origin.

@mattcasters @hudak @rfellows Do you mind reviewing this? Thanks!

P.S. Sorry about the inline Checkstyle changes, I have a habit of typing quickly then hitting the AutoFormat key combo to get it pretty...

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/mergejoin/MergeJoinMeta.java
+++ b/engine/src/org/pentaho/di/trans/steps/mergejoin/MergeJoinMeta.java
@@ -32,6 +32,7 @@ import org.pentaho.di.core.exception.KettleException;
 import org.pentaho.di.core.exception.KettleStepException;
 import org.pentaho.di.core.exception.KettleXMLException;
 import org.pentaho.di.core.row.RowMetaInterface;
+import org.pentaho.di.core.row.ValueMetaInterface;
 import org.pentaho.di.core.variables.VariableSpace;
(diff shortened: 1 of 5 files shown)


engine/src/org/pentaho/di/trans/steps/mergejoin/MergeJoinMeta.java
```diff
@@ -268,26 +266,29 @@ public void check( List<CheckResultInterface> remarks, TransMeta transMeta, Step
...
         }
       }
     }
 
     for ( int i = 0; i < r.size(); i++ ) {
-      r.getValueMeta( i ).setOrigin( name );
+      ValueMetaInterface vmi = r.getValueMeta( i );
+      if ( vmi != null && Const.isEmpty( vmi.getName() ) ) {
+        vmi.setOrigin( name );
+      }
```
mattyb149: This is the actual change, propagated up through the interface, etc.

PR #1067, opened 2015-03-03
[PDI-13508] NPE in DimensionCache class after update to Java 1.7u76

ivanpogodin:
@mattyb149 review it please.

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/dimensionlookup/DimensionCache.java
+++ b/engine/src/org/pentaho/di/trans/steps/dimensionlookup/DimensionCache.java
@@ -213,21 +213,29 @@ public class DimensionCache implements Comparator<Object[]> {
       Date toDate = toDateMeta.getDate( o1[toDateIndex] );
       Date lookupDate = fromDateMeta.getDate( o2[fromDateIndex] );
 
-      if ( toDate != null ) {
-        int cmpTo = lookupDate.compareTo( toDate );
-
-        if ( fromDate == null && cmpTo < 0 ) {
-          return 0; // match!
+      int fromCmpLookup = 0;
+      if ( fromDate == null ) {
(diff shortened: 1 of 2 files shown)


ivanpogodin:
@brosander could you review it?

ivanpogodin:
@mattyb149 @brosander @deinspanjer  could you review it?

deinspanjer:
This refactoring is good, and I don't mind merging it, but I wonder if we might have a deeper problem.

Were you able to reproduce the error as originally described?

The bug mentioned that setting useLegacyMergeSort to true bypassed the problem.  Do we know why exactly that fixes the problem?  I'm wondering if it might be something related to http://bugs.java.com/bugdatabase/view_bug.do?bug_id=5045147 which means we might be putting inappropriate objects into the collection.

Though, if that were the case, I wouldn't necessarily expect this fix that just improves null checking on the compare to prevent the error.

ivanpogodin:
@deinspanjer 
>> Were you able to reproduce the error as originally described?
Yes
>>  it might be something related to http://bugs.java.com/bugdatabase/view_bug.do?bug_id=5045147
This is about TreeMap without Comparator. 
DimensionCache situation is sorting List with Comparator and nothing more.

The bad news is: the Comparator implementation is not symmetric.
It's possible: ( cmp.compare(a, b) == 0 && cmp.compare(b, a) > 0 )
when a["DATE_FROM"] < b["DATE_FROM"] < a["DATE_TO"].
It's a special case. I didn't try to change it.

ivanpogodin:
@deinspanjer
The error occured when 2nd argument contained interval [null, null] (test line 42).
It didn't occur when the argument was the 1st (test line 35).
The order depends on the sorting algorithm that can differ in Java versions.
Hence it's not the case of checking appropriate value.

PR #1049, opened 2015-02-26
[SP-1697] - Regression PDI-13435 - Mapping (sub-transformation) step ret...

akhayrutdinov:
...urns an error (5.3 Suite)

@brosander. @mattyb149 review it please.
It is a backport of https://github.com/pentaho/pentaho-kettle/pull/1027. Since SP sprint ends on Friday, could you please take both PRs into the account?

PR #1068, opened 2015-03-03
[Pdi 13435] - Mapping (sub-transformation) step returns an error

akhayrutdinov:
@mattcasters, could you please take this into the account and merge?
/cc @mattyb149


engine/src/org/pentaho/di/trans/steps/mappinginput/MappingInput.java
```diff
@@ -165,21 +176,6 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
     return true;
   }
 
-  private void replaceRenameIfNecessary( RowMetaInterface rowMeta, MappingValueRename valueRename )
```
deinspanjer: Why are you removing this extracted function?

akhayrutdinov: @deinspanjer, 
It is a revert commit. I created this function in previous PR (that doesn't fix the problem) - https://github.com/pentaho/pentaho-kettle/pull/1027.

deinspanjer: Sorry, I didn't notice there was a revert and a new commit mixed into the same PR.
It still seems like it would be worth removing the duplicated code here, but I notice that there is a very slight change to the extracted method's code where it does an additional check, so reverting first is probably the right thing.

Okay, this looks good to me from a review standpoint.  I'll run through the tests in just a bit.

deinspanjer:
Alright, this looks good to me.  @mattcasters would you like me to merge?

akhayrutdinov:
Please merge the corresponding PR for SP ticket - https://github.com/pentaho/pentaho-kettle/pull/1049
It is the same

PR #1069, opened 2015-03-03
[PDI-13510] - Database Logging does not log Jobs when Executing Remotely

aliakseihaidukou:
initialize kettle.properties when load job from XML

@mbatchelor please review


aliakseihaidukou:
@mattyb149 maybe you can have a look

aliakseihaidukou:
@mattcasters @brosander looks like you can look at this too

PR #1072, opened 2015-03-04
PDI-8746 -- Unable to use Windows UNC path in Property Input Step

kolinus:
What was done:
1) changed the way we load properties from file -- removed intermediate
File usage and switched to direct InputStream use
2) no unit test is applied as we need a network share and heavy
refactoring of the code
3) It was manually tested on windows with the following scenarios: win ini
file both local and share, properties files both share and local

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/propertyinput/PropertyInput.java
+++ b/engine/src/org/pentaho/di/trans/steps/propertyinput/PropertyInput.java
@@ -22,8 +22,7 @@
 
 package org.pentaho.di.trans.steps.propertyinput;
 
-import java.io.File;
-import java.io.FileInputStream;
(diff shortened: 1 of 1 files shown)


kolinus:
@mattyb149  @deinspanjer  @mattcasters  could you please review

deinspanjer:
The code looks okay, but I'm wondering if there is any reasonable way to add a unit test.
One idea I had was if the tests are run on windows with an administrative account, we could read in win.ini from the default admin unc path.
A way to manually test this would be to open a command prompt and run the following command:
type \\localhost\admin$\win.ini

kolinus:
we (both EPAM and Pentaho) have linux on CI and not sure if we can hardcode any windows computer name

PR #1074, opened 2015-03-05
Update KettleDatabaseRepositoryDatabaseDelegate.java

gaojd:
issue:it can not save other database attributes when modify or add database connect.The reason is that PreparedStatement is diffirent when add sql batch.


matthewtckr:
Hi, this appears to be fixed via #1131, for http://jira.pentaho.com/browse/PDI-13502.

pamval:
Closingm as this was fixed in another commit.

PR #1075, opened 2015-03-05
PDI-10088 - UNC path does not work in Get mails(POP3/IMAP) step

kolinus:
What was done:
1) changed java.io.File to KettleVFS
2) no unit tests available as we need network share for that
3) manually tested both network and local folder


kolinus:
@deinspanjer  Daniel,
Could you please take a look.

PR #1079, opened 2015-03-08
repair module:engine compile error . because miss dependency "libformula"

Lihuanghe:
module:engine compile error .
add dependency "libformula" to file  ivy.xml.


pamval:
this has been resolved in the meantime. thanks!

PR #1082, opened 2015-03-09
[PDI-11388] - FTPx settings not retested in Dialog

matthewtckr:
In several FTP, SFTP, and FTPS job entries, when clicking the Test Connection button, the job entry does not retest the connection after the first test is completed.

This fixes false-positive connections resulting from the test.

### DIFF
--- a/ui/src/org/pentaho/di/ui/job/entries/ftpdelete/JobEntryFTPDeleteDialog.java
+++ b/ui/src/org/pentaho/di/ui/job/entries/ftpdelete/JobEntryFTPDeleteDialog.java
@@ -1260,6 +1260,15 @@ public class JobEntryFTPDeleteDialog extends JobEntryDialog implements JobEntryD
       }
       retval = true;
     } catch ( Exception e ) {
+      if ( ftpclient != null ) {
+        try {
+          ftpclient.quit();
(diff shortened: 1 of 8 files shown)


ui/src/org/pentaho/di/ui/job/entries/ftpdelete/JobEntryFTPDeleteDialog.java
```diff
@@ -1307,6 +1307,10 @@ private boolean connectToFTPS() {
       }
       retval = true;
     } catch ( Exception e ) {
+      if ( ftpsclient != null ) {
+        ftpsclient.disconnect();
```
mbatchelor: I believe this should be:

if ( ftpsclient != null ) {
  try {
    ftpsclient.disconnect();
  } catch ( Exception ignored ) {
    // Cannot do anything else here
  }
  ftpsclient = null;

matthewtckr:
Updated the job entries to catch any potential exceptions that might be thrown when the connection used to test the settings is closed/disconnected.  This is OK as the connection is only used by the user to test their connection settings, and the connection is not reused for the actual execution of the job entry.

kolinus:
@deinspanjer  is that ok for you ?

PR #1094, opened 2015-03-11
[PDI-13584] Group By step not showing Percentile

matthewtckr:
Fix GroupByMeta to properly include Percentile aggregates when getting list of output fields for subsequent steps

### DIFF
--- a/engine/test-src/org/pentaho/di/trans/steps/groupby/GroupByTest.java
+++ b/engine/test-src/org/pentaho/di/trans/steps/groupby/GroupByTest.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2013 by Pentaho : http://www.pentaho.com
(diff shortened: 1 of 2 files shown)


engine/test-src/org/pentaho/di/trans/steps/groupby/GroupByTest.java
```diff
@@ -98,4 +104,27 @@ public void testProcessRow() throws KettleException {
 
     assertTrue( groupBySpy.getOutputRowSets().get( 0 ).isDone() );
   }
+
+  @Test
+  public void testGetFields() {
```
matthewtckr: Please also test the TYPE_GROUP_STANDARD_DEVIATION and TYPE_GROUP_MEDIAN, from the GroupByMeta getFields() method.

matthewtckr:
Updated test case to include the two types listed earlier, as well as other aggregation types that always return a specific type.

Some other aggregations not in this test return a type that depends upon the incoming field type.

PR #1102, opened 2015-03-13
[PDI-1495] Implment WebServerShutdownHandler to allow unit tests to disa...

tkafalas:
...ble system.exit on server shutdown


pentaho-nbaker:
Looks good. Just add some javadoc on the interface at least. Thanks!

PR #1112, opened 2015-03-17
[PDI-13570] - Regression: Shared DB-connections without .kdb file not sh...

akhayrutdinov:
...owing in Spoon 5.3

### DIFF
--- a/ui/src/org/pentaho/di/ui/spoon/DatabasesCollector.java
+++ b/ui/src/org/pentaho/di/ui/spoon/DatabasesCollector.java
@@ -1,0 +1,88 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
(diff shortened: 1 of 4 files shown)


mattyb149:
The class and methods should be public not package-protected

akhayrutdinov:
@mattyb149 why do you think so? The collector is a class to be used inside a procedure, I don't really believe it makes any sense to reveal it to the whole world

mattyb149:
Our coding standards discourage package protection in favor of explicit protected access. This is usually to make the access level clear. If package-protected is the best choice, then the class and each method must be documented as such (Javadoc)

mattyb149:
I recommended public access because it appears generally useful. Even if we only call it once from the same package, other people who embed PDI would be able to use it as they see fit :)

akhayrutdinov:
@mattyb149, I've updated the PR

mattyb149:
Cool, thanks!

PR #1123, opened 2015-03-23
BACKLOG-2437 - Index DDL generation do not prepend the schema name in th...

apratkin:
...e table name


apratkin:
It was tested with following steps:
1.	Open kettle
2.	Create new transformation
3.	Add Data Grid
4.	Add combination lookup/update and edit it
5.	Create new connection
6.	At advanced tab set preferred schema name
7.	Select table
8.	Populate key fields
9.	Press SQL button
10.	Verify all sql statement contains schema name
Perform the same steps for Dimension lookup/update step and for Update step

apratkin:
@diogofscmariano, could you take a look?

diogofscmariano:
@apratkin unit tests please

apratkin:
@diogofscmariano, I have added JUnits as separate commit

PR #1129, opened 2015-03-23
[BACKLOG-2176]

diogofscmariano:
[BACKLOG-2176]:
- Extended dialect to support:
 - different sequence nomaxvalue
 - no support for auto generated keys being returned by the prepared statement
 - customization of the value meta interface from the sql type
- Changed CombinationLookup, DimensionLookup and TableOutput steps to handle the new dialect changes
- Kept default behaviour for every new dialect extension
- Added unit tests for the dialect and steps changes


diogofscmariano:
@wgorman @mattcasters @pamval please review

diogofscmariano:
@mattyb149 can you also please take a look?

engine/src/org/pentaho/di/trans/steps/tableoutput/TableOutput.java
```diff
@@ -303,7 +303,7 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
       }
 
       // See if we need to get back the keys as well...
-      if ( meta.isReturningGeneratedKeys() ) {
+      if ( meta.isReturningGeneratedKeys() && meta.getDatabaseMeta().supportsAutoGeneratedKeys() ) {
```
deinspanjer: For table output here, is it possible to return generated keys without supporting autogenerated keys?  Could we maybe put the check for whether the dialect supports it into the meta and incorporate that into the isReturningGeneratedKeys() call?

diogofscmariano:
@deinspanjer can you please check?

PR #1131, opened 2015-03-24
[PDI-13502] Regression: Instance Name on Database conn SQL server not saved to repository

SergeyTravin:


SergeyTravin:
@mattcasters, @brosander, @mattyb149 could you review it?
Backport for 5.3 https://github.com/pentaho/pentaho-kettle/pull/1132

mattyb149:
Can you explain the fix? Are rows being overwritten? It appears to be only storing one attribute, will other attributes be lost?

SergeyTravin:
We should use Database.prepareInsert(RowMetaInterface rowMeta, String tableName) only once because it creates new prepared for each call. 
Next attributes are added to batch by Database.setValuesInsert(RowMetaAndData row) and are commited to database when we call Database.insertRow(PreparedStatement ps, boolean batch, boolean handleCommit) and batch size is reached commit size.
Finally, we call Database.executeAndClearBatch(PreparedStatement preparedStatement). It commits all uncommited values and closes batch. 

We should use batch for performance, please see 
http://jira.pentaho.com/browse/BACKLOG-174

Unfortunately, we do not save attributes now, e.g. http://jira.pentaho.com/browse/PDI-13093

I hope this clarifies the situation.

bkemper-pentaho:
This pull request (and the comments) don't seem to relate at all to the case in question.   According to the header, this is a classic case where what we're saving to the repository doesn't match the metadata of the class.   This appers as if the SQL Server dialog has an extra field for the instance name, and it's being persisted correctly to files, but not to the repository.    The fix listed here has something to do with job execution - not the connection information (unless I'm totally missing something)

bkemper-pentaho:
So - somewhere along the line the prepareStatement being recreated was what was causing the issue with the field not saving?
Also - would have been better to have the test reflect the actual failing case (MS SQL SERVER with extra properties) rather than just H2 which hadn't been reported to be failing.

SergeyTravin:
@bkemper-pentaho, Instance Name field for MS SQL SERVER saves as properties but actually only last property is saved to repository. 
The commit allows to save all field for database connection in database repository. 

It it does not matter which of database connection we use.  
We just should compare loaded from repository  DatabaseMeta  with initial DatabaseMeta.

SergeyTravin:
@mattyb149, @bkemper-pentaho Could you provide any feedback?

PR #1134, opened 2015-03-24
PDI-13367 - ETL Metada Injection does not actualize the Source Step in the Inject Metadata tab when renaming the cited step

yuryilyukevich:


yuryilyukevich:
@mattyb149 Please, look it.

engine/src/org/pentaho/di/trans/step/StepMetaChangeListenerInterface.java
```diff
@@ -0,0 +1,12 @@
+package org.pentaho.di.trans.step;
```
mattyb149: Needs a copyright header

engine/src/org/pentaho/di/trans/step/StepMetaChangeListenerInterface.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.pentaho.di.trans.TransMeta;
+
+public interface StepMetaChangeListenerInterface {
+  /**
+   * 
+   *
```
mattyb149: Please add javadoc here

yuryilyukevich:
@mattyb149 The fix was updated.

kurtwalker:
@mattyb149 @yuryilyukevich Looks like this change has caused an error in all spoon dialogs.  Open a step dialog in any transformation, then click ok.  You'll get an exception complaining about notify listeners.

PR #1146, opened 2015-03-26
[PDI-12278]: MonetDB Bulk Loader error with MonetDB Jan 2014-SP2 bugfix  release

TatsianaKasiankova:
The problem was:
Starting from Jan 2014-SP2 bugfix release we should not use double write of empty line (see the sample http://dev.monetdb.org/hg/MonetDB/file/tip/java/example/SQLcopyinto.java).
At the same time for "old" versions of MonetDB (before Jan 2014-SP2) we are still need to do this one otherwise the error appears.

The fix consists of:
1)Getting current MonetDB version (via JDBC property) during execution the step
2)Defining if the current version is before boundary Jan 2014-SP2 version or not (so-called compatibility with old version db mode).
3)Addition condition for such case during writing buffer.

Also there are several unit tests.

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/monetdbbulkloader/MonetDBBulkLoader.java
+++ b/engine/src/org/pentaho/di/trans/steps/monetdbbulkloader/MonetDBBulkLoader.java
@@ -157,6 +157,7 @@ public class MonetDBBulkLoader extends BaseStep implements StepInterface {
           db.disconnect();
         }
       }
+      meta.setCompatibilityDbVersionMode();
     } catch ( Exception ex ) {
       throw new KettleException( ex );
     }
@@ -502,12 +503,16 @@ public class MonetDBBulkLoader extends BaseStep implements StepInterface {
       if ( error != null ) {
         throw new KettleException( "Error loading data: " + error );
       }
-      data.out.writeLine( "" );
 
-      // and again, making sure we commit all the records
-      error = data.in.waitForPrompt();
-      if ( error != null ) {
-        throw new KettleException( "Error loading data: " + error );
+      if ( !meta.isCompatibilityDbVersionMode() ) {
+        // write an empty line, forces the flush of the stream
+        data.out.writeLine( "" );
+
+        // again...
+        error = data.in.waitForPrompt();
+        if ( error != null ) {
+          throw new KettleException( "Error loading data: " + error );
+        }
       }
 
       if ( log.isRowLevel() ) {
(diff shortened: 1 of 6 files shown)


TatsianaKasiankova:
@deinspanjer Hi, could you please review this one.

engine/src/org/pentaho/di/trans/steps/monetdbbulkloader/MonetDBBulkLoader.java
```diff
@@ -635,6 +642,7 @@ protected static MapiSocket getMonetDBConnection( String host, int port,
         log.logDebug( "Successful MapiSocket connection to MonetDB established." );
       }
     }
+    mserver.getProtocolVersion();
```
deinspanjer: Any reason for this call?

engine/src/org/pentaho/di/trans/steps/monetdbbulkloader/MonetDbVersion.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+    return result;
+  }
+
+  private void parseVersion( String productVersion ) {
+    if ( productVersion == null ) {
+      throw new IllegalArgumentException( BaseMessages.getString( PKG, "MonetDBVersion.Exception.VersionIsNull" ) );
+    }
+
+    if ( !VERSION_PATTERN.matcher( productVersion ).matches() ) {
```
deinspanjer: Any reason why we shouldn't just make the allowed version pattern specifically only two or three segments rather than allowing any number of them?
`^[0-9]+(\\.[0-9]+){2,3}$`

engine/src/org/pentaho/di/trans/steps/monetdbbulkloader/MonetDBBulkLoader.java
TatsianaKasiankova: No reason. This is left after debug. Fixed.

engine/src/org/pentaho/di/trans/steps/monetdbbulkloader/MonetDbVersion.java
TatsianaKasiankova: The reason is we don't know exactly how many digits segments could be present in valid MonetDB version.
So we do an assumption that any sequence of digits segments separated with dots is valid MonetDB version.
In this case we can always compare current MonetDB version with boundary Jan2014-SP2=11.17.17
This allows to be more stable at checking Monet DB Version.

For example if we get version=11.17.17.6.25, we shouldn't stop checking and we can compare such a version with boundary Jan2014-SP2=11.17.17.

TatsianaKasiankova:
@deinspanjer 
Hi, 
I've fixed one of remark and commented another one.
Could you please review again.

PR #1147, opened 2015-03-26
[BACKLOG-2381] - Adding new dialog for preview that can stay open while ...

brosander:
...trans is running


brosander:
sorry for autoformat, add ?w= to the end of the url to make it a little better

PR #1149, opened 2015-03-26
[SP-1779] - Backport of PDI-13654 - Regression: Execute SQL Script with ...

mtseu:
...escaped double-quote causes parsing issue (5.3 Suite)

Correctly passes through hard-coded quotes that appear within a string,
instead of blindly switching back to SQL mode.


mtseu:
@mattyb149 could you please review/merge?
@mbradbury2014 FYI

PR #1151, opened 2015-03-27
[PDI-13661] XSD Validator with HDFS Files

matthewtckr:
Add support for other VFS objects, by way of using the getInputStream for the XSD Schema file.

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/xsdvalidator/XsdValidator.java
+++ b/engine/src/org/pentaho/di/trans/steps/xsdvalidator/XsdValidator.java
@@ -22,10 +22,8 @@
 
 package org.pentaho.di.trans.steps.xsdvalidator;
 
-import java.io.File;
 import java.io.IOException;
 import java.io.StringReader;
-import java.net.URL;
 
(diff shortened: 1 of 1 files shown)


mbatchelor:
Please include a test case for this

engine/src/org/pentaho/di/trans/steps/xsdvalidator/XsdValidator.java
```diff
@@ -205,21 +202,20 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
 
           // We deal with XML file
           // Get XML File
-          File xmlfileValidator = new File( XMLFieldvalue );
+          FileObject xmlfileValidator = KettleVFS.getFileObject( XMLFieldvalue );
           if ( !xmlfileValidator.exists() ) {
```
mbatchelor: Isn't it possible that this could throw an NPE ?

matthewtckr:
Added a null check to short-circuit a possible NPE

pamval:
Please rebase and reiisue if still applicable. This has some merge conflicts right now.

PR #1155, opened 2015-03-30
PDI-13367 - ETL Metada Injection does not actualize the Source Step in the Inject Metadata tab when renaming the cited step

yuryilyukevich:
Decision the resubmit. For looking how it goes.


yuryilyukevich:
@mattyb149 @kurtwalker

kurtwalker:
I will build this locally first and try it with the latest CI build.  stay tuned.

PR #1163, opened 2015-04-02
Improvement of two little bugs

Dentka15:
Improvement of two little bugs like:
PDI-13657 - Loging from unknown step - propouse solution
PDI-13655 - Repository explorer - opening long time - proposed solutions

It's my first commit to pentaho Kettle.
I'm java developer who use Kettle in job. I think this changes is necessary.
I don't want recompile Kettle code for each new release and update libraries.
Please commit it to trunk.

Regards,
Pawel M


mbatchelor:
Hey Pawel M,

Can you please create a test case for your fix of PDI-13655 so we can merge in your code fix. Thanks.

Dentka15:
Yes, ofcourse. I need litle time cuz i have small actident on playground.

Dentka15:
Test case:
Download test_repo.zip from https://github.com/Dentka15/pentaho-kettle/blob/master/test_reop.zip and put folder "test_repo" in your any repository.
There is 2000 *.ktr files in 1000 folders.

Now start spoon. In menu choose File->Open... (Ctrl + O)

I have to wait 3m 15s until window pop up.
After my fix i'm waiting 2s

My cpu: Intel® Xeon(R) CPU E31225 @ 3.10GHz × 4 

Is it enough to test?

deinspanjer:
@Dentka15 
Thank you for the attempt at providing a pull request for this.  I can see the issue and understand how you are attempting to resolve it here.  Here is some feedback, both general and specific to this pull request:

1. Please split this up into two separate pull requests since the issues resolved here are unrelated.  That makes it easier for us to pull them in as the opportunity arises.

2. Not a requirement, but please consider making a separate branch in github for each issue you are working on and issuing the pull request from that branch instead of from your pentaho-kettle/master branch.  That can prevent difficulties when trying to rebase or update the branch.

3. When you have a test case or example support files, make sure to attach them to the JIRA case rather than putting them in your source tree.  If we were to merge this, that sample would end up in the root of the PDI source folder.

4. I think the idea of caching the repository tree is a reasonable plan, but it needs some more work in addition to unit tests and QA.

5. There are many places in the class where loadRepositoryDirectory() is called which would repeatedly blow away the cached object.  

6. There are also several places where the "tree" variable is overshadowed in methods which can lead to nasty bugs if a developer doesn't notice.  

7. The "noop" comment in the clearSharedObjectsCache() method should be removed if it is no longer a no op.

8. Instead of the refreshRepo() method, setting the tree object to null should just happen in clearSharedObjectsCache().  I'd expect a refresh to actually rebuild the tree which is what happens in loadRepositoryDirectory() already.

We'll keep the JIRA case for the repo work on the backlog, and one of our developers can look at it after it is triaged.  If you want to take another crack at the repo pull request, I'd be happy to review it again.

PR #1165, opened 2015-04-03
Backport of PDI-13367 - ETL Metada Injection does not actualize the "Source Step" in the "Inject Metadata" tab when renaming the cited step (5.3 Suite)

yuryilyukevich:


yuryilyukevich:
@mattyb149 Please, look it.

mattyb149:
We've been having problems with the Kettle build when there are changes to core and other modules, some get picked up and some don't (on the first build). Recommend running the CI job twice to make sure (or at least if the first CI assembly doesn't appear to be fixed or has bugs like the one did in master.

PR #1176, opened 2015-04-06
[BACKLOG-2801] Spoon ( when connected to a repository ) should do file l...

pedrofvteixeira:
...oading by selected file's ID instead of  name


pedrofvteixeira:
@rmansoor, @pentaho-nbaker please review; 

Spoon already favours loading by ID when double-clicking a file via "Tools > Repository > Explore"; 

But via "File > Open.." , when double-clicking a file, it continued on loading by path only. 

This fix is to have "File > Open.." favour loading by ID *when a repository exists* ( Spoon CE with no DB repo connection will use FS, and that continues on loading by path as no other alternatives exist ).

PR #1186, opened 2015-04-08
[PDI-13696] - Text File Input: deteriorated performance for compressed i...

akhayrutdinov:
...nput

### DIFF
--- a/engine/src/org/pentaho/di/core/compress/CompressionInputStream.java
+++ b/engine/src/org/pentaho/di/core/compress/CompressionInputStream.java
@@ -35,4 +35,14 @@ public abstract class CompressionInputStream extends InputStream {
   public int read() throws IOException {
     return delegate.read();
   }
+
+  @Override
+  public int read( byte[] b ) throws IOException {
+    return delegate.read( b );
+  }
+
+  @Override
+  public int read( byte[] b, int off, int len ) throws IOException {
+    return delegate.read( b, off, len );
+  }
 }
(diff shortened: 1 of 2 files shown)


mattyb149:
Unit tests please

akhayrutdinov:
@mattyb149 do you really think this should be tested?)
In contrary with the previous PR (related to FileTextOutput), there's no refactoring - just introducing two public methods from standard API. Of course, I can test, for instance, the usage of read([]), but does it make sense to verify classes from JRE?

brosander:
@akhayrutdinov you could mock the input stream you pass into the constructor and just verify that the delegate is invoked with the correct arguments and that the return is the same as from your mock

mattyb149:
Agreed that the test is not heavyweight and perhaps a bit silly in this case, but we're committed to getting the highest code coverage and quality possible. Adding the test Bryan suggests above will help

akhayrutdinov:
@brosander, @mattyb149, ok if you insist on it.
In fact none of step's classes works directly with the stream. Instead, they utilize a Reader above the BufferedInputStream - https://github.com/pentaho/pentaho-kettle/blob/master/engine/src/org/pentaho/di/trans/steps/textfileinput/TextFileInput.java#L1413, therefore it turns the test to verify the buffered reader reads by filling its buffer, but not by hoisting separate ints

akhayrutdinov:
@mattyb149, @brosander done it. Review it please

PR #1189, opened 2015-04-09
[PDI-13136] - deserialize to file timezone problem

akhayrutdinov:
@mattyb149, @brosander, review it please

### DIFF
--- a/core/src/org/pentaho/di/core/row/value/ValueMetaBase.java
+++ b/core/src/org/pentaho/di/core/row/value/ValueMetaBase.java
@@ -2708,7 +2708,7 @@ public class ValueMetaBase implements ValueMetaInterface {
       writeString( outputStream, dateFormatLocale != null ? dateFormatLocale.toString() : null );
 
       // date time zone?
-      writeString( outputStream, dateFormatTimeZone != null ? dateFormatTimeZone.toString() : null );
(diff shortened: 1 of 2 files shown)


core/src/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -2708,7 +2708,7 @@ public void writeMeta( DataOutputStream outputStream ) throws KettleFileExceptio
       writeString( outputStream, dateFormatLocale != null ? dateFormatLocale.toString() : null );
 
       // date time zone?
-      writeString( outputStream, dateFormatTimeZone != null ? dateFormatTimeZone.toString() : null );
+      writeString( outputStream, dateFormatTimeZone != null ? dateFormatTimeZone.getID() : null );
```
mattyb149: getID() returns a normalized timezone string, which is good. However will it break compatibility with existing transformations? If so we need to get Doc and PM involved

akhayrutdinov: @mattyb149 my first impulse was to serialize a TimeZone instance itself. But that approach will break backward-compatibility with no exceptions.

I deem passing getID() instead of toString() is safe, because I left deserialization intact. The following situation is possible: toString() isn't parseable -> timeZone is set to GMT; then this GMT is saved and will be loaded during next executions. However, it means the time zone was incorrect from the very beginning and GMT was used.

PR #1201, opened 2015-04-14
[BACKLOG-2157] Hop icons for tree and add/remove accumulator

mdamour1976:


mdamour1976:
@e-cuellar more gps coordinates for review

PR #1207, opened 2015-04-16
[BACKLOG-3061] restore warning icon

tgf:


pamval:
@mdamour1976 fix for one of those pdi ui related cases

PR #1221, opened 2015-04-20
[BACKLOG-3018] Update scheduler perspective icon

mdamour1976:


mdamour1976:
@e-cuellar more gps coordinates for review :)

PR #1226, opened 2015-04-21
[PDI-11014] - Error Fields not passed from Select Values step (when erro...

akhayrutdinov:
...r handling is enabled)

- pass field's name to the error row
- add a test
- apply formatting

@mattyb149, @brosander, review it please

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/selectvalues/SelectValues.java
+++ b/engine/src/org/pentaho/di/trans/steps/selectvalues/SelectValues.java
@@ -27,7 +27,9 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.pentaho.di.core.Const;
+import org.pentaho.di.core.exception.KettleConversionException;
 import org.pentaho.di.core.exception.KettleException;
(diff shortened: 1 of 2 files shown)


mattyb149:
The unit tests should include the KettleConversionException path(s) as these are new, and also it looks like the binary-to-normal storage logic was inverted/changed a bit, we should have unit tests for that path as well.

akhayrutdinov:
Thanks, Matt!
As for the exception - I'll add such case.
As for logic - the behaviour is keeped, but the code is slightly changed.
It was:
```
if (cond1) {
  rowData[index] = something
}
if (cond2) {
  rowData[index] = yet another something
}
```

So, I introduced an else branch to avoid extra operations.
```
if (cond2) {
  rowData[index] = yet another something
} else if (cond1) {
  rowData[index] = something
}
```
I inverted the order to keep copatibility with the previous behaviour: if it somehow both conditions were truth, then the second assignment overwrote the first; thus it preceedes the first now

akhayrutdinov:
@mattyb149,
I've update the PR. Could you please take a look?

akhayrutdinov:
@mattyb149, @brosander, review it please

akhayrutdinov:
@mattyb149, @brosander, review it please :)

PR #1229, opened 2015-04-21
[PDI-13745] Mask Password Fields

matthewtckr:
Consolidated field masking, in order to be more consistent across step/job entry dialogs


matthewtckr:
Updated commit to always hide password value, per PM's comments in PDI-13542.

mattyb149:
This Jira case is slated for 6.0. That means this PR needs to be issued against the future-develop branch, or the case needs to be moved up to 5.4.0.

matthewtckr:
@mattyb149 There's a related case, PDI-13542, for the MongoDB plugin that's slated for 5.4, which I was hoping to work on next, using the PasswordTextVar class that's included in this PR.  This would help in standardizing Password field behavior via a single class.

mattyb149:
For that case you could bring over PasswordTextVar and the fix for just the plugin, but we can't merge this PR into master unless the parent case is approved for 5.4 (as we will have to regression test those steps)

ui/src/org/pentaho/di/ui/trans/steps/ivwloader/IngresVectorwiseLoaderDialog.java
```diff
@@ -391,7 +390,7 @@ protected TextVar addStandardTextVar( String labelMessageKey, Control prevContro
     int vertPad = verticalPadding;
     Label label = addStandardLabel( labelMessageKey, prevControl );
     verticalPadding = vertPad;
-    TextVar targetControl = new TextVar( transMeta, shell, SWT.SINGLE | SWT.LEFT | SWT.BORDER );
+    TextVar targetControl = new PasswordTextVar( transMeta, shell, SWT.SINGLE | SWT.LEFT | SWT.BORDER );
```
mbatchelor: Don't agree with this change. addStandardTextVar is called to add many text fields, not just password text fields. This would make all text fields in the VW loader password fields. This needs to be reverted.

ui/src/org/pentaho/di/ui/trans/steps/ivwloader/IngresVectorwiseLoaderDialog.java
```diff
@@ -404,9 +403,7 @@ public void widgetDefaultSelected( SelectionEvent event ) {
   }
 
   protected TextVar addPasswordTextVar( String labelMessageKey, Control prevControl ) {
-    TextVar textVar = addStandardTextVar( labelMessageKey, prevControl );
```
mbatchelor: As above - this needs to be reverted - this is the only place  PasswordTextVar should be created.

ui/src/org/pentaho/di/ui/core/widget/PasswordTextVar.java
```diff
@@ -0,0 +1,60 @@
...
+
+package org.pentaho.di.ui.core.widget;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.ModifyEvent;
+import org.eclipse.swt.events.ModifyListener;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Text;
+import org.pentaho.di.core.variables.VariableSpace;
+
```
mbatchelor: I don't see any handling for variables here. So the checks below (checkPasswordVisible) is designed to only use ***** for the text if the field isn't populated by a variable name. If the field has a variable name in it, it displays the name of the variable. That's probably why they didn't go with |SWT.PASSWORD and instead play with the echo character.

mattyb149: I believe that has changed, and we will no longer be showing the password if it contains a variable. The Jira case (and its parent/related case) both refer to a single behavior, to accept variables but always mask it in the UI

ui/src/org/pentaho/di/ui/trans/steps/ivwloader/IngresVectorwiseLoaderDialog.java
matthewtckr: Thank you, I've reverted that change.

ui/src/org/pentaho/di/ui/trans/steps/ivwloader/IngresVectorwiseLoaderDialog.java
matthewtckr: I've removed the addPasswordTextVar method from the step dialog, as it was not used anywhere in the class, package, or project itself.  The only password in this step is done via the Database Connection meta.

ui/src/org/pentaho/di/ui/core/widget/TextVar.java
```diff
@@ -153,7 +153,7 @@ public void setInsertTextInterface( InsertTextInterface insertTextInterface ) {
     this.insertTextInterface = insertTextInterface;
   }
 
-  private ModifyListener getModifyListenerTooltipText( final Text textField ) {
+  protected ModifyListener getModifyListenerTooltipText( final Text textField ) {
     return new ModifyListener() {
       public void modifyText( ModifyEvent e ) {
         if ( textField.getEchoChar() == '\0' ) { // Can't show passwords ;-)
```
mbatchelor: Is it possible that without the change to the echo character that it used to do, that if you hover over a password field, you'll see the password ?

matthewtckr: If the PasswordTextVar class is used, the getModifyListenerTooltipText() method is overridden, so that only the ToolTip text is shown.  I've change the overridden class to be a NO-OP, and added if statements in the TextVar class to avoid NPE issues.

PR #1230, opened 2015-04-23
[BAD-170] Support for Spark

pavel-sakun:
Package spark plugin
@mattyb149 @smaring @lgrill-pentaho


pavel-sakun:
Double check with PM

pavel-sakun:
Doug confirmed it should be distributed as a plugin both in CE and EE versions

PR #1231, opened 2015-04-23
[BACKLOG-3105] - null is not a variable

kurtwalker:


kurtwalker:
@bytekast @tgf @pamval this should fix the failed data-refinery build.

PR #1238, opened 2015-04-27
[BACKLOG-2588] Introduce versioning by file that implements by content t...

tkafalas:
...ype


mattyb149:
Should we deprecate the original methods instead of replacing them?

PR #1223, opened 2015-04-21
[PDI-12847] - TextFileOutput step splits the rows incorrectly

akhayrutdinov:
@mattyb149, @brosander, review it please

PR #1242, opened 2015-04-28
[TEST] - fixing incorrect test data after splitting was repaired in PDI-...

akhayrutdinov:
...12847

@mattyb149, @brosander, merge it please.
Looks like test data was incorrect from the beginning and we faced this issue after I has repaired splitting in PDI-12847


brosander:
@akhayrutdinov do we have a backward compatibility flag on this? If we're changing long-standing behavior, this may break existing transformations.

akhayrutdinov:
@brosander, what do you mean? It is just a test data. The behaviour was changed with https://github.com/pentaho/pentaho-kettle/pull/1223

brosander:
@akhayrutdinov the need to change the test data means that behavior was changed in a way that may affect existing transformations

I'm aware that #1223 was already merged but think that we may need to bubble this up to be sure that this change won't break existing customers' transformations.

akhayrutdinov:
@brosander, yes, if someone fitted in the incorrect behaviour (it is indeed incorrect, just check it), then #1223 indeed broke such transformations

PR #1243, opened 2015-04-28
[BAD-201] Create and Document Spark samples

pavel-sakun:
Sample job for spark submit job entry


pavel-sakun:
@mattyb149

mattyb149:
Is our Spark entry only supported on a CDH shim? I thought we could point to a vanilla Apache Spark instance? If so, we should probably remove/change the references in the notes to CDH or Cloudera

pavel-sakun:
For 5.4 we claim support for spark running on cloudera 5.3 in yarn mode only

mattyb149:
Ok just checking :)

PR #1248, opened 2015-04-29
[BACKLOG-3098] - Added extension to dialect to support different create ...

diogofscmariano:
...table statement. The original behaviour is the same. Unit tests added

@pamval please review


mattyb149:
Adding methods to the interface (even if we add a default impl to the Base class) may result in breaking other people's plugins (though the risk is very low). We should make sure to document this in the Embed/Extend guide, and even in What's New so other folks can take advantage of it. I think Hive can use it for example

core/src/org/pentaho/di/core/database/DatabaseInterface.java
```diff
@@ -1152,9 +1152,17 @@ public String getModifyColumnStatement( String tablename, ValueMetaInterface v,
    * Customizes the ValueMetaInterface defined in the base
    *
    * @param v the determined valueMetaInterface
-   * @param valueType the sql column type
+   * @param rm the sql column type
    * @return ValueMetaInterface customized with the data base specific types
```
dkincade: Since you were in here cleaning up the javadoc for this method, it would have been a good idea to complete the clean and add the index field (3rd parameter) that is missing as well...

diogofscmariano: change added in https://github.com/pentaho/pentaho-kettle/pull/1249

PR #1253, opened 2015-04-29
[BACKLOG-2157] Hop icon tweaks from UX

mdamour1976:


mdamour1976:
@e-cuellar some gps coordinates for review

PR #1260, opened 2015-05-01
[BACKLOG-2157] additional icon tweaks from UX

mdamour1976:


mdamour1976:
@e-cuellar more svgps coordinates for you

PR #1274, opened 2015-05-07
PDI-13789 - Intermittent ConcurrentModificationException in Database.getValueFromSQLType

jasondyjohnson:
PDI-13789 - Intermittent ConcurrentModificationException in Database.getValueFromSQLType - Fix ConcurrentModificationException by protecting valueMetaPluginClasses initialization with synchronize block.


jasondyjohnson:
Hi, this is my first PR for kettle, please let me know if this looks okay. I included steps to reproduce along with sample transformation file in the JIRA Ticket. I have been running this patched in our production environment for a month with no issues. Thank you.

mbatchelor:
Hi Jason,

Thanks so much for the bug report and the PR. I have concerns about the fix for two reasons - first, I believe it would lead to a performance degradation, but most importantly, it doesn't fix the underlying issue - namely the issue I raised in PDI-12194 ( http://jira.pentaho.com/browse/PDI-12194 ). The collections are static, but they're initialized in non-static methods which is the actual underlying issue.

The proper fix is to initialize both of those collections in a static initializer where it's safe to do so. Then, there is no need for synchronization at all (which would prevent the performance hit).

Would you mind re-working and testing the PR to utilize that method instead, and see if it too resolves the issue the way I believe it should be resolved?

jasondyjohnson:
No problem, I'll create a new PR.

PR #1280, opened 2015-05-07
[BACKLOG-2192] prevent exceptions (widget is disposed)

mdamour1976:


mdamour1976:
@e-cuellar review, this fixes the widget is disposed you showed me
@diogofscmariano this should take care of the issue you were asking about as well

PR #1283, opened 2015-05-08
[BACKLOG-2890]  Changes to PARAMETER parsing in SQLCondition to be mo…

mkambol:
…re lenient.

This allows more flexibility in specification of the parameter clause for dataservice
queries, now supporting lower case or mixed case.  For example
   WHERE Parameter('foo') = 'bar'
 will now work, as will
   WHERE PARAMETER('numericParam') = 3.12


mkambol:
@hudak @bmorrise

PR #1284, opened 2015-05-08
[BACKLOG-3294] - Setting parameters in transformation (connected to DI Server) through exception

AndreyBurikhin:
I wasn't able to figure out how to prevent call of FocusListener when items disposed, so  I just add isDisposed() check.  

@mdamour1976  please review.


mdamour1976:
I believe this was fixed by my commit yesterday, but a little extra protection can't hurt.

PR #1297, opened 2015-05-13
PDI-13434 - Pentaho Reporting Output step - Excel output formatting is different from PRD Excel output

yuryilyukevich:
@tmorgner Please, look it


akhayrutdinov:

@tmorgner, it would be nice if you also take a look and consent the usage is correct

engine/src/org/pentaho/di/trans/steps/pentahoreporting/PentahoReportingOutput.java
```diff
@@ -221,33 +225,54 @@ private void processReport( Object[] r, String sourceFilename, String targetFile
         }
       }
 
-      switch ( outputProcessorType ) {
+      ModifiableConfiguration modifiableConfiguration = (ModifiableConfiguration) report.getConfiguration();
+      String property;
+      Runnable exportTask;
+      switch( outputProcessorType ) {
```
deinspanjer: Since the "property" variable isn't used outside of this switch/case, what do you think about moving all these strings to constants in the corresponding ExportTask classes?

Also, since the modifiableConfiguration is retrieved from the report object, and the report object is consistently passed in to the ExportTasks, could we maybe just set that property in the constructor of the ExportTask which would reduce this switch/case to simply selecting the correct task?

deinspanjer:
Commented on the PR with a suggested change.  re-ping when updated.

-Daniel

engine/src/org/pentaho/di/trans/steps/pentahoreporting/PentahoReportingOutput.java
akhayrutdinov: @deinspanjer, thak you for reviewing.

ExportTask classes are a part of pentaho-reporting. I don't really think it is a good idea to modify PRD for Kettle purposes. Instead, we should utilize PRD's classes in the same manner. The only think I'm concerned I bit is that these values (properties) are not defined as constants, but it is how they are written in PRD.

So, the fix is all about unifying approaches. Previously, Kettle used utility methods, whereas PRD itself moved to exporting tasks.

tmorgner:
The whole patch simple changes how the reporting engine is invoked - basically using more words to do exactly the same as before. I suspect that with or without this change the result is the same. 

Were you able to validate/replicate the bug before you made changes?

Your unit test validates that something runs, but it does not attempt to validate the output. As this is dealing with Excel files, at minimum I would expect that you run the report and then actually check the generated excel file that it contains the correct data.

akhayrutdinov:
@tmorgner, although it looks like doing the same, it actually has a slightly different output for XLS files. You may also check it - Those three spaces are written with the new approach.

Speaking about unit test - yes, it is a bit naive - it just checks something is created. However, it does need a lot of effort to validate the content for all the output formats. Since there was no any tests before, I suppose even such test is better then its absence

PR #1313, opened 2015-05-18
[PDI-13802] - It's impossible to create duplicate connections for sla…

akhayrutdinov:
…ves & database but it's possible to edit these connections giving them the same names in Repository Explorer.It's possible to create duplicate connections for clusters in the View tab

- introducing delegates for clusters and partitions that checks shared object does not exist

@mattyb149, @brosander, review it please. As far as I know, the code  freeze has not been announced yet.

/cc @kcruzada


kcruzada:
@mdamour1976 can you process this ?

mattyb149:
Unit tests pass

PR #1314, opened 2015-05-19
BACKLOG-3367  -- "Stop Yarn cluster" step is failing

kolinus:
What was done:
1) added variable to indicate we are in YARn mode so not to use hardcoded
value

### DIFF
--- a/assembly/package-res/spoon.sh
+++ b/assembly/package-res/spoon.sh
@@ -44,8 +44,12 @@ setPentahoEnv
 LIBPATH="NONE"
 STARTUP="$BASEDIR/launcher/launcher.jar"
 
-# Go to directory where spoon.sh located
-cd $BASEDIR
+if [ -z "$IS_YARN" ]; then
+	# Go to directory where spoon.sh located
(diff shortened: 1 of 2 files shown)


kolinus:
@dkincade @kcruzada Ramaiz is validating this works as expected with installer
@rmansoor  here is the solution

kolinus:
@dkincade @kcruzada  we have verified that this works,
could you please process

assembly/package-res/spoon.sh
```diff
@@ -44,8 +44,12 @@ setPentahoEnv
 LIBPATH="NONE"
 STARTUP="$BASEDIR/launcher/launcher.jar"
 
-# Go to directory where spoon.sh located
-cd $BASEDIR
+if [ -z "$IS_YARN" ]; then
+	# Go to directory where spoon.sh located
+	cd $BASEDIR
+else
+	cd "$BASEDIR"
```
cbandy: Why are the `if` and `else` blocks doing the same thing?

PR #1317, opened 2015-05-21
BACKLOG-3367  -- "Stop Yarn cluster" step is failing

kolinus:
What was done:
1) added variable to indicate we are in YARn mode so not to use hardcoded
value


kolinus:
@dkincade  @kcruzada  could you please review

PR #1324, opened 2015-05-23
[PDI-13633] - Adding a fallback logic to connect the datasource using class if JNDI lookup fails

rmansoor:


rmansoor:
@dkincade , @kcruzada , @mattyb149 please review the change

dkincade:
Looks good - thanks!

PR #1331, opened 2015-05-28
[PDI-13633] - Added a check for named datasource during the connection logic in kettle.

rmansoor:


rmansoor:
@kcruzada , @dkincade , @mattyb149 please review

PR #1333, opened 2015-05-28
[BAD-290] Error appears if variables is used in both named cluster and Hadoop Copy Files step.

e-cuellar:


e-cuellar:
@mattyb149 Please review Sir!

mattyb149:
Looks good but needs unit tests, thanks!

PR #1334, opened 2015-05-28
[BAD-290] Error appears if variables is used in both named cluster and

e-cuellar:
Hadoop Copy Files step.


e-cuellar:
@mattyb149 Please review Sir!

PR #1345, opened 2015-06-01
[PDI-13884] Access Output Truncate Table option

matthewtckr:
Adds the ability to truncate an Access table before writing new rows


brosander:
Base build failed with return code 1

PR #1351, opened 2015-06-03
Merging commits necessary for big-data-plugin to compile with kettle changes

mdamour1976:


mdamour1976:
@lgrill-pentaho @lucboudreau Here are the commits which are necessary to fix issues in the big-data-plugin.  A pre-requisite of this commit has already been issued:  https://github.com/pentaho/apache-vfs-browser/pull/12

PR #1356, opened 2015-06-04
[TEST] - remove timeout for the case

akhayrutdinov:

For an unknow reason the case is executed quite long and fails by timeout: http://ci.pentaho.com/job/kettle-code-coverage/485/testReport/org.pentaho.di.trans.steps.textfileinput/TextFileInputTest/test_PDI695/  

I suspect an impact from JaCoCo as the same test is run very (39 ms) fast during casual build: http://ci.pentaho.com/job/Kettle-build/919/testReport/org.pentaho.di.trans.steps.textfileinput/TextFileInputTest/

### DIFF
--- a/engine/test-src/org/pentaho/di/trans/steps/textfileinput/TextFileInputTest.java
+++ b/engine/test-src/org/pentaho/di/trans/steps/textfileinput/TextFileInputTest.java
@@ -73,7 +73,6 @@ public class TextFileInputTest {
     assertEquals( expected, output );
   }
 
-  @Test( timeout = 100 )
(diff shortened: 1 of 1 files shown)


brosander:
@matthewtckr do we need this timeout?  (Is there a reason this might hang if we remove it?)

lucboudreau:
Please investigate if the test can be fixed first.

akhayrutdinov:
@matthewtckr, could you please give a clarification? As for me, I'd better not introduce any timeouts

matthewtckr:
The timeout was only needed as the original issue would cause the system to freeze up for long periods of time (depending upon input length), and was only meant to put an upper-bound in case it regresses in the future.

PR #1365, opened 2015-06-08
[CLEANUP] - fixing circular dependency between pentaho-karaf-assembly…

rfellows:
… and pdi-osgi-bridge.

pdi-osgi-bridge assembly pulled in pentaho-karaf-assembly and pentaho-karaf-assembly (in the standard feature file) pulls in pdi-osgi-bridge-activators. This was the source of the circular dependency and is potentially fixed by splitting out the assembly from the core of the bridge.

@lgrill-pentaho _NOTE: It is still required (by the build) to build the pdi-osgi-bridge core and activator FIRST, then pentaho-karaf-assembly, and LAST the pdi-osgi-bridge assembly._


rfellows:
@lgrill-pentaho and @pentaho-nbaker please review

depends on: https://github.com/pentaho/pdi-osgi-bridge/pull/16

PR #1374, opened 2015-06-10
Evaluate Excel formula before sending the content

gkfabs:
I need to have the content of the field and not the formula.


mbatchelor:
@gkfabs - can you please file a JIRA request, and reference this PR? Trying to determine whether this is a fix we can accept (based on the use case), or whether it should be a setting in the step dialog. I'm afraid that evaluating every cell could cause performance issues - so likely this needs to be a setting optionally turned on.

PR #1380, opened 2015-06-12
[PDI-13927] - Checking type before casting, adding null handling in g…

brosander:
…etDataSourceFromJndi

### DIFF
--- a/core/src/org/pentaho/di/core/database/util/DatabaseUtil.java
+++ b/core/src/org/pentaho/di/core/database/util/DatabaseUtil.java
@@ -30,6 +30,7 @@ import javax.naming.InitialContext;
 import javax.naming.NamingException;
 import javax.sql.DataSource;
 
+import org.pentaho.di.core.Const;
 import org.pentaho.di.core.database.DataSourceNamingException;
 import org.pentaho.di.core.database.DataSourceProviderInterface;
 import org.pentaho.di.core.database.Database;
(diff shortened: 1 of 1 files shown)


brosander:
@dkincade is 5.4 the right branch?
@e-cuellar Mind pulling this down and testing on your box, mysql generic seems to be working with this fix

e-cuellar:
@brosander @dkincade Brian... I bet this will work! But as you might have found the "dsName" is getting passed as an empty String causing the InitialContext to return an instance of a MemoryContext, which in my opinion... if the dsName is empty or null then this method should not even be called.....  or simply throw a NamingException on a empty dsName and let the flow do the rest....  However I'd be ok with your fix if advised!

brosander:
@e-cuellar the if check at the top should throw the correct exception if dsName is null now

e-cuellar:
@brosander Gotcha!... I like it! :) let me check it!

e-cuellar:
@brosander Sorry Brian... the dsName is not getting passed as null but as an empty string.... thus it will fail the if check.... I suggest using the Const.isEmpty method instead....

brosander:
@e-cuellar good catch!  The get sql button was passing null in.  I guess we could run into an issue if there was a jndi connectio with an empty name that was matched.

e-cuellar:
@brosander Awesome!... Checking the fix locally... I will merge in a bit!

brosander:
@e-cuellar added String.valueOf( dsName ) to ensure we won't get a NPE there either

dkincade:
5.4 is the correct branch - once validated please ensure this is included in master as well.

PR #1381, opened 2015-06-12
[PDI-13796] - including key fields in modeling column mappings for Di…

kurtwalker:
…mension Lookup


kurtwalker:
@bytekast @DFieldFL @pminutillo please review

PR #1388, opened 2015-06-12
[PDI-13940] Installing Plugin from Marketplace Causes Spoon to Crash with SWT IllegalArgumentException

mdamour1976:


mdamour1976:
@e-cuellar @dkincade

mdamour1976:
@mbatchelor

PR #1389, opened 2015-06-13
[SP-1964] - Switching plugin type db repo logic to build a map instea…

brosander:
…d of running a bunch of queries with in clauses and chunking


brosander:
@mattyb149 @mattcasters @hudak mind taking a look and making sure this makes sense?

dkincade:
Based on email conversation - including by this and will validate post-inclusion

PR #1390, opened 2015-06-13
[SP-1964] - Switching plugin type db repo logic to build a map instea…

brosander:
…d of running a bunch of queries with in clauses and chunking


brosander:
see #1389

PR #1396, opened 2015-06-16
[PDI-13952] : preserve exit code of pentaho main code to return from spo…

dshvedchenko:
…on.sh


dshvedchenko:
http://jira.pentaho.com/browse/PDI-13952

dannyteok:
Just want to note that it may be necessary to add a "set -o pipefail" (without quotes) to spoon.sh so that the exit code value is set properly. I had to do this because it was still always 0 before.

PR #1403, opened 2015-06-18
PDI-11215: Cannot Edit Multiway Merge Join After Deleting an Input Ho…

rpbouman:
…p. Error occurred because different parts of the dialog had a different idea of what the input steps were (previous steps according to transformation vs infosteps as per step meta configuration). This fix takes both into account so that any newly linked input steps can be configured, and any saved configuration does not get lost when the incoming hop is disabled.

To make this work the step's rowprocessor needed to be modified as well and this now takes all input steps into account that are also present as info stream. This allows a disabled incoming hop to be ignored during row processing (as it should).


rpbouman:
Darn. Have to fix 2 Checkstyle errors.

PR #1406, opened 2015-06-18
[PDI-13713]: XML Input StAX reader doesn't add namespace info on attributes

TatsianaKasiankova:
There is a fix for attributes and EndElement also.
It's added unit tests.


TatsianaKasiankova:
@mattyb149 Hi, please review the fix.

mattyb149:
Do we need null checks around things like a.getName().x()? Or does the API guarantee non-null names, etc.

TatsianaKasiankova:
Javadoc (e.g.http://docs.oracle.com/javase/7/docs/api/javax/xml/stream/events/StartElement.html) doesn't point that a.getName() can return null. At the same time if the method does it (e.g. getAttributeByName ), it is pointed explicitly.
So we don't need null checks around cases a.getName().x()

mtseu:
@mattyb149 apparently - once an element or an attribute exists in XML doc it has to have a name.

PR #1411, opened 2015-06-19
PDI - Jobexecutor and transexecutor bug.  Moved the add row to the beginning of the if checks for new group tes…

gpotts:
Both the Trans Executor step and the Job executor step is always off by one when submitting rows to the sub transforms they are executing.

This example is for the Trans executor step.  If you set a root Trans to have the first step to continuously generate rows every 5 second and then put a log in front of the trans executor you will see that the row will not be passed until the next row is generated.  This is a problem if you are reading from a stream that never terminates and flushes the results to the executor.   So if your grouping is for N rows it will not send the N rows until the N+1 row read. 

I have moved the groupBuffer.add above the if checks so that they will work properly and create a newGroup.

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/jobexecutor/JobExecutor.java
+++ b/engine/src/org/pentaho/di/trans/steps/jobexecutor/JobExecutor.java
@@ -129,6 +129,7 @@ public class JobExecutor extends BaseStep implements StepInterface {
       }
 
       boolean newGroup = false;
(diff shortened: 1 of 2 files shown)


gpotts:
Let me know if I need to do this to a different branch.  I have also ran the ant test and appears that all tests have passed

gpotts:
It appears that pull requests need to come from Master?  I will reissue a pull from the master branch and test.

PR #1417, opened 2015-06-22
[PDI-11839] "Replace in string" step  creates new field without chars…

SergeyTravin:
…et if lazy conversion is on.


SergeyTravin:
@mattcasters, @brosander, @mattyb149  Could you review it and it's backport https://github.com/pentaho/pentaho-kettle/pull/1418

mattyb149:
I see the Jira case refers specifically to character encoding not being honored, and the fix is to copy the setting from the source ValueMeta. Are there any other source fields we should be copying here?

mattyb149:
Please follow up on any extra settings that should be copied in, we'll have to write a Jira to cover them. I will merge this in the meantime

PR #1418, opened 2015-06-22
[SP-1925][PDI-11839] "Replace in string" step  creates new field with…

SergeyTravin:
…out charset if lazy conversion is on.


SergeyTravin:
@mattcasters, @brosander, @mattyb149 
I would be grateful if you could review the PR before end of the July SP (6.26.2015)

It is backport for https://github.com/pentaho/pentaho-kettle/pull/1417

PR #1421, opened 2015-06-23
[PDI-13637] NPE when running looping transformation - at org.pentaho.di.core.gui.JobTracker.getJobTracker(JobTracker.java:125)

akhayrutdinov:

I've put two commits into this PR. The first is the fix itself with several tests.
The second is a kind of refactoring in tests. I noticed, that both concurrent tests have similar runners. Thus it makes sense to generalise it, separating runner from tests logic.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -603,7 +603,7 @@
 
 
   <target name="tests-all"
-          depends="tests-clean, tests-stage, tests-core, tests-trans, tests-run, tests-blackbox"/>
+          depends="tests-clean, tests-stage, tests-core, tests-trans, tests-run, tests-blackbox, tests-concurrency"/>
 
   <target name="tests-stage-init" depends="init, resolve, init-tests, init-test-reports">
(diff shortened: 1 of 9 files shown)


brosander:
@akhayrutdinov if we're going to leave getJobTrackers() in the class, we should make it possible to synchronize access to the list

Before you could synchronize on the JobTracker object.  We may want to add a getter for the lock and deprecate it with a javadoc in getJobTrackers to make sure you use the lock for access.

akhayrutdinov:
@brosander, I would better remove the method at all, than reveal the lock outside. In Kettle, it is used only in tests now. Do you think it is acceptable to replace its behaviour to returning a defensive copy?

brosander:
@akhayrutdinov I think that's a good option, leave signature the same as before but return a defensive copy when called, if someone really needs to modify the returned list, they can do so with the set method

akhayrutdinov:
@brosander, I've updated the PR

PR #1422, opened 2015-06-23
[BACKLOG-3716] Remove Help Tips from Spoon

AlexanderBuloichik:


AlexanderBuloichik:
I removed only en_US translation since all other languages should be created automatically.

PR #1440, opened 2015-06-26
[CLEANUP] - Adding karaf util to assembly dependencies

brosander:


brosander:
@rmansoor I think this is needed for the platform plugin deployer

PR #1442, opened 2015-06-26
[BACKLOG-3959] Remove Streaming XML Input/XML Input steps

e-cuellar:


e-cuellar:
@mdamour1976 Please Sir!

mdamour1976:
@e-cuellar looks good, can you rename the plugin?  Drop the word "meta" from the name.  I think you have that in there because the classname had it in there when you were deciding what to call it.

e-cuellar:
Wiil do. As soon as I get home
mdamour1976 <notifications@github.com> wrote:

@e-cuellar<https://github.com/e-cuellar> looks good, can you rename the plugin? Drop the word "meta" from the name. I think you have that in there because the classname had it in there when you were deciding what to call it.

-

PR #1444, opened 2015-06-26
[CLEANUP] - Adding blueprint-core and commons-configuration to assemb…

brosander:
…ly dependencies as metaverse expects to find them in the system classloader


brosander:
@mattyb149 @rfellows please review

PR #1446, opened 2015-06-29
Drools update

zachzeus:
This doesn't currently build because it doesn't keep the libraries in the /plugins/kettle-drools5-plugin/lib directory when I do a clean build.


pamval:
@dkincade, this should probably be reviewed by the team working on the PDI steps. Looks very close to what they're doing.

mattyb149:
Also the commits should be squashed and rebased against the latest master branch.

plugins/kettle-drools5-plugin/build.properties
```diff
@@ -0,0 +1,10 @@
+project.revision=TRUNK-SNAPSHOT
```
mattyb149: At present our master-branch artifacts are 6.0-SNAPSHOT not TRUNK-SNAPSHOT.

mattyb149:
There are a few things we need to check on this one:

1) Are the build files up to date?
2) Have any new/upgraded dependencies been vetted by the IP/Build team?
3) Does this preserve backwards compatibility with the previous version of these steps?
4) Does it conform to our coding standards (unit tests, CheckStyle, etc.)?

zachzeus:
Hi Matthew,

Happy to work with you on the questions.

1.  I'm not sure what you mean by this question
2.  All dependencies are identical - just moved
3.  As we are not changing dependencies - just moving it from core to a plugin it should.  Are there existing unit tests?  I've got a "test" that I'm using and it passes
4.  Is there documents/references for this?

Kind regards 

Zach

mattyb149:
@zachzeus Many of these are probably already done or no change is needed, I just made a list for our team to double-check before merging it in. If we find anything we will post it here. There are probably no existing unit tests or integration tests but we are requiring them for new code going forward.

One thing that you can do to help is to squash all the commits in the this pull request to a single one (or two if the second adds tests and such) and rebase against the current master branch. Once that's done you can force-push this branch back to your fork and this pull request will update automatically.

Thank you for your contribution!

zachzeus:
All right I've squashed the commits and fixed the build problem that I had originally.  This should be good to go.

mdamour1976:
https://github.com/pentaho/pentaho-kettle/pull/1557

PR #1450, opened 2015-06-29
[SP-1977] - Backport of PDI-13981 - GA Input: Duplicate entry on 1001st entry

mattyb149:
This is a PR containing Andrey's fix for this case, I verified locally that it works (using @matthewtckr 's reproduction transformation).


mattyb149:
Self-merging after review by @matthewtckr

PR #1457, opened 2015-06-29
[BACKLOG-3857]Embedded MetaStore impl

hudak:
Embeddable metastore for use within TransMeta/JobMeta

Metastore elements and types are stored as entries in the attributes map. Elements and types are serialized as JSON when created or updated.

The following structures are added to the attributes map:
 * `METASTORE.<namespace>`:
   * `<type id>` : `<type as json>`
 * `{'_': 'Embedded MetaStore Elements': 'namespace': '<namespace>': 'type':'<type id>'}`
   * `<element id>` : `<element as json>`


hudak:
@deinspanjer @mkambol Please review

hudak:
Actual writing to disk is handled by the various KTR/KJB persistence mechanisms. Existence of the embedded metastore is transparent.

hudak:
@deinspanjer is there any standardization of attributes saved to the transformation? Is it safe to create new attribute groups per metastore element type? Would this have any negative side effects of lineage?

PR #1462, opened 2015-06-30
[SP-1977] - Backport of PDI-13981 - Google Analytics Input Step: Dupl…

akhayrutdinov:
…icate entry on 1001st entry (5.4 Suite)

- fix the problem
- add test

@mattyb149, @brosander, review it please. This PR includes Matt T's fix as well my unit test, where I tried to mimic the library's behaviour
/cc @matthewtckr

### DIFF
--- a/plugins/googleanalytics/src/org/pentaho/di/trans/steps/googleanalytics/GaInputStep.java
+++ b/plugins/googleanalytics/src/org/pentaho/di/trans/steps/googleanalytics/GaInputStep.java
@@ -230,6 +230,7 @@ public class GaInputStep extends BaseStep implements StepInterface {
     return true;
   }
 
+  // made not private for testing purposes
   List<String> getNextDataEntry() throws KettleException {
     // no query prepared yet?
     if ( data.query == null ) {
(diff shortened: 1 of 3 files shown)


plugins/googleanalytics/src/org/pentaho/di/trans/steps/googleanalytics/GaInputStep.java
```diff
@@ -26,6 +26,7 @@
 import java.util.List;
 
 import com.google.api.services.analytics.Analytics;
+import com.google.common.annotations.VisibleForTesting;
```
dkincade: Please remove this import - we don't have a dependency on guava which defines this annotation and we don't want to add it just to make an annotation that can be replaced by a comment for it...

plugins/googleanalytics/src/org/pentaho/di/trans/steps/googleanalytics/GaInputStep.java
```diff
@@ -230,6 +231,7 @@ public boolean init( StepMetaInterface smi, StepDataInterface sdi ) {
     return true;
   }
 
+  @VisibleForTesting
```
dkincade: Remove this line ... see comment above (just mark it protected and add a comment that it is protected vs private for testing purposes...)

plugins/googleanalytics/src/org/pentaho/di/trans/steps/googleanalytics/GaInputStep.java
mattyb149: We've used this annotation quite a few other places in kettle, should this be discouraged?

akhayrutdinov:
@dkincade, I've done it.
However, I have the same question: I've been using the annotation for quite a long time already with no complaints from reviewers. Should I give up doing it?

dkincade:
@akhayrutdinov we will have to discuss with @pentaho-nbaker about the future of that annotation. I am not against it in general but I hate to bring in a 3rd party library just for a dependency which acts as a comment. It will depend on what else that library includes for us (or conflicts with)

PR #1475, opened 2015-07-02
[PDI-14037] - PMR jobs fails with NoSuchMethodException

mtseu:
The cause is in using Guava API available since v. 14.0
In case of PMR the Guava v.11 comes from Hadoop classpath.
This version of the library does not have the methods called.

The fix is to remove the causing Guava calls and replace them by
more verbose but universaly working code


mtseu:
@hudak @mattyb149 @brosander it's quite hot

PR #1481, opened 2015-07-06
[BACKLOG-3989]

diogofscmariano:
- Removed deprecated interface added in 5.4 release context
- Default behaviour of DatabaseMeta kept in BaseDatabaseMeta
- Unit tests updated


diogofscmariano:
@pamval @dkincade please review

PR #1483, opened 2015-07-06
[TEST] - abort the test after 10 minutes

akhayrutdinov:
This is not elegant, but hanging test is not OK either. This approach doesn't lurk the problem (the test fails), but lets the other tests to be executed, so we can see who is failing and who is passing

### DIFF
--- a/test/org/pentaho/di/cluster/CarteLauncherTest.java
+++ b/test/org/pentaho/di/cluster/CarteLauncherTest.java
@@ -22,11 +22,13 @@
 
 package org.pentaho.di.cluster;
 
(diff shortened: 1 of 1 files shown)


akhayrutdinov:
http://ci.pentaho.com/view/Data%20Integration/job/Kettle-integration-tests/

mattyb149:
Can we use JUnit 4's @Timeout annotation? If this is a JUnit 3 test, can we upgrade to 4 or is that a larger effort for all integration tests?

akhayrutdinov:
Awesome catch, Matt!
I've updated the PR following your approach

PR #1492, opened 2015-07-08
[PDI-13894] Example plugin was rebuilt and moved to Deprecated

andrey-markelov:
@dkincade @mdamour1976  please review and merge


mdamour1976:
I believe you are confusing (and it is very confusing) the "DummyJob" example plugin with the built-in special "DummyEntry".  Additionally, we probably should not remove the DummyJob example plugin, it must be first deprecated (which means fixing the issue instead of deleting it).  We can fix the problem, for now if we recompile dummyjob.jar, but while in there, let's add it to the "Deprecated" group in Spoon, away from "General."

The changes made to JobMeta and Spoon are for the "DUMMY" job entry not the "example plugin", these changes need to be backed out.

I'll cherry-pick your commit locally to make sure I'm not totally wrong about my comments.

mdamour1976:
Well, the cherry-pick -> compile failed because of references to JobMeta.createDummyEntry() that still existed.  My guess is that if you do a build like this:

pentaho-kettle$ ant clean-all resolve dist

You'll get an error:  cannot find symbol

See:   SpoonTreeDelegate.java:336

andrey-markelov:
@dkincade please merge

PR #1494, opened 2015-07-08
[PDI-14063] - Changed behaviour - database name of a data source must…

jvelasques:
… not be null/empty


pamval:
needs to be rebased when appropriate.

PR #1495, opened 2015-07-08
[PDI-13980] Google Analytics NPE with "Get Fields"

matthewtckr:
Prevent an NPE exception when "Get Fields" is used in the dialog, but no results are returned for the given query parameters


ryno1234:
Bump

mdamour1976:
Code is merged in another PR

PR #1496, opened 2015-07-08
[BACKLOG-3907] Make "only use saved connections to XML" default to true

e-cuellar:


e-cuellar:
@mdamour1976 Please Sir!

PR #1503, opened 2015-07-09
[BACKLOG-4208] marketplace install/uninstall exceptions (osx classloader)

mdamour1976:


mdamour1976:
@rfellows @e-cuellar Similar fix as Rob had to make for OSX awhile back for classloader issues.

PR #1519, opened 2015-07-15
[BACKLOG-3723] Change perspective switcher to combo box

AlexanderBuloichik:


pamval:
seems to have been fixed in another pull

PR #1521, opened 2015-07-16
SP-1943: Backport of PDI-13867 - Able to save job/transformation in t…

rpbouman:
…he root folder (5.4 Suite)

Add a list of restricted directories to the SelectDirectoryDialog dialog (Disable Ok button if the selection is in the restricted paths list)
Move the code in TransDialog and JobDialog that sets up the SelectDirectoryDialog into RepositoryDirectoryUI so it can be setup consistently
Restrict selection of / and /home  when creating the SelectDirectoryDialog so people can not save their transformation or job there.
Instead of defaulting the repository to a new directory (which points to the root), use the user home directory
Use current repository directory as default when creating a new Transformation with the mapping step
Use current repository directory as default when creating a new Transformation with the etl metadata injection step
Use current repository directory as default when creating a new Transformation with the simple mapping step
Use current repository directory as default when creating a new Transformation with the single threader step
Use current repository directory as default when creating a new Transformation with the transformation executor step
Use current repository directory as default when creating a new Transformation with the transformation job entry
Use current repository directory as default when creating a new Job with the Job executor step
Use current repository directory as default when creating a new Job with the job job entry.

### DIFF
--- a/ui/src/org/pentaho/di/ui/job/dialog/JobDialog.java
+++ b/ui/src/org/pentaho/di/ui/job/dialog/JobDialog.java
@@ -74,7 +74,6 @@ import org.pentaho.di.job.entry.JobEntryInterface;
 import org.pentaho.di.repository.KettleRepositoryLostException;
 import org.pentaho.di.repository.ObjectId;
 import org.pentaho.di.repository.Repository;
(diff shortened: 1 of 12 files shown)


akhayrutdinov:
@rpbouman, are you going to proceed this PR? The sprint is almost over

PR #1527, opened 2015-07-16
[PDI-13210] - Create a map of extension point plugin for later use

rmansoor:


rmansoor:
@mattyb149 , @pentaho-nbaker  please review

mattyb149:
Looks good to me, but since it's a change in behavior (even for the better), we should have a few unit tests and possibly an integration test around it

rmansoor:
@mattyb149 what do you envision the unit test/integration test to do. There is nothing right now in kettle around extension point. Can you point me to a good please where I can see what I need to accomplish

mattyb149:
The lineage stuff has an integration test that registers extension points: https://github.com/pentaho/pentaho-metaverse/blob/master/core/src/it/java/org/pentaho/metaverse/analyzer/kettle/ExternalResourceConsumerIT.java

and unit tests here: https://github.com/pentaho/pentaho-metaverse/tree/master/core/src/test/java/org/pentaho/metaverse/analyzer/kettle/extensionpoints

In this case a performance tests make more sense, but I'm thinking since this code is critical to execution, it should at least have some tests around the individual methods, prove that the cached instance is used when invoked, etc.

rmansoor:
@mattyb149 Please review now. The tests have been added

PR #1539, opened 2015-07-21
[PDI-14065] - Error message appears if user uses incorrect kettle pro…

yshakhau:
…perty name for quick search


deinspanjer:
I wanted to pull this in also, but in this case, it would be better to have a unit test supporting the change as well.  Please consider adding one to the pull request, otherwise, we'll just leave the jira case in the backlog for later triage.

PR #1565, opened 2015-07-27
When I use getOutputStream(), Japanese wasn't translated correctly.

naotoyoshimoto:
When I use getOutputStream(), Japanese wasn't translated correctly.
So, I suppose that I should change getOutputStream to getWriter.


naotoyoshimoto:
This problem has already fixed in https://github.com/pentaho/pentaho-kettle/pull/1600,
So, I close this pull request.

PR #1567, opened 2015-07-27
[PDI-13953] - Excel Input Dialog does not show XLSX files when browsing

yshakhau:
Need to be internationalized for other languages.


mbatchelor:
Hi, and thanks for the contribution. We have a later contribution that also includes internationalized messages. Closing this PR in favor of using #1703

PR #1575, opened 2015-07-28
[BACKLOG-3806] When deleting a step, focus should be set to the "Yes" button

e-cuellar:


mdamour1976:
@e-cuellar Looks good, nice job.

PR #1585, opened 2015-07-30
[BACKLOG-4497] When using undo/redo after deleting a step, only the hop is removed

e-cuellar:


e-cuellar:
@mdamour1976 Please Sir!

PR #1599, opened 2015-08-05
[BACKLOG-4543] [Google Analytics Step] The field Key File doesn't work with windows path

pavel-sakun:
@mattyb149 take a look please


plugins/googleanalytics/test-src/org/pentaho/di/trans/steps/googleanalytics/GoogleAnalyticsApiFacadeTest.java
```diff
@@ -0,0 +1,19 @@
+package org.pentaho.di.trans.steps.googleanalytics;
```
dkincade: Thank you so much for adding a test! 
Please NOTE that tests also require the proper license headers at the top of each file.

pavel-sakun:
@dkincade fixed, thank you for noticing

PR #1604, opened 2015-08-07
[BACKLOG-4068] Identify and fix the places in code where xalan librar…

pedrofvteixeira:
…y is being used specifically

	- only pentaho-kettle will have xalan explicit calls removed;
	- other places, such as mondrian, require xalan libs as part of its Xmla parsing logic; those will be left untouched, and xalan lib will continue being included for the time being
	- WS testing samples added to BACKLOG-4068


pedrofvteixeira:
@rmansoor , @CodeOnCoffee please review

PR #1605, opened 2015-08-07
[BACKLOG-4084] Remove the patched spring and spring core jar from the…

pedrofvteixeira:
… platform assembly

	- same thing needs to be done in pentaho-kettle's assembly


pedrofvteixeira:
@rmansoor ,  @CodeOnCoffee please review

PR #1606, opened 2015-08-09
SetVariables without native2ascii

teruok78:
In SetVariables JobEntry,
I want to write a property file with Japanese strings, not in native2ascii-encoded strings.
With multi-byte character languages such as Japanese, Korean and Chinese,
it's too much trouble with native2ascii encoded strings.

Both UTF8 strings and native2ascii-encoded strings are accepted.


teruok78:
Commit "fixed Japanese garbled page" (f3bf230) was not intentionally added.

I want pullrequest with only "Reading properties file without native2ascii in SetVariables JobEntry" (7179e88).

pamval:
hey, @teruok78. In that case, please close the current pull request and issue a new one containing only that particular commit.

teruok78:
Thank you, @pamval.
It was my mistake.
I will try another pull request.

PR #1610, opened 2015-08-10
[PDI-13661] XSD Validator with HDFS Files

matthewtckr:
Add support for other VFS objects, by way of using the getInputStream for the XSD Schema file.


matthewtckr:
Rebase of #1151

PR #1621, opened 2015-08-13
[PDI-6567] - Transformation entry in kettle job broken in Spoon if transformation is specified by reference and the target is later deleted

akhayrutdinov:

The idea of the fix is to take into account "deleted" flag, that has been ignored previously. Also, it was set incorrectly. There will be a slightly different behaviour now: DB repository displayed ```/null```, EE server's plugin caused NPEs - henceforth, an empty string will be showed.

I don't like code duplication in dialog classes either, thus I created an improvement request for it: http://jira.pentaho.com/browse/PDI-14228


akhayrutdinov:

PR #1638, opened 2015-08-18
fixed DI server Japanese mistranslations

teruok78:
There are two simple mistranslations in DI server messages.


teruok78:
Thank you, Pedro!

PR #1654, opened 2015-08-21
[BAD-215] Add Support for the Simba Driver

stepanovdg:
- Some databases return different url according to parameters sent to them.
- Need to clear cached DatabaseInterfacesMap in case some database plugins are changed/added/removed


stepanovdg:
Wait. Do not merge.

PR #1656, opened 2015-08-21
Fixed invalid error message for Mail step

ryno1234:
If a "Comment" field was not specified and files were to be attached, an erroneous error message would be presented saying that content files could not be found when in fact it is the "Comment" field that was not specified.


dkincade:
@mattyb149 @brosander @mdamour1976 please have a quick look ...

PR #1658, opened 2015-08-24
BACKLOG-2941 -- Shared Dimension step should automatically update data

kolinus:
provider name when the output step name is changed

What was done:
1) During the fix of initial issue the tests using TransMeta.addOrReplace
failed. SO added a unit tetst demonstrating this and created the fix for
the issue.


kolinus:
@deinspanjer @mattyb149  could you please merge

PR #1669, opened 2015-08-26
[BACKLOG-3825] Updates for publish targets

mdamour1976:


mdamour1976:
@tgf For review / hope this does the trick..

PR #1676, opened 2015-08-28
TextFileInput changes

AlexanderBuloichik:


deinspanjer:
Alex, what do you think about changing the calls to TextFileInputUtils.getLine() in other steps such as S3 and CSV over to use OldTextFileInput.getLine() instead?  The code is nearly identical, but I think it might be safer to keep the other steps referencing the old code until they have specifically been refactored themselves.

AlexanderBuloichik:
> calls to TextFileInputUtils.getLine()
You are right. Will change.

deinspanjer:
Hmm.. in that case, shouldn't you change all the other references to OldTextFileInput in those steps to the new ones as well?  Otherwise, that pull request would involve a bunch of changes in addition to the deletes.

deinspanjer:
Can you look at creating pull requests for the other two projects that failed when we tried to merge this in Friday as well?

PR #1685, opened 2015-08-29
[BACKLOG-4882] Get Data Integration.app working without java mac comp…

pedrofvteixeira:
…at issues

.app was not using PENTAHO_JAVA_HOME when called from an installer build on a OS X
        - was always relying on 'a' java natively installed ( Apple or Oracle ) found by universalJavaApplicationStub
	- PENTAHO_JAVA_HOME wasn't being considered: this is due to the fact that set-pentaho-env.sh wasn't being called prior to the universalJavaApplicationStub
	- created a new JavaApplicationStub that first calls set-pentaho-env, sets JAVA_HOME with the PENTAHO_JAVA_HOME value ( if one exists ), and calls for universalJavaApplicationStub afterwards
        - universalJavaApplicationStub only tries to find 'a' java nativelly installed *if* JAVA_HOME has not been set yet

### DIFF
--- a/assembly/build.xml
+++ b/assembly/build.xml
@@ -77,6 +77,7 @@
     <echo>Construct the distribution package...</echo>
     <!-- make JavaApplicationStub executable for OSX -->
     <chmod file="${approot.stage.dir}/Data Integration.app/Contents/MacOS/JavaApplicationStub" perm="ugo+rx"/>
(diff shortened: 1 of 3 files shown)


pedrofvteixeira:
@CodeOnCoffee, @rmansoor , @mbatchelor, @dkincade  please review

rmansoor:
@pedrofvteixeira it does looks ok to me but i will let the folks of have the mac to validate this pull request

pedrofvteixeira:
Updated PR, by adding the assembly/build.xml chmod change ( not included in the original commit )

PR #1694, opened 2015-09-01
[PDI-14288] FTPS Bad Log Message

matthewtckr:


matthewtckr:
Resubmitting as #1696

PR #1703, opened 2015-09-01
[PDI-13953] Excel Input Dialog Browse XLSX

matthewtckr:
Fix a typo to allow showing Excel 2007 files


ui/src/org/pentaho/di/ui/trans/steps/excelinput/ExcelInputDialog.java
```diff
@@ -1121,7 +1121,8 @@ public void widgetSelected( SelectionEvent e ) {
           SpreadSheetType type = SpreadSheetType.getStpreadSheetTypeByDescription( wSpreadSheetType.getText() );
           switch ( type ) {
             case POI:
-              extentions = new String[] { "*.xls;*.XLS;*.xlxs;*.XLSX", "*" };
+            case SAX_POI:
+              extentions = new String[] { "*.xlsx;*.XLSX", "*" };
```
mbatchelor: Matt - shouldn't this line contain *.xls, *.XLS ?

matthewtckr:
The POI and SAX_POI options are shown in the Dialog as only for Excel 2007 formatted files, whereas the JXL option is for the older Excel format.  Removing the extensions for POI/SAX_POI should help to avoid confusion, as they'll need to select the appropriate parser for the file type that is to be selected.

PR #1705, opened 2015-09-02
[PDI-11496] Check if webservice is available has wrong dialog title

stepanovdg:
* Only text properties files were changed.


stepanovdg:
@dkincade Details in the letter.

PR #1706, opened 2015-09-02
[PPP-3413] - Jobs / kjb exported from BA server contain all BA datasources

akhayrutdinov:
@mattyb149, @brosander, @deinspanjer, review it please at your earliest convenience.
The ticket was included into the scope in the nick of time and it does block the service pack's release.

This PR comes with https://github.com/pentaho/pdi-ee-plugin/pull/274 and should **not be merged before it**.

PR #1634, opened 2015-08-17
[SP-1980][PPP-3405] transformations / ktr exported from BA server con…

SergeyTravin:
…tain all BA datasources

Conflicts:
	build.xml
	engine/src/org/pentaho/di/trans/TransMeta.java
	engine/test-src/org/pentaho/di/trans/TransMetaTest.java

PR #1709, opened 2015-09-02
[SP-2095] - Backport of PPP-3413 - Jobs / kjb exported from BA server contain all BA datasources (5.4 Suite)

akhayrutdinov:
- kettle-engine
  - generalize changes from PPP-3405 by adopting that approach for AbstractMeta
  - introduce a test template and two implementations for checking the private DBs capturing routine
- other
  - revert test changes from SP-1980

@mattyb149, @brosander, @deinspanjer, review it please at your earliest convenience.
The ticket was included into the scope in the nick of time and it does block the service pack's release.

This is a backport of https://github.com/pentaho/pentaho-kettle/pull/1706


engine/src/org/pentaho/di/trans/TransMeta.java
```diff
@@ -2878,7 +2880,7 @@ public void loadXML( Node transnode, String fname, IMetaStore metaStore, Reposit
             }
           }
         }
-        setPrivateTransformationDatabases( privateTransformationDatabases );
+        setPrivateDatabases( privateTransformationDatabases );
```
mattyb149: Why replace this call with the AbstractMeta version, rather than just have setPrivateTransformationDatabases (the existing method) call the AbstractMeta version?  

With the fix above, what happens if some external code calls setPrivateTransformationDatabases()? The TransMeta's privateTransformationDatabases would be set but its AbstractMeta's privateDatabases would not be. The inverse appears true as well (if setPrivateDatabases() is called then getPrivateTransformationDatabases() is called)

akhayrutdinov: @mattyb149. your remark makes sense, but I suppose it is not good to keep the old method, as it is useless.

The previous fix was merged recently, it is not included into any published version. Therefore, it is just a part of SNAPSHOT. And no one guarantees snapshot steadiness, especially, if it is not an interface. That's my vision. Do you still disagree?

engine/src/org/pentaho/di/base/AbstractMeta.java
```diff
@@ -1749,4 +1760,21 @@ protected boolean shouldOverwrite( OverwritePrompter prompter, Props props, Stri
     }
     return overwrite;
   }
+
+  /**
+   * Returns the set of databases available only for this meta or <b>null</b> if it was not initialized.
+   * Note, that the internal collection is returned with no protection wrapper!
+   * @return <b>nonSharableDatabases</b>
+   */
+  public Set<String> getPrivateDatabases() {
```
mbatchelor: I would rather this was worded privateConnections not privateDatabases.

engine/src/org/pentaho/di/trans/TransMeta.java
mattyb149: So the old method has never been in a 5.4 SP, patch, or other release?

akhayrutdinov: Yep, here is that PR: https://github.com/pentaho/pentaho-kettle/pull/1634 (a fortnight ago)

PR #1714, opened 2015-09-02
[BACKLOG-4978] - Reordering assignment in get method to avoid initial…

brosander:
…ization before clear remaining valid


brosander:
@rfellows

PR #1722, opened 2015-09-04
[PDI-14304] API changes needed in org.pentaho.di.trans.steps.fileinpu…

AlexanderBuloichik:
…t.BaseFileInputStepMeta for backward compatibility

### DIFF
--- a/core/src/org/pentaho/di/core/gui/BaseFileInputFieldInterface.java
+++ b/core/src/org/pentaho/di/core/gui/BaseFileInputFieldInterface.java
@@ -22,7 +22,7 @@
 
 package org.pentaho.di.core.gui;
 
-public interface TextFileInputFieldInterface extends Comparable<TextFileInputFieldInterface> {
+public interface BaseFileInputFieldInterface extends Comparable<BaseFileInputFieldInterface> {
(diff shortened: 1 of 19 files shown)


core/src/org/pentaho/di/core/gui/BaseFileInputFieldInterface.java
```diff
@@ -22,7 +22,7 @@
 
 package org.pentaho.di.core.gui;
 
-public interface TextFileInputFieldInterface extends Comparable<TextFileInputFieldInterface> {
```
deinspanjer: I'm very sorry, I didn't realize that part of this chain was a core class that has existed for a while.  For org.pentaho.di.core.gui.TextFileInputFieldInterface, please make a new one called BaseFileInputFieldInterface with the same signature, then change the existing TextFileInputFieldInterface to extend that one and deprecate it.  We can refactor any usage we find in pentaho-kettle, but that way we won't break any external usages of the interface.

deinspanjer:
Other comments added in the JIRA case.  I'll go ahead and close this PR out since we know that we don't want quite as drastic of an approach on the refactoring.

PR #1725, opened 2015-09-04
[PDI-14304] Class names refactoring for new TextFileInput and API cha…

AlexanderBuloichik:
…nges needed in BaseFileInputStepMeta for backward compatibility


core/src/org/pentaho/di/core/gui/TextFileInputFieldInterface.java
```diff
@@ -22,6 +22,9 @@
 
 package org.pentaho.di.core.gui;
 
+/**
```
deinspanjer: Need to remove the deprecate since there is no replacing class

AlexanderBuloichik:
Already merged as other pull request

PR #1728, opened 2015-09-04
[BACKLOG-5063] Missing plugins trigger an error when opening transformation/job that were saved on DI repository

mdamour1976:


mdamour1976:
@e-cuellar

PR #1731, opened 2015-09-07
[PDI-14172] - Text file input Repeat doesn't work

akhayrutdinov:
- set parsing value to null to prevent the regression from PDI-11194
- add test

As far as I know, the code freeze was not announced so far, hence I suppose it make sense to include this fix into 6.0 to remove the regression from the brand new release.


akhayrutdinov:

PR #1732, opened 2015-09-07
[MARKET-308] Allow the opening of SpoonBrowser tabs without address bar

nantunes:
Adds new signatures to SpoonTabsDelegate::addSpoonBrowser; keeps the default behaviour (show address bar).

### DIFF
--- a/ui/src/org/pentaho/di/ui/spoon/Spoon.java
+++ b/ui/src/org/pentaho/di/ui/spoon/Spoon.java
@@ -8690,16 +8690,16 @@ public class Spoon extends ApplicationWindow implements AddUndoPositionInterface
     delegates.jobs.addJobGraph( jobMeta );
   }
 
-  public boolean addSpoonBrowser( String name, String urlString, LocationListener locationListener, boolean toolbar ) {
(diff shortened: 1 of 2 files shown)


nantunes:
@graimundo please review

ui/src/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -8690,12 +8690,20 @@ public void addJobGraph( JobMeta jobMeta ) {
     delegates.jobs.addJobGraph( jobMeta );
   }
 
+  public boolean addSpoonBrowser( String name, String urlString, LocationListener locationListener, boolean toolbar ) {
```
graimundo: Use a "predicate" argument name for boolean arguments instead of a noun (toolbar). I suggest using "showControls" which is the name of the argument in the SpoonBrowser class.

nantunes: @graimundo ok, I've done that.

graimundo:
Looks ok to me. 
@mattyb149 please review / merge. This will be used for the new marketplace tab view.

PR #1733, opened 2015-09-08
[BACKLOG-5071] Allow to partial date value parsing, as in previous Kettle version

AlexanderBuloichik:


deinspanjer:
Alex, could you please add a test case that proves this works so we don't have a future regression?

AlexanderBuloichik:
Fixed, sorry.

PR #1734, opened 2015-09-08
[BACKLOG-3431]Fire content listeners on set/clear

hudak:
Replaced AbstractMeta inheritance of ChangeFlag with delegation to prevent circular override.


hudak:
@mattyb149

hudak:
Updated with JobMeta test

PR #1735, opened 2015-09-08
[MARKET-315] Prevent NPE when quitting with open SpoonBrowser tabs on Windows

nantunes:
On Windows, browser tabs are instances of `org.eclipse.swt.ole.win32.OleClientSite`
that manages embedded ActiveX controls.

When disposed (at least in SWT 4.3.2) `OnInPlaceDeactivate` is called and if the
tab has focus forces its move to the next focusable control (the next open tab).
Great when closing a tab!

Unfortunately when quiting the app it seems it can end up on an already disposed tab,
throwing a NullPointerException.

Moving in advance the focus to other non-embedded control prevents it.

@graimundo

### DIFF
--- a/ui/src/org/pentaho/di/ui/spoon/Spoon.java
+++ b/ui/src/org/pentaho/di/ui/spoon/Spoon.java
@@ -4932,6 +4932,13 @@ public class Spoon extends ApplicationWindow implements AddUndoPositionInterface
     }
 
     if ( exit ) {
+      // on windows [...].swt.ole.win32.OleClientSite.OnInPlaceDeactivate can
+      // cause the focus to move to an already disposed tab, resulting in a NPE
+      // so we first move the focus to somewhere else
+      if(this.selectionLabel != null && !this.selectionLabel.isDisposed()) {
+        this.selectionLabel.forceFocus();
+      }
+
       close();
     }
 


graimundo:
Looks ok to me. 
@mattyb149 / @rpbouman please review / merge

ui/src/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -4932,6 +4932,11 @@ public boolean quitFile( boolean canCancel ) throws KettleException {
     }
 
     if ( exit ) {
+      // on windows [...].swt.ole.win32.OleClientSite.OnInPlaceDeactivate can
+      // cause the focus to move to an already disposed tab, resulting in a NPE
+      // so we first move the focus to somewhere else
+      this.selectionLabel.forceFocus();
```
mattyb149: Probably should be a null and/or disposed check on this label?

nantunes: It's the `close()` that will do the actual quitting of the application, so there is no reason for the control be already disposed. Also, that label exists from the beginning to the end of the application, so it shouldn't be null either. But I can add the check, to be extra safe.

mattyb149: Yes thank you. We have a few places where widgets shouldn't be disposed and yet we see weird things sometimes (especially on exit)

nantunes:
@mattyb149 null/disposed check added, please review.

PR #1738, opened 2015-09-09
[MARKET-314] Allow close and disposal of SpoonBrowser tabs other than the Welcome tab

nantunes:
It was assumed the Welcome tab was the only SpoonBrowser. The closing of other
SpoonBrowser tab caused the closing of Welcome tab, and probably the deficient
disposal of the intended tab.

@graimundo


nantunes:
Sorry, disregard this, moved patch to a feature branch, instead of master. Now in #1739.

PR #1746, opened 2015-09-10
[BACKLOG-5064] Added data service client

bmorrise:


mattyb149:
Need to revert this due to unit test (and thus build) failures, it adds a native plugin for a DatabaseMeta not in core.

PR #1756, opened 2015-09-14
PDI-11863: As a Salesforce ETL Developer Make Upsert Mapping exclude non-updatable fields

TatsianaKasiankova:
What was done:
1)It's moved to org.pentaho.di.trans.steps.salesforceinput.SalesforceConnection
2)The fields with type="id" should not be excluded - we need mapped this fied when Upset Comparison Field=Id.
3)Deleted the code that belongs to PDI-11864. It should be in the separate PR.


TatsianaKasiankova:
@pamval @rpbouman Hi, please look at this one.

PR #1758, opened 2015-09-14
[BACKLOG-4449] -Implement the MetaStore portion of the Import process

rfellows:
adding a null check, discovered as part of the backup/restore work relating to the metastore

### DIFF
--- a/plugins/pdi-pur-plugin/src/org/pentaho/di/repository/pur/PurRepository.java
+++ b/plugins/pdi-pur-plugin/src/org/pentaho/di/repository/pur/PurRepository.java
@@ -2378,7 +2378,12 @@ public class PurRepository extends AbstractRepository implements Repository, jav
     // if there's a match on id, replace the element
     boolean found = false;
     for ( int i = 0; i < origSharedObjects.size(); i++ ) {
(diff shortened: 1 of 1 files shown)


plugins/pdi-pur-plugin/src/org/pentaho/di/repository/pur/PurRepository.java
```diff
@@ -2378,7 +2378,9 @@ private void updateSharedObjectCache( final RepositoryElementInterface element,
     // if there's a match on id, replace the element
     boolean found = false;
     for ( int i = 0; i < origSharedObjects.size(); i++ ) {
-      if ( ( (RepositoryElementInterface) origSharedObjects.get( i ) ).getObjectId().equals( idToFind ) ) {
+      RepositoryElementInterface repositoryElementInterface = (RepositoryElementInterface) origSharedObjects.get( i );
+      ObjectId objectId = repositoryElementInterface.getObjectId();
```
mattyb149: Any chance repositoryElementInterface is null here?

rfellows: i don't know, probably. That wasn't what was causing the issue i was dealing with, it was the object id. want me to add another check?

rfellows:
i added another check to make you happy :)

PR #1761, opened 2015-09-15
[MARKET-306] - Moved Show Plugin Information menu entry to the tools …

graimundo:
…menu. This entry will remain a responsibility of spoon and marketplace will add its own menu entry to the same menu grouping.


graimundo:
@mattyb149 please review

PR #1762, opened 2015-09-15
[BACKLOG-5087] Updated the startup memory setting to min:1G max:2G for client tools

dkincade:


dkincade:
@tkafalas @rmansoor @pedrofvteixeira @pentaho-nbaker please review...

PR #1117, opened 2015-03-18
[SP-1769] - Backport of PDI-13570 - Regression: Shared DB-connections wi...

akhayrutdinov:
...thout .kdb file not showing in Spoon 5.3 (5.3 Suite) - fixing an error

@mattyb149 review it please. 
This PR relates to #1113. I missed an error yesterday

PR #1764, opened 2015-09-15
[PDI-14349] Remove password from URI logging

AlexanderBuloichik:


core/src/org/pentaho/di/core/vfs/KettleVFS.java
```diff
@@ -151,8 +154,8 @@ public static FileObject getFileObject( String vfsFilename, VariableSpace space,
 
       return fileObject;
     } catch ( IOException e ) {
-      throw new KettleFileException( "Unable to get VFS File object for filename '"
```
deinspanjer: Please move the fileObject variable in line 147 out of the try scope, then call getFriendlyURI(fileObject) instead of the string filename signature.

core/src/org/pentaho/di/core/vfs/KettleVFS.java
```diff
@@ -390,4 +393,29 @@ public static FileInputStream getFileInputStream( FileObject fileObject ) throws
     return new FileInputStream( fileObject.getName().getPathDecoded() );
   }
 
+  /**
+   * Returns a "friendly path", this is a path without a password.
+   */
+  public static String getFriendlyURI( String uri ) {
```
deinspanjer: Please make a new signature for getFriendlyURI( FileObject fileObject) that calls FileObject.getFriendlyURI().  Then, change this one to call this.getFriendlyURI(this.getFileObject(uri)) // minus the "this."

We want both signatures available to developers, but since they will already have a handle to a fileObject, we should cater to that.

core/test-src/org/pentaho/di/core/vfs/KettleVFSTest.java
```diff
@@ -0,0 +1,39 @@
...
+package org.pentaho.di.core.vfs;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class KettleVFSTest {
+
+  @Test
+  public void checkFriendlyURI() {
```
deinspanjer: Please add a few more asserts for other types of URIs such as:
*  a relative path: "./subdir/file.txt"
*  a file: scheme: "file:///tmp/test.txt"
*  hdfs schemes with and without passwords
*  ftp, sftp, ftps - with and without password
*  zip, gzip, tar
*  a complex wrapped scheme: tar:gz: hdfs://user:pass@somehost:8080/downloads/some_file.tar.gz!/mytar.tar!/path/in/tar/README.txt

Note, I'm not sure whether your test is ever actually hitting the VFS code.  It might be only following the exception path since these tests were not actually valid files.  Try testing some actual files that exist in the test resources directory using the fileObject signature.

engine/src/org/pentaho/di/trans/steps/textfileoutput/TextFileOutput.java
```diff
@@ -647,7 +647,7 @@ public void openNewFile( String baseFilename ) throws KettleException {
         }
 
         if ( log.isDetailed() ) {
-          logDetailed( "Opened new file with name [" + filename + "]" );
+          logDetailed( "Opened new file with name [" + KettleVFS.getFriendlyURI( filename ) + "]" );
```
deinspanjer: Check the changes in this file to see if any of the getFriendlyURI calls should be using the FileObject signature instead.

deinspanjer:
Closed.  Campbell provided an alternate PR in #1117

PR #1765, opened 2015-09-15
PDI-14364 In spoon, the Insert/update step’s “Edit mapping” leaves Update Y/N selection in Update/Insert Table empty.

dhanasgit:
To solve this issue, I added one line of code in 
ui/src/org/pentaho/di/ui/trans/steps/insertupdate/InsertUpdateDialog.java.


bkemper-pentaho:
I don't believe this is a valid fix.   The field setting should be gotten from the metadata step, not simply set to 'Y'...

PR #1767, opened 2015-09-15
[PDI-11864]: As a Salesforce ETL Developer make Salesforce Upsert Ste…

TatsianaKasiankova:
…p Better by having reference fields handled correctly

This implementation has been done as part of possible changes within the PR https://github.com/pentaho/pentaho-kettle/pull/435.
What was done:
1)For all sales force fields we add field.getName() into targets list;
2)For the reference fields we also scan referenced objects and then uses idLookup fields in them to display target field name in format:
objectReferenceTo:externalIdField/lookupField


TatsianaKasiankova:
@rpbouman , @pamval 
Hi, please look at this one.

pamval:
@rpbouman please review

PR #1778, opened 2015-09-17
[BACKLOG-5205] Change GUIOption cast exception to warning

mattyb149:
Backport of #1777 to 6.0.0.0


mattyb149:
Without it, PME will log an error with stack trace. I fixed it as I believe we don't want to ship a product that looks like it throws Exceptions on startup

dkincade:
Ok - thanks!!!!

PR #1782, opened 2015-09-17
[PDI-14358] Add option to TextFileInput to use null for missing values

overtron:
- add a check box on content tab of dialog
- true means empty values are null
- false means empty values are NULL_STRING constant
- defaults check box to true for backward compatibility
- if tag isn't found in XML, defaults to true


pamval:
Pull requests must be issued against the master branch.

Inclusion in past releases will be evaluated internally after that.

thanks!

PR #1783, opened 2015-09-18
[PDI-12814] Need to prohibit empty names when adding new users or roles

MikhailHubanau:
[PDI-12814](http://jira.pentaho.com/browse/PDI-12814) has UX Feedback with two requirement items.
This PR fixes issue but implements only second item of UX Feedback (first item is still need to be implemented). 
Also it should be noted that this PR implements "Dialog Button: Close" requirement (from second UX item) but org.eclipse.swt.widgets.MessageBox (which is used for dialog) does not support button name changes (always will be "OK"). So if we use current MessageBox as dialog base "Dialog Button: Close" implementation will not be visible.


MikhailHubanau:
EE https://github.com/pentaho/pdi-ee-plugin/pull/287

MikhailHubanau:
@dkincade @mbatchelor Could you please review?

dkincade:
@mattyb149 @rmansoor @pentaho-nbaker please review or pass along to others

MikhailHubanau:
Added.

PR #1784, opened 2015-09-21
PDI-14364 In spoon, the Insert/update step’s “Edit mapping” leaves Update Y/N selection in Update/Insert Table empty.

dhanasgit:
Solution
1. Copied the update status of targets to a hashmap (Target as Key,
update status as value) from the wReturn during the existing map
generation
2. At the time of populate the wReturn, get the update status from the
hash map for the target and set the same.
3. If the Target is new and there is no update status, make “Y” as
default.

Why not Copy from Meta/Input?
- If the user change the update status by keying/mouse selection, the
change is not updated to input until the “OK” event, so the wReturn
always has the latest value.


dhanasgit:
Thanks for accepting the fix.

PR #1787, opened 2015-09-22
[PDI-13028] - Export repository with rule: Job/Transformation with Note is not exported

YuryBY:
@mattcasters, @brosander, @mattyb149 would you mind of reviewing/merging?


YuryBY:
@mattcasters, @brosander, @mattyb149, @mbatchelor, already updated.
1) Now if we get an exception we don't write this trans/job and continue looping
2) All messages are moved to bundle

PR #1793, opened 2015-09-23
[PDI-14393] Salesforce Login URL

matthewtckr:


pamval:
@matthewtckr, is the new url available already ?

matthewtckr:
"It works in my testing" :)

PR #1800, opened 2015-09-25
[BACKLOG-5247] Agile BI : Fixed save button not opening save dialog f…

DFieldFL:
…or modeler and the save buttons no longer are dithered for analyzer

Only check for missing plugins if object is of type AbstractMeta. For example AnalyzerVisualizationMeta does not extend AbstractMeta.


DFieldFL:
@e-cuellar please review and merge

PR #1808, opened 2015-09-26
[BACKLOG-5339] - Making PluginRegistry notifyAll() when plugins are r…

brosander:
…egistered or unregistered


dkincade:
http://devci.pentaho.com/job/6.0-snapshot/job/pdi-core-plugins/214/

PR #1821, opened 2015-09-29
[BACKLOG-5360] Fixed import of Trans that calls Job/Trans

mattyb149:
Added check for ObjectLocationSpecificationMethod of "filename" before performing any operations on the imported trans.

Also fixed some CheckStyle violations and added unit tests for TransFileListener.


mattyb149:
@mdamour1976 Mind a review? Please and thanks!

PR #1823, opened 2015-09-29
[PDI-13616] - Import leaves the 'Reference' field blank when resolution error occurs

YuryBY:
@mattcasters, @brosander, @mattyb149, @mbatchelor, take a look, please.
Essentially was done:
1) added two exceptions in order to provide better comunication between layers
2) in case of an exception is thrown in .lookupreference we are continue looping to get all references for repopository object (here was the bug)

Also files checked with checkstyle. Is provided JUnit test which checks behavior after fix: looping through all referenced objects and checking if appropriate exceptions is thrown


pamval:
@brosander @mattyb149  please take a look

core/src/org/pentaho/di/core/exception/IdNotFoundException.java
```diff
@@ -0,0 +1,103 @@
+/*! ******************************************************************************
```
dkincade: Please add unit tests for this class (it will be very simple but it will help code coverage). Be sure the check/specify when null should and should not be valid.

core/src/org/pentaho/di/core/exception/LookupReferencesException.java
```diff
@@ -0,0 +1,115 @@
+/*! ******************************************************************************
```
dkincade: Please add unit tests for this class.

dkincade:
Why does JobMeta show up as all-lines changed? Is this a line-ending change (and should it be)?

YuryBY:
@dkincade, essential changes in JobMeta is here: https://github.com/YuryBY/pentaho-kettle/commit/0af4646e53316bdacab5312dd10a7c8b50a53ce0#diff-90225f35069b79ca28e7e3c1ebb9ad15R32 all other changes is just applying of pentaho-coding-standarts formatter

core/src/org/pentaho/di/core/exception/IdNotFoundException.java
YuryBY: @dkincade is done for this file

YuryBY: @dkincade what about this one: https://github.com/pentaho/pentaho-kettle/pull/1823/files#diff-043e715fee3ba6d93077bd1b66a24c41R2 - what kind of test do you mean? There is no business logic in the class to test - just getetrs and setters. Or do you maen something else? CC: @mtseu

engine/src/org/pentaho/di/repository/RepositoryImporter.java
```diff
@@ -1035,4 +1026,32 @@ public void setOverwrite( boolean overwrite ) {
   public String getVersionComment() {
     return versionComment;
   }
+
+  private void saveTransformationToRepo( TransMeta transMeta, RepositoryImportFeedbackInterface feedback )
```
brosander: @YuryBY can we put an abstract saveToRepo( RepositoryImportFeedbackInterface feedback ) in AbstractMeta and implement it in JobMeta and TransMeta? Now that they have a common parent class, I'd like to start moving away from "if jobs do this, if trans do that" where possible

YuryBY: @brosander, it seems there is no possibility to do it here because of following reasons:
1) "repo" field 
2)  JobMeta and TransMeta both have .lookupRepositoryReferences and AbstractMeta has not
Maybe I missed something. Thanks

core/src/org/pentaho/di/core/exception/IdNotFoundException.java
dkincade: This should be a simple unit test to write. Using each of the constructors, I want to verify a get an Exception object that is what I expect. (When I provide a message, make sure I get the same message back. Same for other fields). What is the expected behavior when I pass a null in for any/all the fields? Does it actually behave that way.

Again, should be simple because this is an Exception class and it is just a bean. But there are lines of code that we can cover with unit tests that will help improve our project-level unit test line coverage.

YuryBY:
@brosander, @mattyb149 tests are added

YuryBY:
@brosander, @mattyb149 any updates? CC: @pamval

YuryBY:
@mattyb149, have done

pamval:
@mattyb149, please review and merge if acceptable. We're working on finalizing the October SP today or tomorrow.

PR #1825, opened 2015-09-29
[PDI-14440] Fixed TransExecutor step to declare main output fields

mattyb149:
Only clear the input row meta if asking about a non-"main" output step.

Also added unit tests


mattyb149:
@rfellows Mind a review? Please and thanks!

PR #1829, opened 2015-10-01
[PDI-12814] - Need to prohibit empty names when adding new users or roles

akhayrutdinov:
Since the ticket is included in October SP, do not ignore it please.

These commits do not contain UI changes, that were refused previously. This PR solves two problems:

1. it makes EE repo and DB repo behave similarly (the previous changes related to this ticket treated only EE)
2. it prohibits not only creating empty names, but also "similar", e.g. "user" and "USER"

/cc @pamval

### DIFF
--- a/engine/src/org/pentaho/di/repository/RepositoryCommonValidations.java
+++ b/engine/src/org/pentaho/di/repository/RepositoryCommonValidations.java
@@ -1,0 +1,58 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
+ * Copyright (C) 2002-2015 by Pentaho : http://www.pentaho.com
+ *
+ *******************************************************************************
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ******************************************************************************/
+
+package org.pentaho.di.repository;
+
+
+import static org.apache.commons.lang.StringUtils.isBlank;
+
+/**
+ * @author Andrey Khayrutdinov
+ */
+public class RepositoryCommonValidations {
+
+  /**
+   * Validates {@code user}'s data. Common rule for all repositories is: both login and name must contain at least one
+   * meaningful char.
+   *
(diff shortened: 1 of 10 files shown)


akhayrutdinov:

bkemper-pentaho:
Can we please just fix the requested case and not add "extra" checks.    The challenge is that we don't know if the system ALREADY has duplicate names "Admin" and "admin".    This could be a behavior change which could cause a regression in existing systems.   We KNOW the empty names cause issues.   The other fix MAY be appropriate - or it may not be.

akhayrutdinov:
@bkemper-pentaho,
That is mentioned in the ticket's description, although its name does not reflect it:
```2. Users and Roles should not be able to be duplicated (which is why the server is throwing the stack trace).```

bkemper-pentaho:
Possibly - except you changed the definition of "duplicate"  - unless there's a different behavioral problem



Bob Kemper
Executive Vice President of Engineering
Pentaho, a Hitachi Data Systems company.
5950 Hazeltine National Dr.
Suite 340
Orlando, FL 32822
(407)-965-4019

[cid:B5A9BB15-0721-4B7C-A823-DD6BD3ABC7E9]

akhayrutdinov:
@bkemper-pentaho, sorry, but I don't understand what is wrong. Can you please explain it?
These changes will not affect the existing repos and let them work as usual. They will just prevent new collisions. Futhermore, we can backport not all of them, if you afraid that is too risky.

bkemper-pentaho:
If a user comes in from a 3rd party provider -   their policy MAY allow users with similar names as long as the case is different.   If they are automating the insertion of those users into the system - you just prevented their ability to add those users.

Additionally - if the users already existed (prior versions of the product) this COULD cause issues if they try to update user information,  import or export an existing system, etc.    The answer is that WE DON'T KNOW.   Which is why you shouldn't just decide on your own to change the System Behavior in these fashions.



Bob Kemper
Executive Vice President of Engineering
Pentaho, a Hitachi Data Systems company.
5950 Hazeltine National Dr.
Suite 340
Orlando, FL 32822
(407)-965-4019

[cid:46175395-0438-4A2A-B83C-847F6E6A01B9]

akhayrutdinov:
@bkemper-pentaho, ok for 3rd party that is possible. Will it be ok from your perspective, if i keep duplication checks, but make them case-intolerant?

bkemper-pentaho:
So technically this work (on the endpoint) is being done in relation to PDI--14357

And I'm fine with adding the duplicate check (not case sensitive) - but be sure to comment that on the case and ensure there are valid unit tests for all of the scenarios - (null, empty,  duplicate,  mixed case non-duplicates, etc.)



Bob Kemper
Executive Vice President of Engineering
Pentaho, a Hitachi Data Systems company.
5950 Hazeltine National Dr.
Suite 340
Orlando, FL 32822
(407)-965-4019

[cid:A7D17B96-BE48-4214-ABF5-CC9FCFBA13DC]

PR #1833, opened 2015-10-02
[BACKLOG-5360] Applied import logic to JobFileListener

mattyb149:
@dkincade Backport of #1831 to 6.0


dkincade:
We will not be pulling this in to the 6.0.0.0 build. We will need to get this in the 6.0.0.1 SP. 
@jgrajewski

PR #1835, opened 2015-10-02
[BACKLOG-4596] [Google Analytics Input Step] Pressing preview display all previous errors displayed in command line (seems is caching)

pavel-sakun:
Removed logging in favor of showing error message to user as logging from steps' editors falls into GENERAL category and will be show for every trans run in preview mode.

@mattyb149

### DIFF
--- a/plugins/googleanalytics/src/org/pentaho/di/trans/steps/googleanalytics/GaInputStep.java
+++ b/plugins/googleanalytics/src/org/pentaho/di/trans/steps/googleanalytics/GaInputStep.java
@@ -231,7 +231,7 @@ public class GaInputStep extends BaseStep implements StepInterface {
       // There is necessarily an account name associated with this, so any NPEs or other exceptions mean bail out
       accountName = analytics.management().accounts().list().execute().getItems().iterator().next().getName();
     } catch ( Exception e ) {
-      logError( e.toString() );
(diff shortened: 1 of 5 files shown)


dkincade:
Please add unit testing to ensure the updated behavior is correct and prevent regressions.

pavel-sakun:
@dkincade No functional changes in non UI code. Joined 2 method together as they were coupled and added checks like this https://github.com/pentaho/pentaho-kettle/pull/1835/files#diff-390e9cc9540fde475cc6e554b57f9288R1219 to prevent NPEs we had previously. Did manual test for methods changed, got expected error.

mattyb149:
Not sure what the code coverage is for existing unit tests (if there are any), but since you're in here changing stuff (not for 6.0 or an SP), can you add other unit tests where applicable? If you can't test the PR changes because they're UI, perhaps we can still get more coverage by adding other tests (in a separate commit) where possible.  Thanks!

pavel-sakun:
Ok

dkincade:
The creation of TableItem can be refactored out to a separate method that can be unit tested.

pavel-sakun:
@dkincade did it

dkincade:
@brosander @mattyb149 @hudak please review ...

PR #1836, opened 2015-10-02
[PDI-12814] - Need to prohibit empty names when adding new users or roles

akhayrutdinov:
@mattyb149, @brosander, this is my third attempt to publish the changes.
This PR has nothing related to UI changes. It prohibits creating users (for DB and EE) and roles (EE) if:

1. Name is empty
2. Name already exists

As a separate test case it is checked, that ```user``` and ```User``` are handled as different.
/cc @bkemper-pentaho

### DIFF
--- a/engine/src/org/pentaho/di/repository/RepositoryCommonValidations.java
+++ b/engine/src/org/pentaho/di/repository/RepositoryCommonValidations.java
@@ -1,0 +1,58 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
+ * Copyright (C) 2002-2015 by Pentaho : http://www.pentaho.com
+ *
+ *******************************************************************************
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


akhayrutdinov:
@mattyb149, @brosander, do you mind looking at this one?

brosander:
@akhayrutdinov shouldn't we internationalize the error messages? If we don't, a user may not know how to fix the problem.

akhayrutdinov:
@brosander, that makes sense. As for EE, the messages are put in the bundles. However, there is no any i18n in DB repo. Not sure it is OK, but that's it. That is why I haven't done my own either

brosander:
@akhayrutdinov it's unfortunate that we don't already have those messages internationalized but we should be using i18n going forward

akhayrutdinov:
@brosander, ok, I'll i18n the affected classes

brosander:
@akhayrutdinov you don't have to do the whole class, just the new messages

brosander:
@akhayrutdinov we can write up a different case for the classes, just don't want to add more tech debt

akhayrutdinov:
@brosander, I added the i18n bundle and updated the PR. Review it please

PR #1839, opened 2015-10-05
PDI-14287: Get Data from XML returns incorrect XML node value

TatsianaKasiankova:
There are fix and unit tests.

The problem was:
The step uses dom4j SAXReader to create DOM4J tree from SAX parsing events.(please see http://dom4j.sourceforge.net/dom4j-1.6.1/apidocs/org/dom4j/io/SAXReader.html)
From SAX documentation: SAX parsers may return all contiguous character data in a single chunk, or they may split it into several chunks (http://www.saxproject.org/apidoc/org/xml/sax/ContentHandler.html#characters%28char[],%20int,%20int%29)
So potentially even plain text like: "text1 text2 " could be devided into several text nodes (e.g. for reasons having to do with parser efficiency and input buffering).
So e.g. we will have the node with two text nodes: "text1" and "text2". Then we call node.valueOf( text() ) (see https://github.com/pentaho/pentaho-kettle/blob/5.4.0.1-R/engine/src/org/pentaho/di/trans/steps/getxmldata/GetXMLData.java#L738-L739)
and we get only one first text node because dom4j uses  org.jaxen.XPath implementation of the metod.
The easiest fix is just to normalize already parsed document.

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/getxmldata/GetXMLData.java
+++ b/engine/src/org/pentaho/di/trans/steps/getxmldata/GetXMLData.java
@@ -171,6 +171,7 @@ public class GetXMLData extends BaseStep implements StepInterface {
           BaseStep.closeQuietly( is );
         }
       }
+      data.document.normalize();
 
       if ( meta.isNamespaceAware() ) {
         prepareNSMap( data.document.getRootElement() );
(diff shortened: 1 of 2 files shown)


TatsianaKasiankova:
@mattyb149, @brosander Hi, please review changes.

mattyb149:
What are the performance implications for normalizing the document? This code can be called once for each row (if the filenames are provided in a field), so if it is a slow operation, this could degrade performance

rpbouman:
Personally I don't think this should be merged. The reason is not so much the chosen solution (although I share @mattyb149's concern); I simply think that from the standpoint of the xpath specification there is no bug at all. I added a comment to http://jira.pentaho.com/browse/PDI-14287 to explain why.

mattyb149:
Your explanation makes sense to me, @dkincade what do you think?

dkincade:
Verified that the original case is actually not a bug. We are going to close this PR.

PR #1851, opened 2015-10-08
[BACKLOG-5441] Fix size retrieve from VFS file.

AlexanderBuloichik:
Additional output field "Size" will be null if size cannot be retrieved.

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/fileinput/BaseFileInputStep.java
+++ b/engine/src/org/pentaho/di/trans/steps/fileinput/BaseFileInputStep.java
@@ -427,9 +427,14 @@ public abstract class BaseFileInputStep<M extends BaseFileInputStepMeta, D exten
     data.path = KettleVFS.getFilename( file.getParent() );
     data.hidden = file.isHidden();
     data.extension = file.getName().getExtension();
-    data.lastModificationDateTime = new Date( file.getContent().getLastModifiedTime() );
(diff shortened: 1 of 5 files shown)


engine/src/org/pentaho/di/trans/steps/fileinput/BaseFileInputStep.java
```diff
@@ -427,9 +427,14 @@ protected void fillFileAdditionalFields( D data, FileObject file ) throws FileSy
     data.path = KettleVFS.getFilename( file.getParent() );
     data.hidden = file.isHidden();
     data.extension = file.getName().getExtension();
-    data.lastModificationDateTime = new Date( file.getContent().getLastModifiedTime() );
     data.uriName = file.getName().getURI();
     data.rootUriName = file.getName().getRootURI();
-    data.size = file.getContent().getSize();
+    if ( file.getType().hasContent() ) {
```
dkincade: I am sure this could use a unit (or maybe an intergration) test to prevent a regression in this area.

AlexanderBuloichik: I already created unit test for that, but I refactored unit tests for the new TextFileInput and CsvInput since Monday. I.e. unit test uses new code that not in git yet (see #1852). #1852 contains unit test against "Size" this issue. I will discuss order of commits with Michael.

PR #1877, opened 2015-10-21
Add argument number in LDAPConnection.Update.Attribute message

mbarre:
The argument number was missing. This caused the LDAP Output Step failure.


mbatchelor:
Thank-you for the contribution!!

adriens:
good job @mbarre !

PR #1936, opened 2015-11-19
[PDI-14608] Added case for DASH so AutoDoc step no longer fails

agnielson:
I reviewed the other relevant code and couldn't for the life of me decide if DASH from JobPainter/TransPainter should have been something else, so I'm recommending another case be added to SwingGC.  Feel free to change the values for the dash variable.


mbatchelor:
Checkstyle failed on the commit by agnielson. I copied over his fix from here to my fork, made it conform, and checked in the updated PR ( https://github.com/pentaho/pentaho-kettle/pull/1938 ). Thanks @agnielson

PR #1970, opened 2015-11-27
[PDI-14618] - Creating a connection from an open KTR or KJB when connected to repository allows to create connections with the same names (identical) or with the same names but with differ case

akhayrutdinov:
- check uniqueness of name when adding and editing connection from step dialog
- add tests
- fix Checkstyle violations

@brosander, @mdamour1976, review it please.


akhayrutdinov:
I re-worked the approach a bit

PR #1975, opened 2015-11-30
[BAD-297] - Hadoop file output fails append to file

stepanovdg:
* UI changes  - enable extension field in case of using filename from field


stepanovdg:
related https://github.com/pentaho/big-data-plugin/pull/571

stepanovdg:
@brosander @deinspanjer Could you please take a look? It`s according the first point of the http://jira.pentaho.com/browse/BAD-297 :
1. In tab File, when the "Accept filename from field" is selected, the "Extension" field is greyed out indicating it is not used. However it still gets used.
So It was decided to enable for editing extension field (as it was and still is being used) to not break somebody`s existing ktrs.

PR #1974, opened 2015-11-30
[PDI-11374] Fix Unit Test

matthewtckr:
Updated unit test as the expected exception in the test was no longer thrown in the init method

PR #1976, opened 2015-12-01
[SP-2258][PDI-11374] Fix Unit Test

SergeyTravin:
Updated unit test as the expected exception in the test was no longer
thrown in the init method


SergeyTravin:
@brosander, can you please take a look at this? It is a backport for unit test from
https://github.com/pentaho/pentaho-kettle/pull/1974

PR #2000, opened 2015-12-22
[PDI-14015] - Repository import does not populate the transformation reference correctly on the Metadata Injection step

AndreyBurikhin:
@mdamour1976 please review.

p.s.: it is a "cheap & dirty" solution. if it is too dirty - let me know i'll present the other one - we can add a new interface which specifies method to recalculate path for current location. Then this interface should be added to all steps/job entries which need path correction on import/export.


mdamour1976:
@AndreyBurikhin love cheap and dirty.  I think this solution is more broad than affecting all steps/jobs that need path correction.  Do you think it would be possible to create a unit test?

AndreyBurikhin:
@mdamour1976  the test was added

mdamour1976:
@AndreyBurikhin thanks for the test, looks good.  Can we squash it too.

AndreyBurikhin:
@mdamour1976 squashed

PR #2005, opened 2015-12-24
[PDI-14724] - Disabling error handling hop doesn't disable error hand…

yshakhau:
…ling of the step


mbatchelor:
Looks like this was superseded by PR #2014

PR #2013, opened 2015-12-30
[PDI-8930] - PluginMessages can avoid synchronization if it uses ConcurrentHashMap

akhayrutdinov:
- replace HashMap with ConcurrentHashMap

### DIFF
--- a/engine/src/org/pentaho/di/core/util/PluginMessages.java
+++ b/engine/src/org/pentaho/di/core/util/PluginMessages.java
@@ -22,8 +22,7 @@
 
 package org.pentaho.di.core.util;
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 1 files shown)


brosander:
@akhayrutdinov this doesn't seem related to PDI-8935

Also, couldn't we use ConcurrentHashMap still?  The readwritelock can have multiple threads blocked waiting on the write lock at the same time, so you'd still get the chance of creating garbage PluginMessages but below you throw away the duplicate one right away (and no locking).  PluginMessages just have a String member variable and no other initialization and so should be really cheap to create in the race condition knowing we'll always return the first one put into the map:

    PluginMessages result = MESSAGES_MAP.get( packageName );
    if (result == null) {
      return MESSAGES_MAP.putIfAbsent( packageName, new PluginMessages( packageName ) );
    }
    return result;

Please let me know what you think.

akhayrutdinov:
@brosander, yes, that makes sense. I've updated the PR and changed its title.

PR #2015, opened 2015-12-30
[PDI-13901] - Export Linked Resources: Metadata Injection step not working

AndreyBurikhin:
@mdamour1976 please review


mdamour1976:
Looks nice @AndreyBurikhin 

It looks like you split the PR across multiple commits on purpose, but I don't think we need this level of granularity (it just adds noise).  If you can squash these it would be great.

AndreyBurikhin:
@mdamour1976 squashed

PR #2024, opened 2016-01-05
[PDI-12967] Prevent Hop Flipping for Start Job Entry

matthewtckr:


deinspanjer:
Very reasonable quick fix for this.  We'll need some unit tests that cover it though.

mdamour1976:
I can't really imagine there being any reasonable unit tests that can be written for this.  I'd be fine merging it as is.

deinspanjer:
Okay, I was basically just thinking about something that would test whether a normal hop was flippable and a start hop was not.

PR #2027, opened 2016-01-08
[BACKLOG-6041] Reducing log message from basic to debug

mdamour1976:
This message is now showing up in the logs since the KettleXmlPluginLifecycleListener exists without implementing GUIOption (nor do we want it to).


mdamour1976:
@rfellows

PR #1737, opened 2015-09-09
[PDI-14172] - Text file input Repeat doesn't work

akhayrutdinov:
- applying changes from 9e3e72d9d85ce5222a44b4ebe46d4e7163057e9e to new implementation of the step

@mattyb149, @brosander, review it please. This is a fix migration from https://github.com/pentaho/pentaho-kettle/pull/1731 to the new implementation

PR #2036, opened 2016-01-12
[PDI-13684] - Change in null behavior in 5.x version

akhayrutdinov:
@brosander, review it please.

I'll give you a context of the case. With PDI-11194, conversion rules were changed, and this caused regressions. Actually, the fix was correct, and it is legacy code, which does not handle all cases properly. Namely, it is not calling `isNull()` to get whether or not the given value is undefined; meanwhile it is possible to force Kettle not to distinguish empty strings and nulls, that's why we get troubles.

There was a discussion (see it on http://jira.pentaho.com/browse/PDI-14358), and the final resolution was: "we need to somehow improve conversion (see http://jira.pentaho.com/browse/PDI-14693), but for now the changes must be as local as possible -- not to cause any other differences in behaviour".

The list of affected classes is given on http://jira.pentaho.com/browse/PDI-13684?focusedCommentId=235753&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-235753.

So, here is my solution. It is quite clumsy, but it does not affect anything outside the list -- let's just explicitly check if the value is null, and set it to `null` manually. In that case, everything (seems to) works fine.

This PR contains changes for all items from the list except two. For TextFileInput, the same solution has been applied with https://github.com/pentaho/pentaho-kettle/pull/1737. For SwitchCase, it is needless, as the very next action after conversion is `isNull()` call, meaning all properties are handled correctly.

Let me know if you have any questions.

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/datagrid/DataGrid.java
+++ b/engine/src/org/pentaho/di/trans/steps/datagrid/DataGrid.java
@@ -59,8 +59,8 @@ public class DataGrid extends BaseStep implements StepInterface {
   }
 
   public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws KettleException {
-    if ( data.linesWritten >= meta.getDataLines().size() ) // no more rows to be written
-    {
+    if ( data.linesWritten >= meta.getDataLines().size() ) {
+      // no more rows to be written
       setOutputDone();
       return false;
     }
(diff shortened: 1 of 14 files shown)


akhayrutdinov:
@brosander, review it please.

akhayrutdinov:
@deinspanjer, may be you can take a look at this?
/cc @pamval

engine/src/org/pentaho/di/trans/steps/datagrid/DataGrid.java
```diff
@@ -84,16 +84,19 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
...
-        outputRowData[i] = StringUtil.EMPTY_STRING;
+        outputRowData[ i ] = StringUtil.EMPTY_STRING;
       } else {
 
         ValueMetaInterface valueMeta = data.outputRowMeta.getValueMeta( i );
         ValueMetaInterface convertMeta = data.convertMeta.getValueMeta( i );
         String valueData = outputLine.get( i );
 
-        outputRowData[i] = valueMeta.convertDataFromString( valueData, convertMeta, null, null, 0 );
+        if ( valueData != null && valueMeta.isNull( valueData ) ) {
```
deinspanjer: ValueMetaBase.isNull() is already designed to check whether valueData is null.  Seems odd to pre-check it before calling the method that is specifically designed to test the nullness of objects.

deinspanjer:
My note about the null comparison is the only thing I see worth mentioning.  It is harmless if it stays in, so I wouldn't say it has to be changed before merge.

engine/src/org/pentaho/di/trans/steps/datagrid/DataGrid.java
akhayrutdinov: @deinspanjer, if `EMPTY_STRING_AND_NULL_ARE_DIFFERENT == false`, then empty strings are treated as `null`s. It is meaningless to invoke `isNull(null)`, thus I put the first condition there

deinspanjer: It is meaningless to invoke isNull(null) if you already know the object is null.  In this case, you know because you've duplicated some of the logic in the isNull() function before you call it.

This type of comparison is usually intended to safety check a reference being passed into a function that might not properly handle a null being passed in.

However, ValueMetaBase.isNull() already does null checking inside of it since that is the whole purpose of the function, and it returns a boolean value.  If you pass a null into that function, it will return true on line 3318, or 3309 if the field is a binary storage field.

Doing the valueData != null comparison as a safety check is unnecessary, and it is redundant to the code that already exists in isNull().

That said, it is harmless, and I'd rather see the PR just merged rather than force a change unless there was already other revisions that needed to be made.

PR #2041, opened 2016-01-13
[BACKLOG-6083] Fixing issue with saving ktr/kjb in DI repository

mdamour1976:


mdamour1976:
@tgf without this file, the platform won't register the resources (REST) in plugin.spring.xml

PR #2065, opened 2016-01-19
[SP-2353] - Backport of PDI-14194 - Error with S3 CSV Input step - Unable to Get Fields Correctly (5.4 Suite)

mr9d:
- s3csvinput.jar was rebuilt and replaced
@brosander please review.


brosander:
@mr9d comment added to jira case

mr9d:
According to discussion with Larry Grill, we should bring to GitHub whole source code of the S3CSVInput plugin, not just compiled jar. Will create separate code review.

PR #2067, opened 2016-01-19
[BACKLOG-6213] Fixing getNrSubdirectories now that /etc is hidden. Te…

pentaho-nbaker:
…st added to verify and prevent regression.


pentaho-nbaker:
@mdamour1976 Should do it

PR #2090, opened 2016-01-26
[CLEANUP] Missing dependency.

mkambol:
@lgrill-pentaho


lgrill-pentaho:
Have reservations because this is a very heavy (2.7mb) jar to bring in for limited functional use.

lucboudreau:
For clarity's sake, this is not a new dependency. It was previously pulled transitively from Mondrian, but Mondrian has dropped it yesterday.

mkambol:
Just for reference, I added the case below to get rid of xalan in kettle-xml-plugin, as well.  xalan has limited use in the plugin, but it may be tricky to assure error messages are consistent after exorcising the jar.
http://jira.pentaho.com/browse/BACKLOG-6382.

PR #2100, opened 2016-01-28
fixes #PDI-14893 don't try again to get MetaData from PreparedStateme…

MarWestermann:
…nt as this is already the fallback if driver doesn't support that

### DIFF
--- a/core/src/org/pentaho/di/core/database/Database.java
+++ b/core/src/org/pentaho/di/core/database/Database.java
@@ -2163,11 +2163,8 @@ public class Database implements VariableSpace, LoggingObjectInterface {
           setValues( par, data, ps );
         }
         ResultSet r = ps.executeQuery();
-        ResultSetMetaData metadata = ps.getMetaData();
-        // If the PreparedStatement can't get us the metadata, try using the ResultSet's metadata
-        if ( metadata == null ) {
-          metadata = r.getMetaData();
-        }
+        ResultSetMetaData metadata = r.getMetaData();
+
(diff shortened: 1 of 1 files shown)


matthewtckr:
This change would impact every single type of database connection in PDI, and it's recommended in this PR to just remove the `ps.getMetaData()` statement altogether for all database connections?  The JDBC standard states that the [PreparedStatement should return null](https://docs.oracle.com/javase/7/docs/api/java/sql/PreparedStatement.html#getMetaData()), which is what the application is doing currently.

Alternatively, you could always write a modified Gupta driver to catch that exception internally before sending the result to PDI, as is done with some of the Apache Hive drivers.

MarWestermann:
but as I already stated, its in the method which is the fallback for the case when meta data on prepared statements are not supported. IMHO it makes no sense to say: hey retrieving meta data on prepared statements are not supported.. lets try it nevertheless. in the method which does the meta data fetch on ps it is wrapped in a try catch and the catch calls the fallback method.. and the fallback method then call again the same.

MarWestermann:
just looked at the code again, the previous implementation additionally has the problem, that it first executes the query and retrieves the resultSet and then uses the resultSet to get the metadata. If that doesn't result in null the execution of the query was needless.

mbatchelor:
Closing this in favor of fixes made in #2631. Thanks for the contribution and the JIRA case. However, this fix didn't fix the GuptaDatabaseMeta to avoid the exception in the first place.

PR #2116, opened 2016-02-01
[PDI-13465] Startup Logging Messages

matthewtckr:


plugins/pdi-pur-plugin/src/org/pentaho/di/repository/pur/PurRepositoryConnector.java
```diff
@@ -135,10 +139,14 @@ public synchronized RepositoryConnectResult connect( final String username, fina
         public Boolean call() throws Exception {
           // We need to add the service class in the list in the order of dependencies
           // IRoleSupportSecurityManager depends RepositorySecurityManager to be present
-          LogChannel.GENERAL.logBasic( "Creating security provider" );
+          if ( log.isBasic() ) {
+            log.logBasic( "Creating security provider" );
```
mbatchelor: I understand that you didn't create the hard-coded message. But the mandate from leadership is that if you modify a class that has hard-coded strings that are emitted in the logs or to users (except for debug messages), they need to be externalized. This class already has a message bundle - can you please take the extra step of moving the hard-coded logBasic strings into the bundle? Thanks.

matthewtckr:
@mbatchelor Updated with message bundles

PR #2117, opened 2016-02-02
[PDI-12988] Analytic Query Feedback Messages

matthewtckr:


engine/src/org/pentaho/di/trans/steps/analyticquery/AnalyticQuery.java
```diff
@@ -130,6 +130,10 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
       processQueue();
     }
 
+    if ( log.isBasic() && checkFeedback( getLinesRead() ) ) {
+      logBasic( "linenr " + getLinesRead() );
```
mbatchelor: Sorry dude, cannot accept hard-coded strings in new PRs. The class already has a message bundle - add the message string to that, and use BaseMessages.getString( PKG, "AnalyticQuery.Log.linenr", Integer.toString(getLinesRead()) (or similar)

matthewtckr:
@mbatchelor Updated with message bundles

PR #2121, opened 2016-02-02
[BACKLOG-6046] Fix cluster label displaying on Linux

e-cuellar:


e-cuellar:
@mdamour1976 Please Sir!

PR #2130, opened 2016-02-04
[PDI-14742] - Mapping (Sub-transformations) step fails when executing…

andrei-viaryshka:
… on remote carte slave server

Password is needed in order to connect to the repository from slave Carte server.


andrei-viaryshka:
@hudak @deinspanjer @mbatchelor Could you please review?

PR #2135, opened 2016-02-05
[PDI-14854] Excel Writer NPE when no rows

matthewtckr:
Generate an empty RowMeta, as some settings require an empty file be created before finishing.

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/excelwriter/ExcelWriterStep.java
+++ b/engine/src/org/pentaho/di/trans/steps/excelwriter/ExcelWriterStep.java
@@ -51,6 +51,7 @@ import org.apache.poi.xssf.usermodel.XSSFWorkbook;
 import org.pentaho.di.core.Const;
 import org.pentaho.di.core.ResultFile;
 import org.pentaho.di.core.exception.KettleException;
+import org.pentaho.di.core.row.RowMeta;
 import org.pentaho.di.core.row.ValueMeta;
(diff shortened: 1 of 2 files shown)


engine/src/org/pentaho/di/trans/steps/excelwriter/ExcelWriterStep.java
```diff
@@ -89,8 +90,13 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
     if ( first ) {
 
       first = false;
-      data.outputRowMeta = getInputRowMeta().clone();
-      data.inputRowMeta = getInputRowMeta().clone();
+      if ( r == null ) {
```
mbatchelor: Don't think this belongs here. If you look at other steps, the pattern is more like:

Object[] r = getRow();
if  ( r == null ) {
  // no more input to be expected
  setOutputDone();
  return false;
}
...
if ( first ) {

matthewtckr: I would agree for most other steps, but there is some functionality that would allow for an "empty" (only header row) file to be created, even if no data rows are received by the step.

By placing this here, instead, it will allow the empty file to be created, if configured, and then drop down to line 197/203 to close out the file.

There's still an integration test needed for this use-case, which I'll add shortly.

matthewtckr:
@mbatchelor Added integration test

PR #2144, opened 2016-02-08
[BACKLOG-6392] - Changed LazyUnifiedRepositoryDirectory so that hidde…

wseyler:
…n files are NOT filtered.


wseyler:
@mdamour1976

PR #2146, opened 2016-02-08
[PDI-10546] Fixing unit test failures due to missing test-config

mdamour1976:


mdamour1976:
@AlexanderBuloichik for reference

PR #2149, opened 2016-02-09
[PDI-14749] Greenplum loader step does not handle TYPE_TIMESTAMP

SergeyTravin:


SergeyTravin:
@brosander, @deinspanjer, review it please

PR #2162, opened 2016-02-11
[SP-2462][PDI-14749] Greenplum loader step does not handle TYPE_TIMESTAMP

SergeyTravin:


SergeyTravin:
@brosander, @deinspanjer,  @pamval  review it please

PR #2165, opened 2016-02-11
[BACKLOG-6496] - Database lookup step with "IS NOT NULL" comparator fails on 6.1

akhayrutdinov:
- check if lookup value really exists
- add a test

@brosander, review it please. This PR fixes another problem in the step.


akhayrutdinov:
@brosander, review it please. @deinspanjer, may be you have a chance to look at it?

PR #2181, opened 2016-02-12
[BACKLOG-6573] Regression with Metadata injecting the Excel input step

e-cuellar:
@mdamour1976 Please Sir!


e-cuellar:
@mdamour1976 Also... please advice on the comment I left in the Jira case and see the attached screenshoot.

PR #2184, opened 2016-02-14
[PDI-14953] Add metadata injection support to Greenplum load step

stevewillcock:
This is to add metadata injection to the Greenplum load step. I have added a unit test with a similar structure to that in TextFileOutputMetaInjectionTest.

Please let me know if there are any problems with this.

Thanks


jbleuel:
This one is outdated, see PDI-14953 for details. Nevertheless thanks for the PR!

stevewillcock:
@jbleuel  Hi, no problem, glad this has been fixed. Just to point out, this PR also used annotations, and it's interesting to see that the implementations are very similar (at least the approach I had taken was correct!) :)

https://github.com/pentaho/pentaho-kettle/pull/2184/files
https://github.com/pentaho/pentaho-kettle/pull/2349/files

I'll probably create another PR in the near future to include the properties file additions from this PR (which can be quite important for end user understanding of allowed values when using metadata injection in some cases including in this case I think) and the extra tests which are in this version.

On a side note, do you know if support for annotation based metadata injection will be included in the 6.1 release or will it not be available until 7.0?

Thanks!

Steve

jbleuel:
Yes, 6.1 supports annotations for metadata injection. Once 6.1 is out we will also provide dev doc on this on help.pentaho.com.

stevewillcock:
Brilliant, thanks for the clarification.

PR #2208, opened 2016-02-19
[BACKLOG-6682] fix injectable trim type for excel input field

tgf:
@mdamour1976 please review


tgf:
@bmorrise please review

PR #2217, opened 2016-02-23
[BACKLOG-6753] Change MDI behavior for non-exist steps, fields and key to don't fail transformation

AlexanderBuloichik:


AlexanderBuloichik:
@bmorrise Merge it, please

PR #2220, opened 2016-02-23
[BACKLOG-6753] Change MDI behavior for non-exist steps, fields and key to don't fail transformation

mdamour1976:


mdamour1976:
@bmorrise

PR #2257, opened 2016-02-29
[PDI-14403] Salesforce Error Message

matthewtckr:
Show real username when testing the Salesforce connection


matthewtckr:
@pamval Created #2277

PR #2260, opened 2016-03-01
[PDB-1427] Default date formats in dashboard and PIR are not same

SergeyTravin:
- added date format with and HH:mm:ss.S
 - added method which return all available date formats
 - removed some dublicate date format


SergeyTravin:
@tkafalas, @rmansoor, @akhayrutdinov  Could you review it?

PR #2292, opened 2016-03-10
[PDI-15085] Moved XML Input Stream step to Input category

bmorrise:
@mdamour1976


mdamour1976:
@kcruzada @bmorrise I wouldn't have merged that, I was just reviewing it.

I think we missed this:
https://github.com/bmorrise/pentaho-kettle/blob/79fb4c49276035a33b18b96f1bef650af829efc2/plugins/kettle-xml-plugin/src/org/pentaho/di/trans/steps/getxmldata/messages/messages_en_US.properties#L7

PR #2322, opened 2016-03-16
[PDI-15057]  correcting arraycopy destination in

mkambol:
GetFilesRowCountMeta.clone()

@pamval


lucboudreau:
Base build failed with return code 1

lucboudreau:
Base build failed with return code 1

PR #2344, opened 2016-03-21
[PDI-14852] - fix issue with setting variables in modified javascript step

andtorg:
Implemented two loops to climb up to the right VariableSpace level when user wants to set variables either at parent job or at grandparent job level


andtorg:
The step needs a better fix because it doesn't handle the case of submapping transformation when setting variables. I'll going to re-open or issue a new pull request when refactoring is finished,

PR #2390, opened 2016-03-31
[PDI-15152] GA Filter allows empty variable values

matthewtckr:
Check if the variable-substituted filter is not empty before trying to set a filter


eliofreitas:
Rejected from automatic build ,authorized users only

PR #2430, opened 2016-04-15
Revert "[PDI-10600] - Merge Join breaks metadata of sub-transformation"

AliaksandrShuhayeu:
This reverts commit 69cfe50dad7e0ab86d43baa4c4fa2c6ce811069c.
Fixing problem Merge Join breaks metadata of sub-transformation http://jira.pentaho.com/browse/PDI-10600 
Regression http://jira.pentaho.com/browse/BACKLOG-4404  is not reproducible now.
@brosander @mchen-len-son Could you please review?

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/mappinginput/MappingInputMeta.java
+++ b/engine/src/org/pentaho/di/trans/steps/mappinginput/MappingInputMeta.java
@@ -242,23 +242,24 @@ public class MappingInputMeta extends BaseStepMeta implements StepMetaInterface 
       // this gets set only in the parent transformation...
       // It includes all the renames that needed to be done
       //
-      if ( selectingAndSortingUnspecifiedFields ) {
-
-        // First rename any fields...
-        if ( valueRenames != null ) {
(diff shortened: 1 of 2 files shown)


eliofreitas:
Hello, This is an automated comment from our build system. Since this pull request originated from a user  not in the Pentaho organization, we will not proceed  with automatic validation of the pull request.  However, rest assured that it will be reviewed manually as soon  as we get the chance. Thanks for your contribution!

brosander:
@AliaksandrShuhayeu will this regress PDI-10600?

mchen-len-son:
@brosander 
let me clarify a bit:

1. Initially PDI-10600 was fixed with exaclty this commit
2. Then regression was found and reported in BACKLOG-4404
3. The first commit was reverted.
4. Now Aliaksandr tried to investigate the regression causes, but with no success - something has changed, and it seems too hard to understand what exactly it was.
5. So he's just reverted back to the state after 1st step.

Does that make sense?

The changes itself look correct to me.

PR #2444, opened 2016-04-21
[PDI-15202] Enable cache on "IS NULL" condition (DatabaseLookup step)

green2k:
Updated DatabaseLookup step -> enabled cache when "IS NULL" condition is used.


eliofreitas:
Hello, This is an automated comment from our build system. Since this pull request originated from a user  not in the Pentaho organization, we will not proceed  with automatic validation of the pull request.  However, rest assured that it will be reviewed manually as soon  as we get the chance. Thanks for your contribution!

mbatchelor:
Thank-you so much for your contribution!

PR #2464, opened 2016-05-02
[Test] - JsonInputTest was adapted to work with .vfs.sftp.userDirIsRo…

YuryBY:
…ot property

@brosander, @pamval, could you please take a look? -  

the test must take into account new property: https://github.com/pentaho/pentaho-kettle/commit/69998716a29f0b5ee45f5d54047da729cc2b3320#diff-431f1d388ffb40a6210f6366764b3adbR1061


pamval:
Why did we lose so much coverage on InputErrorHandler ?

PR #2473, opened 2016-05-05
[PDI-15181] Delete Files job entry will delete all files under the install directory when file location parameter is blank

IvanNikolaychuk:
**General description**
- Populate data before processing files.
- Checking whether the file path is blank before processing the file. Relative paths are permitted, and providing an empty path means deleting all files inside a root pdi-folder.
  It is much more likely to be a mistake than a desirable action, so we don't delete anything. We also don't consider this to be an error (see Jens comment about it, and just log about it if log is detailed).
- Adding some new messages.
- Refactoring (encapsulating  JobEntryDeleteFiles data,  editing some field names, simplifying some checks)
- Tests written
- Checkstyle applied

**More details about implementation**
 For job execution path to files and file masks should be provided.  These values can be obtained in two ways:
1. As an argument of a current job entry
2. As a table, that comes as a result of execution previous job/transformation.

 In the previous implementation we were checking the way we obtain data in if-else block, and were calling process function 2 times for each of the possible ways. Don't see any reason for duplicating this method calls, as the way we process the data doesn't depend on how we obtain it.
 So, I decided to separate the logic of obtaining the data in a populate method and represent populated data as a map.

*Advantages*
- Dividing a logic of obtaining arguments and wildcards and logic of processing this data ~ encapsulating logical of obtaining data in a separate method
- Treating file paths and  wildcards as a map and getting rid from working with two arrays, that is much less error-prone (BTW, I would rather refactor the class itself to have a map inside, instead of two arrays, but don't want to do such changes without a given permissions)
- More readable and maintainable code (in my opinion)

*Disadvantages*
- We are doing some extra work, because we first obtaining data and then processing it, instead of doing it at once. In fact we have 2 * N against N.
- Some extra memory for providing a map.

 I think that advantages outweigh, because in case of this step N (number of paths and wildcards) is not expected to be big. So the performance won't suffer.

### DIFF
--- a/engine/src/org/pentaho/di/job/entries/deletefiles/JobEntryDeleteFiles.java
+++ b/engine/src/org/pentaho/di/job/entries/deletefiles/JobEntryDeleteFiles.java
@@ -22,12 +22,17 @@
 
 package org.pentaho.di.job.entries.deletefiles;
 
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import org.pentaho.di.core.exception.KettleValueException;
 import org.pentaho.di.job.entry.validator.AbstractFileValidator;
 import org.pentaho.di.job.entry.validator.AndValidator;
 import org.pentaho.di.job.entry.validator.JobEntryValidatorUtils;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


IvanNikolaychuk:
@brosander , could you review it please?

PR #2477, opened 2016-05-06
[PDI-15181] Delete Files job entry will delete all files under the install directory when file location parameter is blank

IvanNikolaychuk:
- Before processing file, first checking whether it is a kettle variable, extracting it, and then checking if it is an empty value.
- Small refactoring of filed names in processFile method

Related to https://github.com/pentaho/pentaho-kettle/pull/2473


IvanNikolaychuk:
There is a drawback in https://github.com/pentaho/pentaho-kettle/pull/2473. 
There could be a case, where filePath is defined by a kettler-variable and a value under this variable is blank. So:

- Before processing file, first checking whether it is a kettle variable, extracting it, and then checking if it is an empty value. 
- Small refactoring of filed names in processFile method.

@brosander , could you please review this PR?

PR #2490, opened 2016-05-11
[BACKLOG-7661] Updated thin dialog to be modal

bmorrise:


bmorrise:
@ddiroma @mdamour1976

mdamour1976:
Can we pass in a modal flag instead?  Might be more generally useful.  Just a thought to consider.

bmorrise:
@mdamour1976 Updated.

PR #2497, opened 2016-05-13
[SP-2676] Backport of PDI-15118 - User with no schedule permissions has access to schedule perspective (6.1 Suite)

IvanNikolaychuk:
- When logging to repository check whether user is allowed to schedule content and remove "Schedule perspective" it he has no permissions for this.
- provided an ability to manipulate perspective's UI list from SpoonPerspectiveManager.
- Tests written
- Checkstyle applied

If I understand the idea of PerspectiveInitializer class, it's goal was to make lazy initialization for all perspectived, except the first one, that would be inited immediately. Other perspectives would be initialized when needed and removed from initializerMap.

So, PerspectiveInitializer encapsulates the list of perspectives, and is the only place where perspective list can be accessible and changed.

I added few more methods there: for adding and removing perspective names (unfortunately there is no oppotunity to 'hide' element in CCombo class, so we are removing and adding), keeping the existing behavior the same:
1. In first init only main perspective is initialized.
2. Each perspective is initialized only once, regarding how many times it was active.

### DIFF
--- a/plugins/pdi-pur-plugin/src/org/pentaho/di/repository/pur/PurRepositoryConnector.java
+++ b/plugins/pdi-pur-plugin/src/org/pentaho/di/repository/pur/PurRepositoryConnector.java
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2010 - 2015 Pentaho Corporation.  All rights reserved.
+ * Copyright 2010 - 2016 Pentaho Corporation.  All rights reserved.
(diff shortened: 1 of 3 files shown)


IvanNikolaychuk:
@brosander , could you please review this PR?
It's a backport of https://github.com/pentaho/pentaho-kettle/pull/2437

PR #2501, opened 2016-05-13
[TEST] - Added clone and Load/Save testers

mbatchelor:


mbatchelor:
Note - my code builds, it's Head that doesn't. Checkstyle run locally all passes.

PR #2504, opened 2016-05-16
[BACKLOG-8011] - Added expand/collapse shell to takeover transformation tab

wseyler:
…iew menu that will show a generic browser when selected while a transformation is open.  This menu item isn't part of the final product and is just for QA testing. @bmorrise @mdamour1976.


mdamour1976:
@bmorrise @wseyler Let's try to make this even more generic, I'll ping you guys privately for details.  This work would easily satisfy the result of a spike, but we need to figure out the boundary between Spoon and the plugin.

wseyler:
@bmorrise @mdamour1976 Made these even more generic.  Now we only refer to a transformation browser.  I think this makes sense since the way this browser works is that it has to be parented to a TransGraph.

PR #2514, opened 2016-05-17
[TEST] Allow meta changes during LoadSaveTester

matthewtckr:
Some steps require calling allocate() before calling some setter methods.
Pass the object back to a user-defined Initializer class, in order for the allocate() or other methods to be called as needed.


matthewtckr:
It looks like `pentaho-reporting-engine-classic-core` brings in asm-3.2, causing the RestIT test failure as it doesn't support Java 8.  We might need to update to asm 5.0 or later to get JDK 8 support ([ASM History](http://asm.ow2.org/history.html)).

PR #2515, opened 2016-05-18
[TEST] - Add load/save test and checkstyle fix

mbatchelor:


lucboudreau:
Had to interrupt Wingman. Will trigger manually after the update.

PR #2523, opened 2016-05-18
[BACKLOG-7972] - Hadoop File Output: Hadoop Clusters dropdown doesn't…

vasilina-terehova-epam:
… preserve selected cluster after reopen a transformation

return api in TextFileInputMeta to have metastore for loadSource method, needed for HadoopFileInputMeta, adaptate interface to use StringBuilder instead of StringBuffer as in engine, write tests to fail on api changed in engine again

update version for jersey to 1.19.1 due to incompatibility jersey before 1.19 with java 1.8(was reproduced with fail )


vasilina-terehova-epam:
@dkincade @hudak could you take a view?
could you please send logs with reasons of build failure

dkincade:
This is failing because it can not find the source folder in the assembly module. This is a known issue with wingman. I am going to include the pull request.

PR #2524, opened 2016-05-18
[PDI-14852] setVariable in Javascript Step with scope of parent job does not work

IvanNikolaychuk:

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/scriptvalues_mod/ScriptValuesAddedFunctions.java
+++ b/engine/src/org/pentaho/di/trans/steps/scriptvalues_mod/ScriptValuesAddedFunctions.java
@@ -79,6 +79,7 @@ import org.pentaho.di.core.row.RowMetaInterface;
 import org.pentaho.di.core.util.EnvUtil;
 import org.pentaho.di.core.variables.VariableSpace;
 import org.pentaho.di.core.vfs.KettleVFS;
+import org.pentaho.di.trans.Trans;
 import org.pentaho.di.trans.step.StepInterface;
 import org.pentaho.di.trans.steps.loadfileinput.LoadFileInput;
 
@@ -106,6 +107,15 @@ public class ScriptValuesAddedFunctions extends ScriptableObject {
     "moveFile", "execProcess", "isEmpty", "isMailValid", "escapeXml", "removeDigits", "initCap",
     "protectXMLCDATA", "unEscapeXml", "escapeSQL", "escapeHtml", "unEscapeHtml", "loadFileContent",
     "getOcuranceString", "removeCRLF" };
+
+
+  enum VariableScope {
+    SYSTEM,
+    ROOT,
+    PARENT,
+    GRAND_PARENT
+  }
+
 
   // This is only used for reading, so no concurrency problems.
   // todo: move in the real variables of the step.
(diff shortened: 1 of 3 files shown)


IvanNikolaychuk:
@brosander, can you please take a look at this PR?

IvanNikolaychuk:
@bmorrise , could you review this PR please?

bmorrise:
Yes, I'll look at it today.

IvanNikolaychuk:
Thanks :)

engine/src/org/pentaho/di/trans/steps/scriptvalues_mod/ScriptValuesAddedFunctions.java
```diff
@@ -2580,7 +2605,7 @@ public static String execProcess( Context actualContext, Scriptable actualObject
         while ( ( ligne = br.readLine() ) != null ) {
           buffer.append( ligne );
         }
-        // if (processrun.exitValue()!=0) throw Context.reportRuntimeError("Error while running " + ArgList[0]);
+        // if (processrun.exitValue()!=0) throw Context.reportRuntimeError("Error while running " + arguments[0]);
```
bmorrise: Is this commented line needed?

engine/src/org/pentaho/di/trans/steps/scriptvalues_mod/ScriptValuesAddedFunctions.java
```diff
@@ -2641,12 +2666,12 @@ public static Boolean isMailValid( Context actualContext, Scriptable actualObjec
...
         }
 
-        retval = Boolean.TRUE;
+        isValid = Boolean.TRUE;
 
       } catch ( Exception e ) {
         throw Context.reportRuntimeError( "Error in isMailValid function: " + e.getMessage() );
       }
-      return retval;
+      return isValid;
```
bmorrise: It looks like isValid can potentially be null. Is this acceptable for code that calls this?

engine/src/org/pentaho/di/trans/steps/scriptvalues_mod/ScriptValuesAddedFunctions.java
IvanNikolaychuk: Looks like not, but this comment wasn't brought by me, so I prefered not to touch it.

bmorrise: That's fine.

engine/src/org/pentaho/di/trans/steps/scriptvalues_mod/ScriptValuesAddedFunctions.java
IvanNikolaychuk: You are right, it can.

But It's just a place where I refactored a variable name to camel case style (with no intention to change logic)
Actually there are bunch of such drawbacks in this class (it's ~ 3 k. lines length), so I decided just to make sure that my code is safe, and refactored a few things that strike the eyes without changing any logic.

So, I highly recomend to refactore this class and to make it more safe, but I would rather do this in context of the hole class, but not just in this method. But I need time and aggrement for this :)

IvanNikolaychuk:
@bmorrise , so, is it ok to merge it now?

PR #2526, opened 2016-05-18
[ENGOPS-1918] Update Commons-DBCP in Core

matthewtckr:
COMP-431


dkincade:
Are we planning on upgrading all the rest of commons-dbcp across the suite?

matthewtckr:
This is just for the use-case of embedding PDI into a Java project using Maven, which caused a conflict.  I'll take a look at the other projects as well, as it appears to be sprinkled across Mondrian and Reporting Engine, among a few others.

mbatchelor:
Important note - in the final BA/DI assemblies, DBCP-1.4 is already there in the WEB-INF/lib (even as far back as 4.8.3). This really just makes sure we're compiling against what we already run with. On that basis alone, this is a no-brainer.

PR #2542, opened 2016-05-20
[BACKLOG-8421] Added MDI support to Get Variable step

jmwoliver:


jmwoliver:
@mdamour1976 @bmorrise

PR #2546, opened 2016-05-23
[TEST] - Added more load/save tests

mbatchelor:


mbatchelor:
More commits to come including fix of License Header for MailMeta. Needed the above tests in the nightly.

PR #2547, opened 2016-05-23
[BACKLOG-8265] Changes in engine to support MDI to the Hadoop File Input

e-cuellar:


e-cuellar:
@bmorrise Please Sir!

engine/src/org/pentaho/di/core/injection/bean/BeanInjectionInfo.java
```diff
@@ -162,7 +176,21 @@ public String getName() {
     }
 
     public String getDescription() {
-      return BaseMessages.getString( clazz, clazzAnnotation.localizationPrefix() + name );
+      String description = BaseMessages.getString( clazz, clazzAnnotation.localizationPrefix() + name );
```
bmorrise: Since it looks like these two getDescription() methods are the same does it make sense to create a parent class with the getDescription() method that these two classes can inherit?

PR #2549, opened 2016-05-23
[BACKLOG-7972] - jersey version back to 1.16 due to using this versio…

vasilina-terehova-epam:
…n on di ba server
integration tests will now fail with old jersey version but
more investigation will be done in spike http://jira.pentaho.com/browse/BACKLOG-8388


rmansoor:
wingman is failing for weird reasons. I am processing this pull request, so we have a good build tomorrow

PR #2571, opened 2016-05-31
[PDI-15210] Made the Collections.synchornizedList(...) a CopyOnWriteA…

ccaspanello:
…rrayList so that it is now threadsafe throughout the stack.  Also added a test fromt he Crate perspective that passes the threadsafe list into the WebServer.  @hudak Can you peer review this.  Please and thank you.

### DIFF
--- a/engine/src/org/pentaho/di/www/WebServer.java
+++ b/engine/src/org/pentaho/di/www/WebServer.java
@@ -63,6 +63,8 @@ import java.util.Timer;
 import java.util.TimerTask;
 
 public class WebServer {
+  private static final String DEFAULT_DETECTION_TIMER_PROPERTY = "pentaho.crate.detectionTimer";
+  private static final int DEFAULT_DETECTION_TIMER = 20000;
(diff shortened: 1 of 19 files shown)


engine/src/org/pentaho/di/www/WebServer.java
```diff
@@ -63,6 +63,8 @@
 import java.util.TimerTask;
 
 public class WebServer {
+  private static final String DEFAULT_DETECTION_TIMER_PROPERTY = "pentaho.crate.detectionTimer";
```
lucboudreau: This should be documented somewhere.

matthewtckr: Can we switch `crate` for `carte`?

hudak: @lucboudreau maybe here? https://help.pentaho.com/Documentation/5.3/0P0/0U0/050/060/010#Setting_the_Jetty_Server_Parameters_in_the_kettle.properties_file

PR #2572, opened 2016-05-31
[PDI-13748] Hide Template Sheet option

matthewtckr:
Add a UI setting to enable hiding the Template sheet after the sheet for output is created


mbatchelor:
You didn't update ExcelWriterStepMetaTest to include the new attribute in the load/save tester.

matthewtckr:
Good catch.  Updated the PR.

PR #2574, opened 2016-06-01
[SP-2694] Backport of PDI-14852 - setVariable in Javascript Step with scope of parent job does not work (6.1 Suite)

IvanNikolaychuk:

It's a backport of https://github.com/pentaho/pentaho-kettle/pull/2524


IvanNikolaychuk:
@mchen-len-son , could you please review it?
It's a backport of https://github.com/pentaho/pentaho-kettle/pull/2524

mchen-len-son:
Checktyle violation contain only JUnit and Mockito static import errors.
Code coverage hasn't been dropped.
Merging.

PR #2575, opened 2016-06-01
[SP-2695] Backport of PDI-14852 - setVariable in Javascript Step with scope of parent job does not work (5.4 Suite)

IvanNikolaychuk:

@mchen-len-son, @pamval  could you please take a look at this PR? It's a backport of https://github.com/pentaho/pentaho-kettle/pull/2524


mchen-len-son:
Checktyle violation contain only JUnit and Mockito static import errors.
Code coverage hasn't been dropped.
Merging.

PR #2581, opened 2016-06-03
[PDI-15118] User with no schedule permissions has access to schedule perspective

IvanNikolaychuk:
- Removing asynchronous call to UI components, when connecting to PurRepository.
- Moving the logic of schedule perspective updates to EESpoonPlugins (will be triggered on each repo-connect/disconnect event).
- Test written
- Checkstyle applied


IvanNikolaychuk:
@mchen-len-son , could you take a look at this PR please?

PR #2582, opened 2016-06-03
[BACKLOG-8241] - Move the writing out of the Metadata-Injected KTR ...

rfellows:
…from before execution to after to allow for customizations to MDI to happen in a step's init() method. This allows for complex data structures to be handled in a manual fashion when traditional annotation-based MDI isn't sufficient.


rfellows:
@AlexanderBuloichik @bmorrise please review

AlexanderBuloichik:
Looks good

PR #2584, opened 2016-06-03
[BACKLOG-8520] Data Service with Join Rows (cartesian product) step f…

denis8703:
We can get NPE if temporary files are null

### DIFF
--- a/engine/test-src/org/pentaho/di/trans/steps/joinrows/JoinRowsTest.java
+++ b/engine/test-src/org/pentaho/di/trans/steps/joinrows/JoinRowsTest.java
@@ -1,0 +1,93 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
+ * Copyright (C) 2002-2016 by Pentaho : http://www.pentaho.com
+ *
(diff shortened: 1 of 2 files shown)


engine/test-src/org/pentaho/di/trans/steps/joinrows/JoinRowsTest.java
```diff
@@ -0,0 +1,83 @@
...
+  @Test
+  public void dataFileIsNull() throws Exception {
+    getJoinRows().dispose( meta, data );
+  }
+
+  @Test
+  public void disposeDataFiles() throws Exception {
+    data.file = new File[]{ Mockito.mock( File.class ), Mockito.mock( File.class ) };
+    getJoinRows().dispose( meta, data );
+  }
```
mkambol: @denis8703  Would be good to actually verify that something happens when the .dispose call is made-- i.e. that the files are deleted.

And with the dataFileIsNull() test above, you're really just testing that no NPE is thrown.  So why not check for that explicitly?  And maybe name your test something that reflects what it's testing for, or add a comment?

PR #2588, opened 2016-06-03
[TEST] - ignoring some recently added tests that fail in known condit…

kurtwalker:
…ions.  Tests should be revisited and turned back on if appropriate.


kurtwalker:
@mbatchelor ignoring a couple of the tests you've committed recently as they fail intermittently .

@rfellows @DFieldFL @pminutillo please review and merge.  last commit change that hasn't wing-manned yet is removing the unused import reported in the last run

kurtwalker:
@matthewtckr tagging you on this also as it looks like testSerialization is using a test class that you have also changed recently

PR #2589, opened 2016-06-03
[BACKLOG-8011] Added expand/collapse shell to takeover transformation…

wseyler:
… tab


wseyler:
@mdamour1976 fixed checkstyle problems

ui/src/org/pentaho/di/ui/spoon/ExpandedContentManager.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * maximizeBrowser
+   * 
+   * @param browser
+   *          the browser object to maximize. We try to take up as much of the Spoon window as possible.
+   */
+  private static void maximizeExpandedContent( Browser browser ) {
+    SashForm sash = (SashForm) browser.getParent().getParent().getParent().getParent();
```
bmorrise: Can we add a getter to spoon to return the SashForm so we don't have to do all these getParent() calls?

wseyler:
@mdamour1976 @bmorrise Added support for retaining sash size on a per tab instance.  Removed fragile "getParent()" code.  Fixed problem with tabDeselected events.  Added code so that full screen sashforms weight won't be saved when spoon closes (as per conversation with Mike).

PR #2576, opened 2016-06-01
[PDI-15203] Made the Collections.synchornizedList(...) a CopyOnWriteA…

ccaspanello:
…rrayList so that it is now threadsafe throughout the stack.  Also added a test fromt he Crate perspective that passes the threadsafe list into the WebServer.

@hudak Here's the pull request

PR #2590, opened 2016-06-05
[SP-2705][PDI-15210] Made the Collections.synchornizedList(...) a Cop…

mchen-len-son:
…yOnWriteArrayList so that it is now threadsafe throughout the stack.  Also added a test fromt he Crate perspective that passes the threadsafe list into the WebServer.

### DIFF
--- a/core/src/org/pentaho/di/core/Const.java
+++ b/core/src/org/pentaho/di/core/Const.java
@@ -605,6 +605,11 @@ public class Const {
   public static final String KETTLE_JOB_LOG_SCHEMA = "KETTLE_JOB_LOG_SCHEMA";
 
   /**
+   * The name of the variable that defines the timer used for detecting slave nodes.
+   */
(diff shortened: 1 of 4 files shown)


mchen-len-son:
@ccaspanello @bmorrise @e-cuellar please review
This is a backport of #2576

pamval:
There's a big drop in coverage on KettleLogStore.new TimerTask(). Is that expected, somehow?

mchen-len-son:
@pamval 
The log cleaning TimerTask, where the coverage has been dropped, is scheduled to be executed every 10s.
And new Carte test has CountDownLatch waiting 10s, and then exits.

I think, on master branch the log cleaning task executed BEFORE the test exit.
But this time the test exited BEFORE the task started execution.
Thus, we have coverage drop there.

I'll try to make the test more stable from that point of view.

PR #2592, opened 2016-06-06
[PDI-15032] Metadata Injection on all tabs of select values step is causing error - NPE

AndreyBurikhin:
@matthewtckr , @bmorrise please review.

@matthewtckr I added parameters validation for setters, so client code could not set null values.
In other cases the arrays have non-null values. Please, let me know if you see other possible issues.


matthewtckr:
Suggest #2597 instead, as we don't want to throw NPEs.

AndreyBurikhin:
We'll use https://github.com/pentaho/pentaho-kettle/pull/2597.

PR #2595, opened 2016-06-07
[BACKLOG-8208] Add MDI support to Append Streams step

AlexanderBuloichik:


dkincade:
@bmorrise @e-cuellar @mdamour1976 Looks like WingMan is not happy with some of the minor decreases in complexity and branch coverage changes. Please review for content.

PR #2598, opened 2016-06-07
[BACKLOG-8454] Spoon: "User Defined Java Expression" step fails

matthewtckr:
Cast an "Integer" return value to Long, to match the expected type of a ValueMetaInteger


matthewtckr:
[Original logic](https://github.com/matthewtckr/pentaho-kettle/blob/6.0.0.0-R/engine/src/org/pentaho/di/trans/steps/janino/Janino.java#L228-L234)

matthewtckr:
Wingman doesn't support PRs with changes in the /test directory.  This will either need a manual review, or moving it into the unit tests for the engine module.

PR #2601, opened 2016-06-08
[SP-2676] Backport of PDI-15118 - User with no schedule permissions has access to schedule perspective (6.1 Suite)

IvanNikolaychuk:
- Removing asynchronous call to UI components, when connecting to PurRepository.
- Moving the logic of schedule perspective updates to EESpoonPlugins (will be triggered on each repo-connect/disconnect event).
- Test written
- Checkstyle applied

There are two differs from fix in master:
1. Perspectives in 6.1 are build by "CCombo" class, that throws an error, when removing an element, that doesn't exist their. So, we check if an element exists before removing.
2. On master, after disconnecting the repo, Schedule perspective becomes disabled. To be consistent in behavior, I added this disabling in 6.1 too, by changing perspective visibility on "repository disconnect" event.


IvanNikolaychuk:
@mchen-len-son , could you please take a look at this PR?
It's a backport of https://github.com/pentaho/pentaho-kettle/pull/2581

PR #2603, opened 2016-06-08
[PDI-15415] Always give back result row does not work in Memory Group by

matthewtckr:


matthewtckr:
Wingman is failing as it's trying to grab the top-level build file (because of the integration test), which isn't supported currently.

engine/src/org/pentaho/di/trans/steps/memgroupby/MemoryGroupBy.java
```diff
@@ -75,16 +75,13 @@ public MemoryGroupBy( StepMeta stepMeta, StepDataInterface stepDataInterface, in
...
   }
 
+  @Override
   public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws KettleException {
     meta = (MemoryGroupByMeta) smi;
     data = (MemoryGroupByData) sdi;
 
     Object[] r = getRow(); // get row!
     if ( first ) {
-      if ( r == null ) {
```
mbatchelor: Shouldn't the change only happen if meta.isAlwaysGivingBackOneRow() is true ? Seems to me this needs to go back as:

if ( ( r == null ) && ( !meta.isAlwaysGivingBackOneRow() ) ) { ...

matthewtckr:
It works either way, but I agree from a readability/simplicity perspective that it should be there.

PR #2604, opened 2016-06-08
PDI-14853 - add remove enclosure property

stanislau-strelchanka:


stanislau-strelchanka:
@mbatchelor, @kolinus, @pamval , @pmalves, @matthewtckr added changes according to recent email discussion. Please review.

mbatchelor:
Concerned about the new LoadSaveTest failure, but I don't suspect this is the culprit. I'm inclined to process the PR. Any objections ?

jbleuel:
I would call out (in the description and variable naming) that this only effects the Split Fields step – otherwise people could think it effects enclosures all over the place.
Proposal for the variable naming:
KETTLE_SPLIT_FIELDS_REMOVE_ENCLOSURE
Proposal for the description: 
Set this variable to false to preserve enclosure symbol after splitting the string **in the Split fields step**. Changing it to true will remove first and last enclosure symbol from the resulting string chunks.

stanislau-strelchanka:
@jbleuel , @mbatchelor updated variable and description. Investigating test failure.

stanislau-strelchanka:
@mbatchelor, @kolinus, @pamval , @pmalves, @matthewtckr looks like some problems reading 'dateFormatLocale' attribute. Is there a chance to get full log file of this error? Thanks.

PR #2644, opened 2016-06-22
[BACKLOG-8827] design-time check and remove invalid target keys

tgf:
@bmorrise please review


tgf:

PR #2647, opened 2016-06-23
[PDI-6748] - Mondrian Input Step: pan.sh / kitchen.sh ignore mondrian…

YuryBY:
….properties

@CodeOnCoffee, @mbatchelor, here is the most relevant mondrian.properties since we don't know what exactly customers will need. Any objections? Thanks.


YuryBY:
@CodeOnCoffee, @mbatchelor, could you please check it manually, since on my environment all checks were passed. Thanks.

PR #2651, opened 2016-06-23
[PDI-15440] NPE when creating error hop

matthewtckr:


lucboudreau:
I don't think these `@Override` annotations are necessary.

PR #2657, opened 2016-06-23
[BACKLOG-8272] Added MDI Support to Multiway Merge Join

jmwoliver:


jmwoliver:
@mdamour1976

kurtwalker:
@mdamour1976 @jmwoliver This change broke meta-injection for the refinery steps.  We get a NPE at MetaInject:121.  Simple fix is to wrap in an if != null, but I don't really know the purpose behind this change, so don't want to commit it.  Would really love to have something in by eod so that QA can do a complete test of our new stuff

kurtwalker:
Also see MetadataInjectionModelTest.testJobEntry in the pentaho-data-refinery project.  It has a failure due to this change

PR #2660, opened 2016-06-27
[PDI-14948] Transformation with Metadata Injection runs endlessly if no input is received by Injection Step

IvanNikolaychuk:
- When starting threads, check if transformation has any step to run, and finish it, if it has no steps.
- Test written
- Checkstyle applied

### DIFF
--- a/engine/src/org/pentaho/di/trans/Trans.java
+++ b/engine/src/org/pentaho/di/trans/Trans.java
@@ -1442,6 +1442,10 @@ public class Trans implements VariableSpace, NamedParams, HasLogChannelInterface
 
     heartbeat = startHeartbeat( getHeartbeatIntervalInSeconds() );
 
+    if ( steps.isEmpty() ) {
+      fireTransFinishedListeners();
+    }
+
     if ( log.isDetailed() ) {
       log.logDetailed( BaseMessages.getString( PKG, "Trans.Log.TransformationHasAllocated", String.valueOf( steps
           .size() ), String.valueOf( rowsets.size() ) ) );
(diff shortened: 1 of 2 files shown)


pamval:
Are we able to understand the impacts of this change? This touches the base class for the Transformation and changes its behavior.

pamval:
@kurtwalker @mdamour1976 @e-cuellar @tkafalas @tgf (and everyone else at R2D2), can you please review this?

kurtwalker:
Change was originally my suggestion.  I think it's safe, but I was worried about use cases I may have been unaware of.  Suggest thorough regression test

IvanNikolaychuk:
@pamval, @kurtwalker, well, initially it's Kurt who suggested this fix.
I did some regression testing and covered some cases that I'm aware of, contacted to testers that work with PDI closely and asked my dev-colleagues about this changes.
It looks safe for me, but I would also prefer this to be reviewed by as much devs as possible.

IvanNikolaychuk:
@AndreyBurikhin , @mchen-len-son , could you please take a look at this PR?

IvanNikolaychuk:
@mdamour1976 @e-cuellar @tkafalas @tgf @bmorrise @wseyler, could you please review this PR?

mdamour1976:
@IvanNikolaychuk I think it looks safe enough, but I'm wondering about that wingman newly failed case.  I'm not sure how that is related.  Can that part be validated independent of wingman?

IvanNikolaychuk:
@mdamour1976, agree, don't see any correlation between changes and this PR. TextFileInputMetaLoadSaveTest passes localy for me either.
Thanks for review!

PR #2667, opened 2016-06-29
[PDI-15463] Execute Process Arguments

matthewtckr:


engine/src/org/pentaho/di/trans/steps/execprocess/ExecProcess.java
```diff
@@ -110,7 +128,21 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
...
 
       // execute and return result
-      execProcess( processString, processResult );
+      if ( meta.isArgumentsInFields() ) {
+        List<String> cmdArray = new ArrayList<>();
+        cmdArray.add( processString );
+
+        for ( int i = 0; i < data.indexOfArguments.length; i++ ) {
+          String argString = data.previousRowMeta.getString( r, data.indexOfArguments[i] );
+          if ( !Const.isEmpty( argString ) ) {
```
mbatchelor: Why restrict them from having empty arguments ? Seems that people do this often enough ( e.g. PDI-12510 ).

matthewtckr: Fixed in the rebased commit.  It now converts any NULL values to an empty string, to avoid an NPE from being thrown by [Runtime.exec](https://docs.oracle.com/javase/7/docs/api/java/lang/Runtime.html#exec(java.lang.String[]))

PR #2671, opened 2016-07-01
6.1

FFFFFurry:
Hi, I am very  excited to read your code with my privilege.
I think I find something in the  code segment,maybe it's a bug.
the path is :pentaho-kettle/engine/src/org/pentaho/di/trans/steps/script/ScriptMeta.java
the line number:887
the code segment:
                        String[] strings = stepName.split( "\\." );
                        String extension = strings[strings.length > 0 ? 1 : 0]; // skip the script number extension

if stepName= 'a', 
then strings.length=1
then 'strings.length > 0 ? 1 : 0' is 1
then throw java.lang.ArrayIndexOutOfBoundsException


pamval:
hello. thanks for your feedback. Would you mind opening an issue with that info at jira.pentaho.com, PDI project, please?

PR #2675, opened 2016-07-04
[PDI-15120] Export Linked Resources to XML fails if Text File Input contains empty row

IvanNikolaychuk:
- Adding null-checks before obtaining File Object from filename, that can be null.

**Summary**

After this fix behavior is the following:
- When exporting trans with empty filenames, they are kept empty and not converted to related path

**More details**

Situation is the following here:
In transformation's xml fileNames with empty strings are stored as <fileName/> and extracted as nulls when transformation is loaded.
First I wanted to store and load them as empty strings. But it was a bad idea - it leads us to some problems that are hard to overcome with such storing. Here the pitfall of this approach: when exporting trans with empty (non-null) strings in file name - they are converted into related path (root of kettle in this case). And in 99 % cases it's not what user expects.

So what we want here is exporting empty strings as they are, without converting them to related path.
We should also do this with empty strings for keeping consistency. Try running trans with textFileInput step with empty file name. It won't be treated as related path. But when exporting if with empty string it becomes related, that is not expected.

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/fileinput/text/TextFileInputMeta.java
+++ b/engine/src/org/pentaho/di/trans/steps/fileinput/text/TextFileInputMeta.java
@@ -34,6 +34,7 @@ import org.pentaho.di.core.CheckResultInterface;
 import org.pentaho.di.core.Const;
 import org.pentaho.di.core.database.DatabaseMeta;
 import org.pentaho.di.core.exception.KettleException;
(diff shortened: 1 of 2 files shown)


IvanNikolaychuk:
@mchen-len-son , could you please review it?
Coverage drops are definitely WN's errors

IvanNikolaychuk:
@bmorrise, could you please review this PR?

PR #2248, opened 2016-02-25
[PDI-5537] MemoryGroupBy Aggregation Return Types

matthewtckr:
SUM and Average should return the same type as they are aggregating, instead of always Numeric (w/ decimal) values.

PR #2676, opened 2016-07-04
[SP-2717] Backport of PDI-5537 - "Memory Group By" changes data types  of the summation fields (5.4 Suite)

IvanNikolaychuk:
SUM and Average should return the same type as they are aggregating, instead of always Numeric (w/ decimal) values.


IvanNikolaychuk:
It's partly a backport of https://github.com/pentaho/pentaho-kettle/pull/2248. Not all tests are backported because of some differs between 5.4 and 6.1/master. Thats fine, because tests that are written cover all added code completely.

Verified it localy.
@mchen-len-son , could you please revie this PR?

PR #2686, opened 2016-07-12
[PDI-8931] SlaveServer Synchronized methods need to be examined

IvanNikolaychuk:
- Test written
- Checkstyle applied

### DIFF
--- a/engine/src/org/pentaho/di/cluster/SlaveServer.java
+++ b/engine/src/org/pentaho/di/cluster/SlaveServer.java
@@ -41,6 +41,7 @@ import java.util.Random;
 
 import org.apache.commons.httpclient.Header;
 import org.apache.commons.httpclient.HttpClient;
+import org.apache.commons.httpclient.HttpState;
 import org.apache.commons.httpclient.UsernamePasswordCredentials;
 import org.apache.commons.httpclient.auth.AuthScope;
 import org.apache.commons.httpclient.methods.ByteArrayRequestEntity;
@@ -99,7 +100,7 @@ import org.w3c.dom.Document;
 import org.w3c.dom.Node;
 
 public class SlaveServer extends ChangedFlag implements Cloneable, SharedObjectInterface, VariableSpace,
-    RepositoryElementInterface, XMLInterface {
+  RepositoryElementInterface, XMLInterface {
   private static Class<?> PKG = SlaveServer.class; // for i18n purposes, needed by Translator2!!
 
   public static final String STRING_SLAVESERVER = "Slave Server";
@@ -139,41 +140,41 @@ public class SlaveServer extends ChangedFlag implements Cloneable, SharedObjectI
 
   private LogChannelInterface log;
 
-  private String name;
+  private volatile String name;
 
-  private String hostname;
+  private volatile String hostname;
 
-  private String port;
+  private volatile String port;
 
-  private String webAppName;
+  private volatile String webAppName;
 
-  private String username;
+  private volatile String username;
 
-  private String password;
+  private volatile String password;
 
-  private String proxyHostname;
+  private volatile String proxyHostname;
 
-  private String proxyPort;
+  private volatile String proxyPort;
 
-  private String nonProxyHosts;
+  private volatile String nonProxyHosts;
 
   private String propertiesMasterName;
 
   private boolean overrideExistingProperties;
 
-  private boolean master;
+  private volatile boolean master;
 
-  private boolean shared;
+  private volatile boolean shared;
 
-  private ObjectId id;
+  private volatile ObjectId id;
 
-  private VariableSpace variables = new Variables();
+  private volatile VariableSpace variables = new Variables();
 
-  private ObjectRevision objectRevision;
+  private volatile ObjectRevision objectRevision;
 
   private Date changedDate;
 
-  private boolean sslMode;
+  private volatile boolean sslMode;
 
   private SslConfiguration sslConfig;
 
@@ -224,7 +225,7 @@ public class SlaveServer extends ChangedFlag implements Cloneable, SharedObjectI
     this.nonProxyHosts = XMLHandler.getTagValue( slaveNode, "non_proxy_hosts" );
     this.propertiesMasterName = XMLHandler.getTagValue( slaveNode, "get_properties_from_master" );
     this.overrideExistingProperties =
-        "Y".equalsIgnoreCase( XMLHandler.getTagValue( slaveNode, "override_existing_properties" ) );
+      "Y".equalsIgnoreCase( XMLHandler.getTagValue( slaveNode, "override_existing_properties" ) );
     this.master = "Y".equalsIgnoreCase( XMLHandler.getTagValue( slaveNode, "master" ) );
     initializeVariablesFrom( null );
     this.log = new LogChannel( this );
@@ -241,7 +242,7 @@ public class SlaveServer extends ChangedFlag implements Cloneable, SharedObjectI
     return log;
   }
 
-  public String getXML() {
+  public synchronized String getXML() {
     StringBuilder xml = new StringBuilder();
 
     xml.append( "      " ).append( XMLHandler.openTag( XML_TAG ) ).append( Const.CR );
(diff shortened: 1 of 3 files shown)


IvanNikolaychuk:
@mchen-len-son , could you please review this PR?

IvanNikolaychuk:
@mchen-len-son , looks like there are some problems with WN.
Works fine localy.

mchen-len-son:
@mbatchelor can you please take a look at the PR?

mbatchelor:
Not a fan actually. Instead of reducing synchronization, it was substantially increased which I believe will lead to a performance regression. I also will be making some inline comments - but I think this needs to be reworked using re-entrant read/write locks instead of sprinkling the synchronized keyword everywhere.

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -241,7 +242,7 @@ public LogChannelInterface getLogChannel() {
     return log;
   }
 
-  public String getXML() {
+  public synchronized String getXML() {
```
mbatchelor: This makes absolutely no sense. This method creates a new StringBuilder, appends constants to the string, and returns the toString() of it. No actual instance variables referenced. This should absolutely not be synchronized!!!

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -266,13 +267,13 @@ public String getXML() {
     return xml.toString();
   }
 
-  public Object clone() {
+  public synchronized Object clone() {
```
mbatchelor: This makes no sense either. The substantive "work" done here reading instance variables is done in replaceMeta. Yet, you synchronize here, and synchronize replaceMeta. This is not rational.

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -266,13 +267,13 @@ public String getXML() {
...
     SlaveServer slaveServer = new SlaveServer();
     slaveServer.replaceMeta( this );
     return slaveServer;
   }
 
-  public void replaceMeta( SlaveServer slaveServer ) {
+  public synchronized void replaceMeta( SlaveServer slaveServer ) {
```
mbatchelor: a re-entrant write lock would make sense here.

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -443,14 +445,16 @@ public String getPortSpecification() {
     return portSpec;
   }
 
-  public String constructUrl( String serviceAndArguments ) throws UnsupportedEncodingException {
+  public synchronized String constructUrl( String serviceAndArguments ) throws UnsupportedEncodingException {
```
mbatchelor: A readlock would make more sense than synchronizing the whole method.

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -604,25 +608,35 @@ public String sendExport( String filename, String type, String load ) throws Exc
...
-    String host = environmentSubstitute( this.hostname );
-    String phost = environmentSubstitute( this.proxyHostname );
-    String pport = environmentSubstitute( this.proxyPort );
-    String nonprox = environmentSubstitute( this.nonProxyHosts );
+    String hostName;
+    String proxyHost;
+    String proxyPort;
+    String nonProxyHosts;
+
+    synchronized ( this ) {
```
mbatchelor: Again - a write lock would be much better here.

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -604,25 +608,35 @@ public String sendExport( String filename, String type, String load ) throws Exc
...
       }
-      client.getHostConfiguration().setProxy( phost, Integer.parseInt( pport ) );
+      client.getHostConfiguration().setProxy( proxyHost, Integer.parseInt( proxyPort ) );
     }
   }
 
   public void addCredentials( HttpClient client ) {
-    client.getState().setCredentials(
+    HttpState state = client.getState();
+    synchronized ( this ) {
```
mbatchelor: A read lock makes more sense here.

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -1123,7 +1137,7 @@ public long getNextSlaveSequenceValue( String slaveSequenceName, long incrementV
     }
   }
 
-  public SlaveServer getClient() {
+  public synchronized SlaveServer getClient() {
```
mbatchelor: a Read Lock (if necessary) on the variables makes more sense than synchronizing the entire method.

engine/src/org/pentaho/di/cluster/SlaveServer.java
IvanNikolaychuk: Sorry, Marc, you are wrong here.
Lets assume we have a a slaveServer shared among 2 threads (Thread-1 and Thread-2).
Lets assume getXml() method is not synchronized.

Thread-1 calls getXml() method and in the middle of XML-method execution Thread-2 calls replaceMeta method. replaceMeta begins mutating fields of sharedSlaveServer object.
As a result getXml() method can mix up it's variables (one half of variables can be from the set before replaceMeta call here, another half - after replaceMeta call).

Synchronization is needed here, agree?

pentaho-nbaker: A ReadWrite Lock can accommodate this just fine. If a Read lock is in effect the write will wait. If a Write is in affect the read will wait. Concurrent reads are not mutexed

engine/src/org/pentaho/di/cluster/SlaveServer.java
IvanNikolaychuk: Same for here.
Lets assume clone is not synchronized here.
Lets assume that Thread-1 is currently executeing replaceMeta() and it changed half of the fields of sharedSlaveServer.

Thread-2 calls clone method and clone method passes this reference to replace method. But this is not consistent, becaue not all properties were replaced yet. So, we get half of the properties replaced, half - not yet. Thats why synchronization is needed either.

Agree?

mbatchelor:
One last comment - if this class is really this volatile, why would you have synchronization around some things that look at variables and not all things looking at variables. For example, you synchronize on constructUrl (ostensibly because it reads hostname and other instance variables). However, you aren't synchronizing on getHostname() . . .  

So once again - the purpose of me writing this JIRA case back then was that I did a review of synchronization in PDI, and found places where it was inconsistent. This class was inconsistent before you started in it, and has become even more inconsistent and incoherent.

Closing this PR as it doesn't hit the mark.

engine/src/org/pentaho/di/cluster/SlaveServer.java
IvanNikolaychuk: Yeah, I agree that Reentrant lock would be a better solution.
I'm just saing that this this should be synchronized.

mbatchelor: How would it make sense to have the rug pulled out from under Thread-1? Perhaps a ThreadLocal makes more sense...

mbatchelor: Once again - this returns "      <slaveserver><cr> " - Nothing at all to synchronize here ... It's a constant.

IvanNikolaychuk: I think read lock would be a good solution, as Nick said abouve.
But this deffinately needs synchronization

IvanNikolaychuk: It references all the fields of slaveServer object

mbatchelor: Then at a read lock - not synchronizing the whole method.

IvanNikolaychuk: Agree.

engine/src/org/pentaho/di/cluster/SlaveServer.java
IvanNikolaychuk: What about this?

IvanNikolaychuk:
About your last comment:

I synchronize on constructUrl , because it takes a bunch of parameters for it's work.
This parameters need to be taken from one sharedServer and should not be mutaded in the middle of operation from another thread.

Again, lets assume it's not synchronized. We call constructUrl method from Thread-1, and execute this line:
`String realHostname = environmentSubstitute( hostname );`

Thread-2 calles replace method on sharedSlaveServer. It mutates all fields.

Thread-1 continues it's work.
We reach line:

`if ( !StringUtils.isBlank( webAppName ) ) {
`
but webAppName is mutated already and is no longer refers to that hostname, that we added before Thread-2 called constructUrl.

PR #2694, opened 2016-07-18
[PDI-8931] SlaveServer Synchronized methods need to be examined

IvanNikolaychuk:
- Test written
- Checkstyle applied

### DIFF
--- a/engine/src/org/pentaho/di/cluster/SlaveServer.java
+++ b/engine/src/org/pentaho/di/cluster/SlaveServer.java
@@ -38,9 +38,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Random;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import org.apache.commons.httpclient.Header;
 import org.apache.commons.httpclient.HttpClient;
+import org.apache.commons.httpclient.HttpState;
 import org.apache.commons.httpclient.UsernamePasswordCredentials;
 import org.apache.commons.httpclient.auth.AuthScope;
 import org.apache.commons.httpclient.methods.ByteArrayRequestEntity;
@@ -99,7 +102,7 @@ import org.w3c.dom.Document;
 import org.w3c.dom.Node;
 
 public class SlaveServer extends ChangedFlag implements Cloneable, SharedObjectInterface, VariableSpace,
-    RepositoryElementInterface, XMLInterface {
+  RepositoryElementInterface, XMLInterface {
   private static Class<?> PKG = SlaveServer.class; // for i18n purposes, needed by Translator2!!
 
   public static final String STRING_SLAVESERVER = "Slave Server";
@@ -177,11 +180,14 @@ public class SlaveServer extends ChangedFlag implements Cloneable, SharedObjectI
 
   private SslConfiguration sslConfig;
 
+  private ReadWriteLock lock;
+
   public SlaveServer() {
     initializeVariablesFrom( null );
     id = null;
     this.log = new LogChannel( STRING_SLAVESERVER );
     this.changedDate = new Date();
+    lock = new ReentrantReadWriteLock();
   }
 
   public SlaveServer( String name, String hostname, String port, String username, String password ) {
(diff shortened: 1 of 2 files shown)


IvanNikolaychuk:
@mchen-len-son, could you please review this PR?

ui/src/org/pentaho/di/ui/cluster/dialog/SlaveServerDialog.java
```diff
@@ -525,35 +525,38 @@ public void ok() {
...
       }
     }
 
-    originalServer.setName( slaveServer.getName() );
-    originalServer.setHostname( slaveServer.getHostname() );
-    originalServer.setPort( slaveServer.getPort() );
-    originalServer.setWebAppName( slaveServer.getWebAppName() );
-    originalServer.setUsername( slaveServer.getUsername() );
-    originalServer.setPassword( slaveServer.getPassword() );
+    synchronized ( originalServer ) {
```
mchen-len-son: this syncronized looks useless.
it uses object's monitor to lock whereas the object's methods use readwrite lock

mchen-len-son: can we replace those setters calls with a single method?

engine/test-src/org/pentaho/di/concurrency/SlaveServerConcurrentTest.java
```diff
@@ -0,0 +1,184 @@
...
+ * with the same id).
+ *
+ * And that's exactly what {@link Getter} does:
+ *
+ * It takes a snapshot of {@code sharedSlaveServer} and verifies it's consistency.
+ * We have to take a snapshot here because {@code sharedSlaveServer} can be updated in the middle of consistency
+ * verification. It's absolutely fine, cause it will be updated with consistent values.
+ *
+ */
+public class SlaveServerConcurrentTest {
```
mchen-len-son: Have you checked your new r/w lock based class test on Oracla Java? Doesn't it suffer from the same memory leak as those ignored ones?

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -235,6 +242,8 @@ public SlaveServer( Node slaveNode ) {
       setSslMode( true );
       this.sslConfig = new SslConfiguration( sslConfig );
     }
+
+    lock = new ReentrantReadWriteLock();
```
mbatchelor: As the first line of the constructor calls the zero-args constructor, you shouldn't have this line.

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -209,6 +215,7 @@ public SlaveServer( String name, String hostname, String port, String username,
     this.master = master;
     initializeVariablesFrom( null );
     this.log = new LogChannel( this );
+    this.lock = new ReentrantReadWriteLock();
```
mbatchelor: As the first line of this constructor calls the zero-args constructor (which instances the lock), you shouldn't have this line.

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -139,49 +142,52 @@ public static int getBackoffIncrements() {
 
   private LogChannelInterface log;
 
-  private String name;
+  private volatile String name;
```
mbatchelor: Note - I believe if you're using a ReentrantReadWriteLock, and you're always locking before reading and writing, you *shouldn't* use the volatile keyword. The volatile keyword should only be used when accessing a variable _outside_ the lock.

engine/src/org/pentaho/di/cluster/SlaveServer.java
```diff
@@ -1124,39 +1302,62 @@ public long getNextSlaveSequenceValue( String slaveSequenceName, long incrementV
...
-    String name = MessageFormat.format( "Dynamic slave [{0}:{1}]", pHostName, pPort );
-    SlaveServer client = new SlaveServer( name, pHostName, "" + pPort, getUsername(), getPassword() );
-    client.setSslMode( isSslMode() );
-    return client;
+    lock.readLock().lock();
+    try {
+      String pHostName = getHostname();
+      String pPort = getPort();
+      String name = MessageFormat.format( "Dynamic slave [{0}:{1}]", pHostName, pPort );
+      SlaveServer client = new SlaveServer( name, pHostName, "" + pPort, getUsername(), getPassword() );
```
mbatchelor: I believe the "" + pPort was a vestige of the old days when getPort() returned an integer (before it handled variables). I believe they just forgot to fix that notation. Since the class is being touched, it makes sense to fix it here.

ui/src/org/pentaho/di/ui/cluster/dialog/SlaveServerDialog.java
mbatchelor: So what is being protected here by synchronizing at all ? This is code that only runs in a thick-client (Spoon) as it's the dialog. This code doesn't run on the server. So we're protecting someone that ostensibly starts a transformation or job using one or more SlaveServers, and then has edits the SlaveServer definition while the transformation/job is running. Seems to me that's irrational and they get what they get. That's a "don't do that".

With that behavior, I think the right answer is that the dialog should work solely on a clone of the SlaveServer. Only at the moment when the "Save" is performed should the underlying SlaveServer be mutated. But there's no knowing what will happen to the executing job/transformation.

PR #2695, opened 2016-07-19
Collator Support for Sort Rows step

matthewtckr:
Set the Collator in the  ValueMetaBase instead of in the string compare method, so it is not regenerated for each `processRow()`

Set collator enabled to false per default.

Added the option to set collator strength. If locale sorting is enabled it overrides case sensitive.
Allowing users to set collator strength is better than assuming they want to use Secondary or tertiary
More intuitive when setting collator strength

Changed SortRowsMetaInjection Entries to reflect what really is happening.
Strength is now an Int throughout the entire project. setCollatorStrength now throws an IllegalArgumentException if invalid values are given.


matthewtckr:
Checkstyle violations handled in PR #2696

PR #2701, opened 2016-07-20
[BACKLOG-8792] change preview size option text

tgf:


tgf:

PR #2703, opened 2016-07-21
[BACKLOG-9100] Pluginize the Metadata Injection step

bmorrise:


bmorrise:
This pull request removes the metadata inject step from kettle and creates a plugin. There is a pull request which adds the new osgi-based plugin to the assembly here: https://github.com/pentaho/pentaho-karaf-assembly/pull/204

bmorrise:
@mdamour1976

lucboudreau:
@bmorrise Are you sure this PR is good to go? Seems like an enormous drop of coverage across the board. Is this intended?

bmorrise:
A bunch of tests were moved into the new plugin, which are being run and really, very little new code was added, so I'm wondering if the metrics are being calculated properly.

lucboudreau:
I think you're right. Wingman built core and engine, but not plugin, so the coverage has not been calculated correctly.

matthewtckr:
@bmorrise Was the dialog class (MetaInjectDialog) also supposed to be moved to the plugin?  It appears to still be in the UI module

bmorrise:
@matthewtckr It is in the plugin, but yeah, it looks like it wasn't removed from ui when this work was done. I'll pull it out of there.

lucboudreau:
@bmorrise @mdamour1976 Seems that this change broke the master branch.

https://github.com/pentaho/pentaho-kettle/pull/2717#issuecomment-235363987
https://github.com/pentaho/pentaho-kettle/pull/2710#issuecomment-235364363

PR #2704, opened 2016-07-21
[PDI-7502] Reporting Output Error on Failure

matthewtckr:
The Step should throw a PDI error if the report fails to render properly


test/org/pentaho/di/trans/steps/pentahoreporting/PDI7502Test.java
```diff
@@ -0,0 +1,176 @@
+/*! ******************************************************************************
```
mbatchelor: This is an integration test IMO - please move this over and rename with IT in the name. Thanks!

PR #2699, opened 2016-07-20
[PDI-5458] Convert Salesforce from Axis to WSC

matthewtckr:

PR #2705, opened 2016-07-21
[PDI-5458] Convert Salesforce from Axis to WSC

matthewtckr:


matthewtckr:
Addresses additional feedback in #2699 related to when the login happens in the SalesforceConnection class

plugins/salesforce/src/org/pentaho/di/trans/steps/salesforce/SalesforceConnection.java
```diff
@@ -273,6 +273,9 @@ public void setPassword( String value ) {
   public void connect() throws KettleException {
     ConnectorConfig config = new ConnectorConfig();
     config.setAuthEndpoint( getURL() );
+    config.setServiceEndpoint( getURL() );
+    config.setUsername( getUsername() );
+    config.setPassword( Encr.decryptPasswordOptionallyEncrypted( getPassword() ) );
```
lucboudreau: Seems that null passwords cause problems here.

matthewtckr: It appears to have been happening as I wasn't returning the ConnectorConfig object, due to the series of mocks.  It's been updated to pass back a mock ConnectorConfig, as we just care about the original that was passed in, that it had the correct credentials.

mbatchelor: Quick note about this - it is best practices to leave the password encrypted until the last possible moment where it needs to be decrypted. Your code here unfortunately decrypts the password into the data structure which would allow anyone to inspect/dump memory and see the password. The better answer is what used to be done before where the password was decrypted at the exact moment needed.

PR #2710, opened 2016-07-25
JIRA PDI-15531 Rest ssl ignore self signed

mikebski:
15531 - adds option (via a checkbox) to ignore invalid SSL certs.


mbatchelor:
Re-running Wingman now

lucboudreau:
Before we merge this PR, please make sure to squash the commits together, as well as making sure that the commit message starts with the Jira case number between brackets.

```
[PDI-15531] foo bar.
```

engine/src/org/pentaho/di/trans/steps/rest/Rest.java
```diff
@@ -335,10 +317,6 @@ public boolean verify( String hostname, SSLSession session ) {
     }
   }
 
-
-  protected MultivaluedMap<String, String> searchForHeaders( ClientResponse response ) {
```
matthewtckr: The [RestIT class](https://github.com/pentaho/pentaho-kettle/blob/master/engine/test-src/org/pentaho/di/trans/steps/rest/RestIT.java#L110) failed to compile because a method is said to Override this now-removed method in the Rest class.

Is there a need to remove this and other areas of this class to handle self-signed SSL certificates?

mikebski:
No, I think this is IntelliJ trying to "help".  I tried to commit the thing without these changes, but obviously it did not work.  I think it modified some imports (by removing the * and explicitly importing each package) which is not needed...

mbatchelor:
Mike,

Can you re-base and squash your commits please? I cannot process until Wingman is successful and we have things squashed.

PR #2713, opened 2016-07-25
[PDI-8928]  KettleDatabaseRepositoryConnectionDelegate - multiple un-…

AliaksandrShuhayeu:
[PDI-8928]  KettleDatabaseRepositoryConnectionDelegate - multiple unsynchronized methods that mutate objects.
extracted code for modifying sqlMap in separated synchronized method

### DIFF
--- a/engine/src/org/pentaho/di/repository/kdr/delegates/KettleDatabaseRepositoryConnectionDelegate.java
+++ b/engine/src/org/pentaho/di/repository/kdr/delegates/KettleDatabaseRepositoryConnectionDelegate.java
@@ -32,6 +32,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.pentaho.di.core.Const;
 import org.pentaho.di.core.Counter;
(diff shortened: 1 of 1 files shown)


AliaksandrShuhayeu:
@mbatchelor Could you please review it?

AliaksandrShuhayeu:
@lucboudreau Could you please review it?
Fails of wingman  don't applies to my changes.

engine/src/org/pentaho/di/repository/kdr/delegates/KettleDatabaseRepositoryConnectionDelegate.java
```diff
@@ -1917,6 +1897,15 @@ public synchronized void insertTableRow( String tablename, RowMetaAndData values
     }
   }
 
+  private synchronized PreparedStatement getPreparedStatement( String sql ) throws KettleDatabaseException {
```
mbatchelor: Even though there’s no time now to completely rewrite this tangled mass of code, we can’t just add another issue to an already bad problem. It has to start somewhere, and this is where. Please re-work your fix as follows:
•	Change sqlMap from a HashMap to a ConcurrentHashMap
•	Remove the synchronized keyword from your new method.
•	You should use the putIfAbsent method to add to the map

AliaksandrShuhayeu:
Code issue is fixed.
@mbatchelor Could you please review it?

PR #2716, opened 2016-07-25
[BACKLOG-8729] Connected repository display name not updated in PDI a…

ddiroma:
…fter editing in thin repo UI--does not address any concerns by Ben in Comments, only addresses display name update per description

@bmorrise @mdamour1976


lucboudreau:
I'm afraid that Wingman has misled you here. There was no build command, so nothing was built. I'll look into it, but in the meanwhile, you will have to validate this one manually.

lucboudreau:
If you care to know why it failed, notice the detector guessed this was a Ant project.

```
BuildToolDetector guessing ant project because build.xml exists in root dir
```

But this particular module is a Maven build. So the AntCommandBuilder doesn't know how to build it and it bails out. I'll fix this some way.

PR #2724, opened 2016-07-26
[SP-2784] Backport of PDI-15386 - PDI 5.4/6.1.0 - Excel Writer step does not retain the cell style when data format is also applied (6.1 Suite)

IvanNikolaychuk:
- clone existing cell style when applying data format
   - created a unit test for applying Format and Style from cell (from a template) when writing fields in Excel Writer step

Backport of https://github.com/pentaho/pentaho-kettle/pull/2680


IvanNikolaychuk:
@pamval , @mchen-len-son , could you please take a look at this PR?

PR #2731, opened 2016-07-26
[PDI-4048] Text file input cannot properly handle a fixed-length-of-bytes file

HiromuHota:
Hi,

This patch adds a 'length' option so that a user can specify in which length ('Characters' or 'Bytes') a fixed-length file is formatted.

Currently, a file containing multi-byte characters can be properly parsed only when the length is character-based.
This issue occurs when a file contains multi-byte characters and each field has a fixed-length of bytes.
As stated by @Ex-S in the [Forum](http://forums.pentaho.com/archive/index.php/t-54882.html), this is not an unusual case since columns of a database table may be defined by bytes instead of characters and so are the exported fixed-length files.
Related issues are [PDI-6559](http://jira.pentaho.com/browse/PDI-6559) and [PDI-15469](http://jira.pentaho.com/browse/PDI-15469).

Regards,
Hiromu Hota


mbatchelor:
Hello Hiromu-san,

Thank-you so much for your contribution. There are just a few issues with your submission though, and I'd like to work with you to get them corrected:

1- You changed the items being serialized (by adding Length), but you didn't update TextFileInputMetaLoadSaveTest to test for your new variable.

2- I ran wingman on your PR, and there were two violations - one on checkstyle (indenting) and the other on the copyright date in the class header (see above).

Can you please make the fixes and update this Pull Request so I can process it?

Thanks.

Marc

ui/src/org/pentaho/di/ui/trans/steps/fileinput/text/TextFileInputDialog.java
```diff
@@ -1491,20 +1495,41 @@ public void focusGained( org.eclipse.swt.events.FocusEvent e ) {
...
+    props.setLook( wlLength );
+    fdlLength = new FormData();
+    fdlLength.left = new FormAttachment( 0, 0 );
+    fdlLength.top = new FormAttachment( wEncoding, margin );
+    fdlLength.right = new FormAttachment( middle, -margin );
+    wlLength.setLayoutData( fdlLength );
+    wLength = new CCombo( wContentComp, SWT.BORDER | SWT.READ_ONLY );
+    wLength.setText( BaseMessages.getString( PKG, "TextFileInputDialog.Length.Label" ) );
+    props.setLook( wLength );
+    wLength.add( "Characters" );
```
matthewtckr: @HiromuHota

We missed this in the first review, but there is also a need to have the "Characters" and "Bytes" strings defined by message properties, so that it can be localized into other languages.  As an example, the [CheckSumDialog class](https://github.com/pentaho/pentaho-kettle/blob/6.1.0.1-R/ui/src/org/pentaho/di/ui/trans/steps/checksum/CheckSumDialog.java#L155-L160) does this for selecting the CheckSum Type.

HiromuHota:
Hello Marc and Matt,

Thank you for reviewing my request and gave me feedbacks.
I fixed license (year) and style (indenting), and added message properties for later localization.
Also I updated TextFileInputMetaLoadSaveTest to test the new variable "length," but I noticed that other variables might also need testing load/save.

If you are going to update the load/save test for other variables, I'm happy to do that.
Please show me a reference implementation.

Regards,
Hiromu Hota

mbatchelor:
@HiromuHota - Go to https://github.com/pentaho/pentaho-kettle/blob/master/engine/test-src/org/pentaho/di/trans/steps/textfileinput/TextFileInputMetaLoadSaveTest.java and observe the "commonAttributes" list at the top of the setup() method. Your new field name needs to be added to array list. Provided your getter/setter method names are standard ( getLength(), setLength( String value ) ) then that's all you need to do.

If your getter/setter method names are non-standard ( like hasHeader for a boolean instead of isHeader( ) or like setFileNameForTest( String value ) ) then you will need to add an entry into the getter/setter map as shown. Special object types are also handled by the load/save tester infrastructure with the typeValidators.

The LoadSaveTester framework tests the following for you automatically:
* Load/Save round trip through different repository and file types ( save as XML, load into repo, read from repo, etc).
* Clone

HiromuHota:
Thank you very much for your kind instruction, but unfortunately I got a little bit confused.

There are two implementations of the TextFileInput: one in org.pentaho.di.trans.steps.textfileinput, and another in org.pentaho.di.trans.steps.fileinput.text.
According to [this commit](https://github.com/pentaho/pentaho-kettle/commit/463c6d4357f2d5e7b7ba696ca71e3e21bffc14e2), the former one (textfileinput) has been deprecated and replaced with the latter one (fileinput.text). Please correct me if I'm wrong.
In the above comment, you suggest me to look at the LoadSaveTest for the old one (textfileinput/TextFileInputMetaLoadSaveTest.java).
Does this mean I should add the new variable "length" to the old implementation as well?

The new TextFileInput has the corresponding [LoadSaveTest](https://github.com/pentaho/pentaho-kettle/blob/master/engine/test-src/org/pentaho/di/trans/steps/fileinput/text/TextFileInputMetaLoadSaveTest.java), but the "commonAttributes" list has only three elements (once it had all the variables until [this commit](https://github.com/pentaho/pentaho-kettle/commit/ef4526c8e2a5a5486dfd114ce09569a32c516c44#diff-d4a203d4bc3913b7b190841038af2d27) deleted them)
I'm adding the new variable "length" to the new one. So I added it to the corresponding LoadSaveTest and found that many other variables were missing from the test.

lucboudreau:
@mbatchelor Marc, mind if we keep this one opened a bit? I'd like to understand why Wingman fails to do the merge of this PR's code. The file listed above was created after this PR's parent commit, and this seems to cause issues. I'll look into it.

lucboudreau:
@mbatchelor Hi Marc, I've discovered that the Kettle project [overrides the Git behavior](https://github.com/pentaho/pentaho-kettle/blob/master/.gitattributes) and forces the client to make line endings conversions.

This means that:

- We clone the repository locally.
- The git client changes all line endings.
- We merge the pull request over the original repo.
- The merge fails because the file was modified locally.

I've tried all sorts of changes to the git command line to try and override this, but I have not found a way yet. Sorry, but this PR won't be able to be validated until we figure this out.

HiromuHota:
Thank you for all your efforts on this.
If there is anything I can do, please let me know.

mbatchelor:
Closing in favor of reworked #3340. Thank-you for the contribution, sorry we couldn't just merge this one.

HiromuHota:
Thank you.

PR #2732, opened 2016-07-26
[BACKLOG-9099] Moved MetaInjectDialog to meta-inject-plugin

bmorrise:
@mdamour1976 @matthewtckr


lucboudreau:
@bmorrise Still failing. Let me know if we need to clear the wingman ivy / maven caches.

bmorrise:
@lucboudreau This would fail when trying to build master though, correct?

bmorrise:
@lucboudreau I mean, since master is already broken, doesn't it try to build master initially before building the pull request?

lucboudreau:
I see. Yes, you are correct. Let me check if head is still building.

lucboudreau:
Still fails on HEAD. Go there and search the bottom for "head:"

http://wingman.pentaho.com:8080/job/wingman/2970/console

ui/src/org/pentaho/di/ui/spoon/trans/TransGraph.java
```diff
@@ -4223,16 +4222,6 @@ public void openMapping( StepMeta stepMeta, int index ) {
         && ( stepMeta.getStepMetaInterface().getActiveReferencedObjectDescription() == null || index < 0 ) ) {
         TransMeta subTransMeta = subTrans.getTransMeta();
         referencedMeta = subTransMeta;
-        if ( stepMeta.getStepMetaInterface() instanceof MetaInjectMeta ) {
-          // Make sure we don't accidently overwrite this transformation so we'll remove the filename and objectId
-          // Modify the name so the users sees it's a result
-          subTransMeta.setFilename( null );
```
matthewtckr: It looks like this was added so that if someone "opens referenced file" while running a transformation in Spoon, it modifies the object shown in Spoon so they can't save and overwrite the original (executing) sub-transformation.

Is there a way that this can be kept, even with the step being moved to a plugin?

bmorrise: Yes, I'll move to an extension point.

PR #2740, opened 2016-07-27
[BACKLOG-8808] Restoring job/trans hover tips

mdamour1976:


mdamour1976:
@tgf @bmorrise @e-cuellar

PR #2747, opened 2016-07-27
[PDI-15538] Unable to see logs in Console

matthewtckr:


assembly/package-res/SpoonDebug.sh
```diff
@@ -31,7 +31,7 @@ done
 echo "-"
 SPOONDIR=$(dirname "$(readlink -f "$0")")
 echo Launching Spoon: "$SPOONDIR/spoon.sh $SPOON_OPTIONS"
-if [ "$SPOON_REDIRECT" -ne "1" ] ; then
+if [ "$SPOON_REDIRECT" != "1" ] ; then
```
mbatchelor: Unfortunately, the script starts with: 
#!/bin/sh
This is a bash specific fix. I believe to accept this change, we need the script to begin with #!/bin/bash

mbatchelor:
Actually, research shows I'm wrong in my assertion above. Merging the fix.

PR #2748, opened 2016-07-27
[BACKLOG-8808] Adding step highlighting and small styling tweaks

mdamour1976:


mdamour1976:
@e-cuellar slightly modified version of your commit

PR #2751, opened 2016-07-27
[BACKLOG-8808] Rolling back change of the stroke width on step selected

e-cuellar:


e-cuellar:
@mdamour1976 Please Sir!

PR #2745, opened 2016-07-27
[PDI-15105] Update javassist version

matthewtckr:

PR #2755, opened 2016-07-28
Revert "[PDI-15105] Update javassist version"

matthewtckr:
Reverts pentaho/pentaho-kettle#2745


matthewtckr:
Reverting due to build-flow problems in downstream components which are trying to use javassist-3.6.0.GA.

PR #2761, opened 2016-07-30
[SP-2766] Backport of PDI-6748 - Mondrian Input Step: pan.sh / kitche…

mchen-len-son:
…n.sh ignore mondrian.properties (5.4 Suite)

- added classes folder to classpath in order to make Kettle read mondrian.properties


mchen-len-son:
@pamval @mbatchelor please take a look

This is an additional PR for backport. 'classes' folder is not on classpath, and the mondrian.properties therefore is not used.

PR #2762, opened 2016-08-01
README: further notes about ant and Maven.

claudioscordino:


mbatchelor:
Thank-you for the contribution.

PR #2763, opened 2016-08-01
Sniffrows

claudioscordino:
This is the implementation of a new servlet on Carte for continuous
preview of row data.

Features:
- Implemented as HTML/XML servlet to avoid allocating a further socket
  (otherwise it might be problematic in some production environments)
- The Carte server keeps a fixed amount of latest rows to let monitoring data.
  This amount can be set through a GET parameter.
- The fixed size of the queue allows to:
   * avoid memory exhaustion on server-side
   * have multiple reading clients for the same step
- A HTML GET argument allows to stop listening rows freeing memory used by the buffer
- It returns an auto-refhesh HTML page with rows or a XML file containing also the row metadata.
  The refresh period is currently hard-coded and equal to 1 second.


claudioscordino:
Please ignore this pull request.

PR #2777, opened 2016-08-04
[PDI-15523] - Select Values Step Throws Errors when multiple fields with same field name occur

mtseu:
there are two commits:
 - fix a couple minor dependency inconsistencies
 - real fix and unit tests


mtseu:
@bmorrise @mdamour1976 @mbatchelor @mattcasters @matthewtckr could some of you review/merge?
not sure why coverage for some classes has decreased - I did not touch them.
everything I have implement was covered with unit tests.

PR #2789, opened 2016-08-08
[BACKLOG-9099] Moved MetaInjectDialog to meta-inject-plugin

bmorrise:
@mdamour1976


bmorrise:
This pull request is for the plugin and this is to add it to karaf: https://github.com/pentaho/pentaho-karaf-assembly/pull/210

PR #2790, opened 2016-08-09
[TEST] removed unit tests backported by a mistake

mchen-len-son:
- some of them are failing, some not. but even those that are not failing should be removed, cause corresponding changes they were written for are not contained by 5.4 branch


mchen-len-son:
@pamval @mbatchelor please review

PR #2791, opened 2016-08-09
[PDI-15507] Connections not being released after SQL Exception

AliaksandrShuhayeu:
-added tests for throw exception in methods commit() and pstm.close()
 -added fix in disconnect() for close connection  in these cases

@brosander @mchen-len-son Could you please review?


AliaksandrShuhayeu:
Copyright year is fixed.

PR #2794, opened 2016-08-10
PDI-15519: Value Mapper step has strange behaviour when the source va…

stanislau-strelchanka:
PDI-15519: Value Mapper step has strange behaviour when the source value of the first line is blank - check for null


stanislau-strelchanka:

PR #2798, opened 2016-08-11
[SP-2819] - Backport of PDI-15523 - Select Values Step Throws Errors when multiple fields with same field name occur (6.1 Suite)

mtseu:
[PDI-15523] - Select Values Step Throws Errors when multiple fields with same field name occur

Earlier it used to work since RowMeta clone was implemented as sequential addValueMeta calls.
After reworking the RowMeta implementation this logic was eliminated, what is right:
cloning should replicate the state as is (w/o) taking care of fixing incosistencies.

Real inconsistancy (duplicated fields' names) were caused by the fact that setValueMeta method
did not process the case when duplicated field's name is introduced.
The fix is to incorporate deduplication to the setValueMeta() call logic.

license header

### DIFF
--- a/core/src/org/pentaho/di/core/row/RowMeta.java
+++ b/core/src/org/pentaho/di/core/row/RowMeta.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2013 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2016 by Pentaho : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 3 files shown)


mtseu:
@bmorrise @mdamour1976 @mbatchelor @mattcasters @matthewtckr may I ask you to review?
the issue with check-style errors has already been communicated to sir @wingman-pentaho.
The same code for master doesn't cause check-style errors (https://github.com/pentaho/pentaho-kettle/pull/2777).

pamval:
@mtseu can you please cherry-pick this to 6.1.0.5?

PR #2803, opened 2016-08-12
[PDI-15478] Error quickly on invalid connection

matthewtckr:


matthewtckr:
SalesforceConnectionIT test failure addressed in #2805

PR #2811, opened 2016-08-15
[PDI-14278] Calculator step provides incorrect values when converting…

mchen-len-son:
… time zones

- hourOfDay calculation now considers valueMeta`s time zone
- added unit test

### DIFF
--- a/core/src/org/pentaho/di/core/Const.java
+++ b/core/src/org/pentaho/di/core/Const.java
@@ -1023,6 +1023,14 @@ public class Const {
   public static final String KETTLE_SPLIT_FIELDS_REMOVE_ENCLOSURE = "KETTLE_SPLIT_FIELDS_REMOVE_ENCLOSURE";
 
   /**
+   * Compatibility settings for {@link org.pentaho.di.core.row.ValueDataUtil#hourOfDay(ValueMetaInterface, Object)}.
+   *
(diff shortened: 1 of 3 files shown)


core/src/org/pentaho/di/core/Const.java
```diff
@@ -1023,6 +1023,12 @@ public String getMessage() {
   public static final String KETTLE_SPLIT_FIELDS_REMOVE_ENCLOSURE = "KETTLE_SPLIT_FIELDS_REMOVE_ENCLOSURE";
 
   /**
+   * The name of the variable that switches off fix for calculation of timezone decomposition.
+   */
+  public static final String KETTLE_OLD_DATE_CALCULATION_TIMEZONE_DECOMPOSITION =
```
matthewtckr: Can we update the variable name and value to use the existing naming standard for backwards-compatibility switches?

`KETTLE_COMPATIBILITY_*`

https://github.com/mchen-len-son/pentaho-kettle/blob/a06ded8a52b1bde7c56e04d4cc7bf91eca157f09/core/src/org/pentaho/di/core/Const.java#L731

mchen-len-son:
@matthewtckr, thanks for pointing out. I've renamed the property

@mbatchelor @matthewtckr @bmorrise please review

PR #2820, opened 2016-08-16
[BACKLOG-9351] Ensure default values for publishing to a server fall …

pdesai16:
…in line with Pentaho Server's hostname and port

- Changed the default-url in pdi from http://localhost:9080/pentaho-di to http://localhost:8080/pentaho
- Made the same chanegs in sample transformation


pdesai16:
@pentaho/rogueone  please review

PR #2831, opened 2016-08-17
[PDI-11194] - Formula now treating empty strings as null

YuryBY:
@CodeOnCoffee, @pamval could you please take a look?

Current behavior is following: if KETTLE_EMPTY_STRING_DIFFERS_FROM_NULL=Y then null is retrieved as empty string like it was in old product versions ( 4.4.0 ).

when KETTLE_EMPTY_STRING_DIFFERS_FROM_NULL=N we have null as returned value.


YuryBY:
@CodeOnCoffee, @pamval, @mchen-len-son  any updates?

PR #2845, opened 2016-08-19
PDI-15133 KTR with Filter Rows step does not save TRUE/FALSE settings…

stanislau-strelchanka:
PDI-15133 KTR with Filter Rows step does not save TRUE/FALSE settings when exported to XML from repository


stanislau-strelchanka:

PR #2863, opened 2016-08-23
[SP-2811][PDI-8931] SlaveServer Synchronized methods need to be examined

AliaksandrShuhayeu:
- Test written
- Checkstyle applied
@lucboudreau @mchen-len-son Could you please merge it?


mchen-len-son:
Checked locally, the tests are passing.
Checkstyle is OK.
Merging.

PR #2872, opened 2016-08-24
Revert "[SP-2811][PDI-8931] SlaveServer Synchronized methods need to be examined"

mchen-len-son:
Reverts pentaho/pentaho-kettle#2863


mchen-len-son:
#2863 was made from cherry-picking the wrong commit.
Reverting.

PR #2873, opened 2016-08-24
[CLEANUP] Mavenize Kettle Shapefile Reader Plugin

matthewtckr:


mbatchelor:
Did you work with Engops on this one? I don't want to process it and break the build tonight.

matthewtckr:
I spoke wth Steve about this last week, we'll need to coordinate as it requires a change to the PDI jobs.

PR #2891, opened 2016-08-29
[PDI-15067] created fallback for resolving repository paths in 'file' field in the Mapping step

andrey-markelov:
@mdamour1976 please review


mbatchelor:
@andrey-markelov - please fix test case and copyright date.

engine/src/org/pentaho/di/trans/steps/mapping/MappingMeta.java
```diff
@@ -712,75 +707,139 @@ public static final synchronized TransMeta loadMappingMeta( MappingMeta mappingM
     return mappingTransMeta;
   }
 
-  public void check( List<CheckResultInterface> remarks, TransMeta transMeta, StepMeta stepMeta,
-    RowMetaInterface prev, String[] input, String[] output, RowMetaInterface info, VariableSpace space,
-    Repository repository, IMetaStore metaStore ) {
+  /*
+   * loads transformation from local file system or from repository if failed and connected to repository
+   */
+  public static TransMeta loadTransFromAnyPath( String realFilename, Repository repo, IMetaStore metaStore )
```
mbatchelor: I'm not sure I agree with this at all. This seems to me to be flawed reasoning and potentially a security vulnerability. It should not grab something off the file system if you attempt to load it from the repository and it fails. That's just bad. The issue to resolve is that ${Internal.Entry.Current.Directory} isn't being resolved when connected to a repository and the transformation/job are stored in the repository. Please fix that.

andrey-markelov:
@mbatchelor
Hi Marc
It's related not only to ${Internal.Entry.Current.Directory} , as I understood, it's issue when user is migrating a whole directory with transformations (that contains links to sub-trans via Mapping step) to a repository

adding @mdamour1976 to discussion

andrey-markelov:
@mbatchelor this variable is succesfully resolving. If we use it in repository path field, it works well. But customer wants to use repository path in the first 'file' field, that we can see on the attachment.

mchen-len-son:
This has been moved from a PDI case to a Backlog case so a sprint team gets it.
Closing.

PR #2897, opened 2016-08-31
[BACKLOG-10562] Fix calls to extension points

nantunes:
Most of the extension points introduced in https://github.com/pentaho/pentaho-kettle/pull/911 aren't being called.

@pentaho/millenniumfalcon and @mdamour1976 please review.


mdamour1976:
This looks good, my only fear is what might depend on this being "wrong." We'll have to pay attention and deal with any fallout in other projects.

PR #2898, opened 2016-08-31
[BACKLOG-10068] When Lazy Conversion is enabled aggregations cannot b…

DFieldFL:
…e calculated since the aggregrated value is converted to a numeric even though the meta type is still set to Binary String. Converting the aggregation meta type from binary string to normal when a convert call is made.


DFieldFL:
Emailed appropriate people about the test failures. This code is good to go.

PR #2905, opened 2016-09-08
BACKLOG-10651 Ensuring that logel level is restored

flbrino:
* Correction for the cases where Log Level was changed to Basic Level and not restored.


flbrino:
@mkambol can you please validate?

PR #2906, opened 2016-09-08
[PDI-15255] kitchen.sh/pan.sh cannot use relative paths to jobs/transformations stored on the filesystem

AliaksandrShuhayeu:
[PDI-15255] kitchen.sh/pan.sh cannot use relative paths to jobs/transformations stored on the filesystem

-calculate current (run) directory in scripts,
-sent current directory to kitchen and pan,
-calculate absolute path to file if used relative.

### DIFF
--- a/assembly/package-res/Kitchen.bat
+++ b/assembly/package-res/Kitchen.bat
@@ -1,7 +1,8 @@
 @echo off
 setlocal
+SET initialDir=%cd%
 pushd %~dp0
 SET STARTTITLE="Kitchen"
 SET SPOON_CONSOLE=1
-call Spoon.bat -main org.pentaho.di.kitchen.Kitchen %*
+call Spoon.bat -main org.pentaho.di.kitchen.Kitchen -initialDir "%initialDir%"\ %*
 popd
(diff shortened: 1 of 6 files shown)


assembly/package-res/Kitchen.bat
```diff
@@ -1,7 +1,8 @@
 @echo off
+SET initialDir=%cd%
 setlocal
 pushd %~dp0
 SET STARTTITLE="Kitchen"
 SET SPOON_CONSOLE=1
-call Spoon.bat -main org.pentaho.di.kitchen.Kitchen %*
+call Spoon.bat -main org.pentaho.di.kitchen.Kitchen -initialDir %initialDir%\ %*
```
mbatchelor: This will fail if the folder has spaces in it. I know from past experience. It needs to be quoted

assembly/package-res/Kitchen.bat
```diff
@@ -1,7 +1,8 @@
 @echo off
+SET initialDir=%cd%
```
mbatchelor: Why are you doing this before the setlocal - I believe this needs to be done after the setlocal so that the variable doesn't stay in the user's environment after the script runs.

AliaksandrShuhayeu:
@mbatchelor. Code issues are fixed. New way for fix it is added to comment to PDI.

PR #2908, opened 2016-09-09
PDI-15653

jonjarvis:
Added helper methods so plugins compiled with previous API (not using CharSequence call) will continue to run, and not error with link error:

An unexpected error occurred in Spoon: 
org.pentaho.di.core.Const.isEmpty(Ljava/lang/String;)Z
java.lang.NoSuchMethodError: org.pentaho.di.core.Const.isEmpty(Ljava/lang/String;)Z

Plugins compiled in versions prior to 6.1.0.6 and 7 expect there to be an isEmpty(L/java/lang/String) signature, not isEmpty(L/java/lang/CharSequence). Adding the helper methods will make sure plugins compiled and on the marketplace will not need to be compiled two ways.


mbatchelor:
I reverted this merge because I noticed that it was a Merge commit. Please squash your commits, submit a new PR, and I'll process that.

PR #2912, opened 2016-09-09
PDI-15653

jonjarvis:
Marc - New pull request after squash


mbatchelor:
Thank-you again Johathan, but I redid my work to be able to properly include deprecation notices. Unfortunately to do that, the methods moved to the Utils class, and I had to touch nearly everything. This was done on PR #2913 .

PR #2913, opened 2016-09-12
[PDI-15007] - Deprecate isEmpty method and move to Utils

mbatchelor:


mbatchelor:
Note - the newly broken tests don't fail on my machine.

PR #2924, opened 2016-09-16
[BACKLOG-9901] - Added not supported label to file and database repos…

wseyler:
…itory types.  Removed the Kettle reference from the success screens.


wseyler:
@bmorrise @mdamour1976

PR #2936, opened 2016-09-20
[SP-2856] - Backport of PDI-14619 - Data Grid step - Impossible to wr…

YuryBY:
…ite the "Internet address" type value to the text file (6.1 Suite)

@mchen-len-son, @pamval, here is partly backport of https://github.com/pentaho/pentaho-kettle/pull/2766 to 6.1. Thanks.


mbatchelor:
Yuri - you only got one of the two commits on that SP. The SP says PDI-14619, but we need both commits as PDI-12807 links to the same SP cases so it's expected that both commits will come across. Let me know if you need different SP cases for the other commit (Excel Output).

PR #2937, opened 2016-09-20
KTR with Filter Rows step does not save TRUE/FALSE settings when expo…

stanislau-strelchanka:
KTR with Filter Rows step does not save TRUE/FALSE settings when exported to XML from repository - fix regression


stanislau-strelchanka:
regression fix of #2845

stanislau-strelchanka:

PR #2939, opened 2016-09-20
[SP-2857] - Backport of PDI-14619 - Data Grid step - Impossible to wr…

YuryBY:
…ite the "Internet address" type value to the text file (5.4 Suite)

@mchen-len-son, @pamval, here is partly backport of https://github.com/pentaho/pentaho-kettle/pull/2766/files to 5.4. Thanks.


mbatchelor:
Yuri - as on the 6.1 PR for SP-2856 - we need BOTH commits for this to move forward. PDI-14619 and PDI-12807 which were both addressed in the same PR. This only brought over the commit for PDI-14619. PDI-12807 linked to the same SPs so it's expected that the SP will pick up *both* commits.

PR #2942, opened 2016-09-20
[BACKLOG-10799] - creating method to close expanded content rather th…

kurtwalker:
…an only hiding it.  this will prevent errors and flickering from unintended page loads


kurtwalker:
fyi waiting to merge because end of sprint

kurtwalker:
Safe to merge now.

PR #2946, opened 2016-09-21
[SP-2882] Backport of PDI-15133 - KTR with Filter Rows step does not …

stanislau-strelchanka:
[SP-2882] Backport of PDI-15133 - KTR with Filter Rows step does not save TRUE/FALSE settings when exported to XML from repository (6.1 Suite)


stanislau-strelchanka:
backport of:
#2845
#2937

PR #2953, opened 2016-09-22
[SP-2848][PDI-15507][PDI-15669] Connections not being released after SQL Exception

AliaksandrShuhayeu:
@mbatchelor @mchen-len-son Could you please merge it?


mbatchelor:
Aliaksandr - we cannot merge this without getting the changes for PDI-15669. The fix here breaks transactional transformations.

PR #2972, opened 2016-09-27
[BACKLOG-9764]: Sentry: "Columns exist in a table" fails for the user…

TatsianaKasiankova:
… with column select privileges

"Columns exist in a table" checks if there is a table in DB before starting checking existing columns.
The check of existing tables performs via "select * from table" clause. It requires READ access on the table.
If the user has no select permission on the whole table but has select permissions only on the column, than such check will fail with no access error and etc.

The fix has been done taking into account not to change existing behavior of this job entry.
So it has been added method to check existing table via data base meta data.


TatsianaKasiankova:
@dkincade , @hudak Hi,please review the PR. The changes have been covered with unit tests and tested locally.

org.pentaho.di.core.database.Database

Branch Change: + 1.03%
Complexity Change: + .58%
Instruction Change: + 1.05%
Line Change: + .95%
Method Change: + .63%

PR #2976, opened 2016-09-28
[BACKLOG-11172] As an administrator, I need the impersonation logging to

SergeyTravin:
work out of the box to identify what user is impersonating within the
cluster.


SergeyTravin:
@rmansoor, can you please review the changes?

SergeyTravin:
The changes related to the ee project so we should not change the common files

PR #2977, opened 2016-09-28
[BACKLOG-10653] repository status getting changed when not connected

tgf:
@mdamour1976 @bmorrise please review


tgf:
fixed checkstyle

PR #2980, opened 2016-09-28
Fixed Content-Length miscalculation

luciano-medallia:
The second parameter of InputStreamRequestEntity expects the Content-Length in bytes, which is not necessarily the same as the length of the Java String. For example, the following code:

`        String s = "©";
        System.out.println(s.length());
        System.out.println(s.getBytes(StandardCharsets.UTF_8).length);`

outputs:

> 1
> 2
> 

because the copyright-sign character has a 2-byte encoding in UTF-8, but requires only 1 Java char.


mbatchelor:
Fixed by PR #3006

PR #2982, opened 2016-09-29
[PDI-15549]-fixed job loading from PurRepository. Now metaStore property is not null after job loading.

ArtsiomYurhevich:


ArtsiomYurhevich:
@bmorrise, @wseyler, @e-cuellar  Could you please take a look?

PR #2986, opened 2016-09-29
[BACKLOG-11071] - Blocked keyboard shortcuts breaking DET interactions

diogofscmariano:
@pentaho/orlandocalisbon please review

I had to move away from the previous implementation because the key combination number exploded. This if-else can have multiple implementations, but i prefer to have it more verbose and at the same time easier to understand if we need to add more cases. Let me know what you think.

No unit tests are expected :)

### DIFF
--- a/ui/src/org/pentaho/di/ui/spoon/ExpandedContentManager.java
+++ b/ui/src/org/pentaho/di/ui/spoon/ExpandedContentManager.java
@@ -33,6 +33,8 @@ import org.eclipse.swt.events.KeyListener;
 import org.eclipse.swt.layout.FormAttachment;
 import org.eclipse.swt.layout.FormData;
 import org.eclipse.swt.widgets.Control;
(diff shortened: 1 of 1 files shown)


krivera-pentaho:
It's pretty ugly to have all of these nested if statements. I would prefer to store each "set" of conditions in a variable and then "or" them together and assign that value to keyEvent.doit.

For example
```
boolean isReload = ( keyEvent.stateMask == SWT.CTRL && ( keyEvent.keyCode == SWT.F5 || keyEvent.keyCode == 82 ) || keyEvent.keyCode == SWT.F5 );

boolean isPrint = ( keyEvent.stateMask == SWT.CTRL && keyEvent.keyCode == 80 );

keyEvent.doit = !(isReload || isPrint);

```

diogofscmariano:
@krivera-pentaho pr was updated :) thanks for the suggestion

PR #2990, opened 2016-09-30
[BACKLOG-11204] - Fix for: Data Services deadlocks if transformation …

rfellows:
…contains Table Output step with 'Truncate Table' selected


rfellows:
DON'T MERGE THIS YET. waiting for confirmation on the approach

PR #2991, opened 2016-10-03
[BACKLOG-11071] - Updated keys to be blocked. Removed print key block

diogofscmariano:


diogofscmariano:
@pentaho/orlandocalisbon please review
Removed print block. Going to try to block the mouse wheel events from the client side in DET

annapopova1:
Need fix code style before merge

jvelasques:
fixed by #2993

PR #2992, opened 2016-10-04
[PDI-15635] - Changed implementations of collections, which are used in multithreaded environment im order to get proper synchronization and better performance.

ArtsiomYurhevich:
resolved


ArtsiomYurhevich:
@bmorrise,@wseyler, @e-cuellar Could you please take a look?

PR #2997, opened 2016-10-05
[CLEANUP] Update jquery version for repositories

bmorrise:
@ddiroma

### DIFF
--- a/plugins/repositories-plugin/src/main/resources/META-INF/js/require.json
+++ b/plugins/repositories-plugin/src/main/resources/META-INF/js/require.json
@@ -20,7 +20,7 @@
   "map": {
     "repositories_7.0": {
       "angular": "angular_1.5.6",
-      "jquery": "common-ui/jquery",
+      "jquery": "jquery_2.2.1",
       "angular-sanitize": "angular-sanitize_1.5.6",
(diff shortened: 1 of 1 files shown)


nantunes:
@bmorrise,

The common-ui version disappeared from the CE assembly because the marketplace no longer depends on it. It is a good choice to also update jquery in this plugin, but it reveals the `pom.xml` dependencies aren't correct and other assembly changes due to unrelated projects can lead to future problems.

Also noticed the pom.xml has a reference to the unsafe angular-sanitize 1.2.15 (http://jira.pentaho.com/browse/PPP-3559). It has no effect, as https://github.com/bmorrise/pentaho-kettle/commit/22ffa257b338bab63c08d3ebb83baefbc419816a updated the `META-INF/js/require.json` referece, but it would be recommended to update it.

FYI @pamval

bmorrise:
@nantunes Originally the decision was made to piggyback on what marketplace had in terms of libraries, but as you point out, this was a poor decision. I'm going to update this project in the next couple of days so we won't have potential problems going forward. In addition to angular-sanitize having no effect, I discovered that other dependencies in this pom are also not necessary. I have a commit that will do some of this clean up as well.

PR #3001, opened 2016-10-07
[PDI-15070] Get rows from result" step does not pass rows when a transformation is executed on a remote carte server

AliaksandrShuhayeu:
[PDI-15070] Get rows from result" step does not pass rows when a transformation is executed on a remote carte server
-fix using encode XML tags for "binary-string" node

I saw that the same issue is available for STORAGE_TYPE_INDEXED (tag "index-value") but I codln't find using this storage type (except junit tests).

### DIFF
--- a/core/src/org/pentaho/di/core/row/value/ValueMetaBase.java
+++ b/core/src/org/pentaho/di/core/row/value/ValueMetaBase.java
@@ -3220,7 +3220,8 @@ public class ValueMetaBase implements ValueMetaInterface {
             // at all.
             //
             string = XMLHandler.addTagValue( "binary-string", (byte[]) object );
-            break;
+            xml.append( XMLHandler.openTag( XML_DATA_TAG ) ).append( string ).append( XMLHandler.closeTag( XML_DATA_TAG ) );
+            return xml.toString();
 
           case STORAGE_TYPE_INDEXED:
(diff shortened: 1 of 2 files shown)


core/src/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -3220,7 +3220,8 @@ public String getDataXML( Object object ) throws IOException {
             // at all.
             //
             string = XMLHandler.addTagValue( "binary-string", (byte[]) object );
-            break;
+            xml.append( XMLHandler.openTag( XML_DATA_TAG ) ).append( string ).append( XMLHandler.closeTag( XML_DATA_TAG ) );
```
mbatchelor: So isn't this what the code originally did except using xml.append( XMLHandler.addTagValue( XML_DATA_TAG, string ) ); instead of openTag/closeTag ? What is the difference between what it did before and what your change has ?

AliaksandrShuhayeu: @mbatchelor
XMLHandler.addTagValue( XML_DATA_TAG, string ) escapes XML specific tags < > 
and after that xml will be not correct parse, we will not get node <binary-string>

PR #3002, opened 2016-10-07
[PDI-15270] - CSV File Input mangles field containing substring in Enclosure

YuryBY:
@rmansoor, @kurtwalker, @mbatchelor, could you please take a look?


kurtwalker:
In addition to figuring out the failed tests.  I'd prefer that the PR had one commit for correcting checkstyle and one with both the change and the unit tests.  Makes tracking down things easier later.

YuryBY:
@kurtwalker, the PR is updated. I minimized changes so that now the commit contains only fix and junit tests. There are no checkstyle violations localy. Thanks.

PR #3004, opened 2016-10-07
[BACKLOG-11051] Hover state locking area

e-cuellar:


e-cuellar:
@mdamour1976 @bmorrise Please Sir!

mdamour1976:
It would seem to be that you should be able to make this work only using the existing MouseUp extention point.  It already passes the object that you added the "lock" flag to.  The existing DET mouseup extension can be tweaked emit the flag and detect the click area and return.

https://github.com/pentaho/pentaho-kettle/pull/3004/files#diff-3aa40ebb21643228592ac0f1e1d322ddR1019

Also, can we rename the lock flag to something like "preventDefault"

PR #3011, opened 2016-10-10
[BACKLOG-10097]Default data services connections to the Pentaho Server

hudak:
http://jira.pentaho.com/browse/BACKLOG-10097


hudak:
See also: https://github.com/pentaho/pdi-dataservice-plugin/pull/75

PR #3012, opened 2016-10-10
Edited default case of convertDataUsingConversionMetaData method to m…

roboguy:
…itigate

NPE when using a custom value meta plugin.
Added unit tests for method convertDataUsingConversionMetaData.
Address checkstyle warnings.

	modified:   core/src/org/pentaho/di/core/row/value/ValueMetaBase.java
	modified:   core/test-src/org/pentaho/di/core/row/value/ValueMetaBaseTest.java

Resubmitting pull request after updating to latest master code.


mbatchelor:
NPE cleanup case. Includes expanded test cases.

PR #3017, opened 2016-10-12
PDI-8937 - Fixed inconsistent synchronization in TransformationMap. Replaced with concurrent collections and removed  redundant synchronized blocks.

ArtsiomYurhevich:
Fixed

### DIFF
--- a/engine/src/org/pentaho/di/www/TransformationMap.java
+++ b/engine/src/org/pentaho/di/www/TransformationMap.java
@@ -22,12 +22,17 @@
 
 package org.pentaho.di.www;
 
-import java.util.ArrayList;
+
+
+
 import java.util.Date;
-import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.ArrayList;
+import java.util.Collections;
 
 import org.pentaho.di.core.exception.KettleException;
 import org.pentaho.di.core.util.Utils;
@@ -41,18 +46,15 @@ import org.pentaho.di.trans.TransConfiguration;
  *
  */
 public class TransformationMap {
-  private Map<CarteObjectEntry, Trans> transformationMap;
-  private Map<CarteObjectEntry, TransConfiguration> configurationMap;
+  private final Map<CarteObjectEntry, TransData> transMap;
 
-  private Map<String, List<SocketPortAllocation>> hostServerSocketPortsMap;
+  private final Map<String, List<SocketPortAllocation>> hostServerSocketPortsMap;
 
   private SlaveServerConfig slaveServerConfig;
 
   public TransformationMap() {
-    transformationMap = new Hashtable<CarteObjectEntry, Trans>();
-    configurationMap = new Hashtable<CarteObjectEntry, TransConfiguration>();
-
-    hostServerSocketPortsMap = new Hashtable<String, List<SocketPortAllocation>>();
+    transMap = new ConcurrentHashMap<>();
+    hostServerSocketPortsMap = new ConcurrentHashMap<>();
   }
 
   /**
@@ -67,18 +69,16 @@ public class TransformationMap {
    * @param transConfiguration
    *          the transformation configuration to add
    */
-  public synchronized void addTransformation( String transformationName, String containerObjectId, Trans trans,
+  public void addTransformation( String transformationName, String containerObjectId, Trans trans,
     TransConfiguration transConfiguration ) {
     CarteObjectEntry entry = new CarteObjectEntry( transformationName, containerObjectId );
-    transformationMap.put( entry, trans );
-    configurationMap.put( entry, transConfiguration );
+    transMap.put( entry, new TransData( trans, transConfiguration ) );
   }
 
-  public synchronized void registerTransformation( Trans trans, TransConfiguration transConfiguration ) {
+  public void registerTransformation( Trans trans, TransConfiguration transConfiguration ) {
     trans.setContainerObjectId( UUID.randomUUID().toString() );
     CarteObjectEntry entry = new CarteObjectEntry( trans.getTransMeta().getName(), trans.getContainerObjectId() );
-    transformationMap.put( entry, trans );
-    configurationMap.put( entry, transConfiguration );
+    transMap.put( entry, new TransData( trans, transConfiguration ) );
   }
 
   /**
@@ -87,10 +87,10 @@ public class TransformationMap {
    * @param transformationName
    * @return the first transformation with the specified name
    */
-  public synchronized Trans getTransformation( String transformationName ) {
-    for ( CarteObjectEntry entry : transformationMap.keySet() ) {
+  public Trans getTransformation( String transformationName ) {
+    for ( CarteObjectEntry entry : transMap.keySet() ) {
       if ( entry.getName().equals( transformationName ) ) {
-        return transformationMap.get( entry );
+        return transMap.get( entry ).getTrans();
       }
     }
     return null;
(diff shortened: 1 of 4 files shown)


ArtsiomYurhevich:
@mbatchelor, @bmorrise, @e-cuellar Could you please take a look?

engine/src/org/pentaho/di/www/TransformationMap.java
```diff
@@ -41,18 +46,15 @@
  *
  */
 public class TransformationMap {
-  private Map<CarteObjectEntry, Trans> transformationMap;
-  private Map<CarteObjectEntry, TransConfiguration> configurationMap;
+  private final Map<CarteObjectEntry, TransData> transMap;
 
-  private Map<String, List<SocketPortAllocation>> hostServerSocketPortsMap;
+  private final Map<String, List<SocketPortAllocation>> hostServerSocketPortsMap;
```
mbatchelor: Instead of a <String, List>, I would rather you used a <String, Deque<SocketPortAllocation>> with the implementation of a ConcurrentlinkedDeque. The iteration over the ConcurrentLinkedDeque will be identical, your Deque will be in the same order as added - but you will be able to do all your other operations (iteration, etc) without synchronization below.

engine/src/org/pentaho/di/www/TransformationMap.java
```diff
@@ -219,27 +204,24 @@ public synchronized SocketPortAllocation allocateServerSocketPort( int portRange
...
-      if ( serverSocketPortsMap == null ) {
-        serverSocketPortsMap = new ArrayList<SocketPortAllocation>();
-        hostServerSocketPortsMap.put( hostname, serverSocketPortsMap );
-      }
+    List<SocketPortAllocation> serverSocketPorts;
+    serverSocketPorts = hostServerSocketPortsMap.computeIfAbsent( hostname, k -> new CopyOnWriteArrayList<>() );
+    serverSocketPorts = serverSocketPorts != null ? serverSocketPorts : hostServerSocketPortsMap.get( hostname );
 
-      synchronized ( serverSocketPortsMap ) {
+    synchronized ( serverSocketPorts ) {
```
mbatchelor: No synchronization needed if using the ConcurrentLinkedDeque as I mention above

engine/src/org/pentaho/di/www/TransformationMap.java
```diff
@@ -314,15 +294,13 @@ public synchronized SocketPortAllocation allocateServerSocketPort( int portRange
...
   public void deallocateServerSocketPorts( String transName, String carteObjectId ) {
-    synchronized ( hostServerSocketPortsMap ) {
-      for ( String hostname : hostServerSocketPortsMap.keySet() ) {
-        List<SocketPortAllocation> spas = hostServerSocketPortsMap.get( hostname );
-        synchronized ( spas ) {
-          for ( SocketPortAllocation spa : spas ) {
-            if ( spa.getTransformationName().equalsIgnoreCase( transName )
+    for ( String hostname : hostServerSocketPortsMap.keySet() ) {
+      List<SocketPortAllocation> spas = hostServerSocketPortsMap.get( hostname );
+      synchronized ( spas ) {
```
mbatchelor: No synchronization needed if using the ConcurrentLinkedDeque as I mention above

engine/src/org/pentaho/di/www/TransformationMap.java
```diff
@@ -337,9 +315,12 @@ public void deallocateServerSocketPorts( String transName, String carteObjectId
    */
   public void deallocateServerSocketPorts( CarteObjectEntry entry ) {
     for ( String hostname : hostServerSocketPortsMap.keySet() ) {
-      for ( SocketPortAllocation spa : hostServerSocketPortsMap.get( hostname ) ) {
-        if ( spa.getTransformationName().equalsIgnoreCase( entry.getName() ) ) {
-          spa.setAllocated( false );
+      List<SocketPortAllocation> serverSocketPorts = hostServerSocketPortsMap.get( hostname );
+      synchronized ( serverSocketPorts ) {
```
mbatchelor: No synchronization needed if using the ConcurrentLinkedDeque as I mention above

engine/src/org/pentaho/di/www/TransformationMap.java
```diff
@@ -348,24 +329,27 @@ public void deallocateServerSocketPorts( CarteObjectEntry entry ) {
...
+    List<SocketPortAllocation> serverSocketPorts = hostServerSocketPortsMap.get( hostname );
+
+    if ( serverSocketPorts == null ) {
       return; // nothing to deallocate
     }
 
-    // Find the socket port allocation in the list...
-    //
-    for ( SocketPortAllocation spa : new ArrayList<SocketPortAllocation>( serverSocketPortsMap ) ) {
+    synchronized ( serverSocketPorts ) {
```
mbatchelor: No synchronization needed if using the ConcurrentLinkedDeque as I mention above

engine/src/org/pentaho/di/www/TransformationMap.java
```diff
@@ -391,16 +375,8 @@ public void setSlaveServerConfig( SlaveServerConfig slaveServerConfig ) {
...
-  }
-
-  /**
-   * @param hostServerSocketPortsMap
-   *          the hostServerSocketPortsMap to set
-   */
-  public void setHostServerSocketPortsMap( Map<String, List<SocketPortAllocation>> hostServerSocketPortsMap ) {
-    this.hostServerSocketPortsMap = hostServerSocketPortsMap;
+  public List<SocketPortAllocation> getHostServerSocketPorts( String hostname ) {
+    return Collections.unmodifiableList( hostServerSocketPortsMap.get( hostname ) );
```
mbatchelor: This would need to change to:
return Collections.unmodifiableList( new ArrayList( hostServerSocketPortsMap.get( hostname ) ) );

engine/src/org/pentaho/di/www/TransformationMap.java
ArtsiomYurhevich: Hello ,@mbatchelor . Here we can't change to Deque implementation as https://github.com/ArtsiomYurhevich/pentaho-kettle/blob/d12f1978ce5c1625c5e862975695c64d98b2f922/engine/src/org/pentaho/di/www/TransformationMap.java#L255 uses set by index. And i have left synchronization blocks due to consistent synchronization over operations with boolean value isAllocated. Look https://github.com/ArtsiomYurhevich/pentaho-kettle/blob/d12f1978ce5c1625c5e862975695c64d98b2f922/engine/src/org/pentaho/di/www/TransformationMap.java#L234#L240 , https://github.com/ArtsiomYurhevich/pentaho-kettle/blob/d12f1978ce5c1625c5e862975695c64d98b2f922/engine/src/org/pentaho/di/www/TransformationMap.java#L265#L269 and deallocate methods which set the value to false. Without synchronization it can be possible such cases as allocating already allocated SocketPortAllocation and etc. Also List implementation CopyOnWriteArrayList is  a thread safe one https://github.com/ArtsiomYurhevich/pentaho-kettle/blob/d12f1978ce5c1625c5e862975695c64d98b2f922/engine/src/org/pentaho/di/www/TransformationMap.java#L210 and we can use it in this case.

engine/src/org/pentaho/di/www/TransformationMap.java
ArtsiomYurhevich: @mbatchelor, Could you please explain why we need to create new ArrayList from existing one? Anyway it just copies references from the list in the constructor .

engine/src/org/pentaho/di/www/TransformationMap.java
```diff
@@ -41,18 +46,15 @@
...
+  private final Map<String, List<SocketPortAllocation>> hostServerSocketPortsMap;
 
   private SlaveServerConfig slaveServerConfig;
 
   public TransformationMap() {
-    transformationMap = new Hashtable<CarteObjectEntry, Trans>();
-    configurationMap = new Hashtable<CarteObjectEntry, TransConfiguration>();
-
-    hostServerSocketPortsMap = new Hashtable<String, List<SocketPortAllocation>>();
+    transMap = new ConcurrentHashMap<>();
```
ArtsiomYurhevich: Changed in new pr

ArtsiomYurhevich:
Added type arguments to <> in collections.  @mbatchelor, Could you please take a look for the comments above?

engine/src/org/pentaho/di/www/TransformationMap.java
mbatchelor: That is only necessary if you're not working with a List. My suggestion is to use a ConcurrentLinkedDeque, but that method returns a List. So my suggestion above is how you convert the Deque to a List and you do it by constructing a new ArrayList as I showed.

PR #3019, opened 2016-10-12
[BACKLOG-8212] Add Metadata Injection (MDI) support to the Join Rows …

AlexanderBuloichik:
…(cartesian product) step


AlexanderBuloichik:
Coverage calculation is wrong

PR #3024, opened 2016-10-12
[PDI-12503]-Fixed getting wrong status "Waiting" of finished transformation in the DI-Server server. Now when a transformation finishes it will return "Finished" status.

ArtsiomYurhevich:
Fixed

### DIFF
--- a/engine/src/org/pentaho/di/trans/Trans.java
+++ b/engine/src/org/pentaho/di/trans/Trans.java
@@ -47,7 +47,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


ArtsiomYurhevich:
@mbatchelor, @bmorrise, @e-cuellar Could you please take a look?

engine/src/org/pentaho/di/trans/Trans.java
```diff
@@ -308,10 +308,10 @@
   private volatile boolean preparing;
 
   /** Whether the transformation is initializing. */
-  private boolean initializing;
+  private volatile boolean initializing;
```
mbatchelor: Seems we are using three variables to describe the state of execution for the transformation. This seems really weird to me. Why not one AtomicInteger called state where 0 = Initializing, 1 = Running, 2 = Paused, and 3 = Finished. Or, if something can be both initializing and running, then at most 2 variables - executionState and executionPhase - executionState = Running, Paused, Finished and executionPhase = Preparation, Execution, Finalization. Something like that makes more sense to me.

ArtsiomYurhevich: @mbatchelor I suggest using Enums for execution state and execution phase: `enum ExecutionState { RUNNING, STOPPED, PAUSED}` and `enum ExecutionPhase {PREPARING, INITIALIZING, EXECUTING, STOPPED, FINISHED}` and use them instead of AtomicInteger fields because it  will help to avoid  redundant code to check if the number we set to this AtomicInteger values is valid. So what do you think about it?

ArtsiomYurhevich:
Fixed checkstyle.

lucboudreau:
```
    [javac] /home/buildguy/project/engine/src/org/pentaho/di/trans/Trans.java:4677: error: cannot find symbol
    [javac]     int exist = status.get() & PAUSED.mask;
    [javac]                                ^
    [javac]   symbol:   variable PAUSED
    [javac]   location: class Trans
```

ArtsiomYurhevich:
Fixed import

PR #3036, opened 2016-10-14
[BACKLOG-11582]-Escaping variable to correctly work with spaces

stepanovdg:
* To be able to return to folder which contains spaces in name


stepanovdg:
@rmansoor Could you please take a look?

PR #3037, opened 2016-10-14
[BACKLOG-10967] Displaying the version number only if versioning flag…

flbrino:
… is enabled

### DIFF
--- a/ui/src/org/pentaho/di/ui/spoon/Spoon.java
+++ b/ui/src/org/pentaho/di/ui/spoon/Spoon.java
@@ -5212,10 +5212,10 @@ public class Spoon extends ApplicationWindow implements AddUndoPositionInterface
 
             boolean versioningEnabled = true;
             boolean versionCommentsEnabled = true;
-            String fullPath = meta.getRepositoryDirectory() + "/" + meta.getName() + meta.getRepositoryElementType().getExtension();
+            String fullPath = getJobTransfFullPath( meta );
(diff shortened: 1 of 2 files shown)


flbrino:
@mkambol please review

ui/src/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -6760,20 +6791,24 @@ public void setShellText() {
...
-    }
     JobMeta jobMeta = getActiveJob();
+
     if ( jobMeta != null ) {
-      changed = jobMeta;
-      filename = jobMeta.getFilename();
-      name = jobMeta.getName();
-      version = jobMeta.getObjectRevision() == null ? null : jobMeta.getObjectRevision().getName();
+      metaProperties = getMetaProperties( rep, jobMeta );
+    } else if ( transMeta != null ) {
```
mkambol: Getting back an array of Objects feels really clunky.  
Is there a reason we can't just do something like

```java
    AbstractMeta meta = Objects.firstNonNull( getActiveJob(), getActiveTransformation() );
    changed = meta;
    filename = meta.getFilename();
    name = meta.getName();
    version = meta.getObjectRevision() == null ? null : meta.getObjectRevision().getName();
    versioningEnabled = isVersioningEnabled( meta );
```

That way we don't need to have to pass around an untyped Object array.

flbrino:
@mkambol suggestions added. For review

PR #3039, opened 2016-10-14
[BACKLOG-11622] Const.getHostname() is adds considerable delay in Spoon/Job/Trans startup/execution

mdamour1976:


mdamour1976:
@bmorrise

PR #3043, opened 2016-10-14
[BACKLOG-8210] Fixed issues with send-true and send-false step fields

e-cuellar:


mdamour1976:
@e-cuellar very minor (annoying) checkstyle issues, let me know when they are cleaned :)

PR #3049, opened 2016-10-18
[BACKLOG-11640] Update the Spoon help link to point to a page in the help system

davidmsantos90:
@pentaho/millenniumfalcon @pamval please review.


mdamour1976:
Use a URL fragment for the documentationUrl:
documentationUrl=0D0/Pentaho_Data_Integration

Then later, simply use Const.getDocUrl( url )

mdamour1976:
IMO, also, if we are no longer using InformationMap, that should be removed.

mdamour1976:
This might also be helpful to update:
https://github.com/pentaho/pentaho-kettle/blob/master/core/src/org/pentaho/di/core/messages/messages_en_US.properties

davidmsantos90:
@mdamour1976 So should we also remove the mechanism that provides local documentation altogether?

davidmsantos90:
In BACKLOG-11645 we have this:
"Requirement Note: This link needs to be configurable via a properties file or some other "non-compile" method so customers can change where this points to if they have internal help or are OEMs."

But by using Const.BaseDocUrl we lose the ability to change this without compiling the project.

Do you know a better place to configure this?

mdamour1976:
@davidmsantos90 Const.getDocUrl uses a properties file..  It has been accepted in the past as a "non-compile" option.  I don't know of a scenario where someone would use "different" documentation, and if they did, well, they can open the jar and change the properties file.

davidmsantos90:
@pentaho/millenniumfalcon @mdamour1976 @pamval please review

davidmsantos90:
@pentaho/millenniumfalcon @mdamour1976 @pamval doing the PR without changing welcome/index.html, wingman runs without any problems.

But when running with the file changed, it fails. This looks like a problem with wingman, but I'm not sure.
@lucboudreau can you also take a look at this error?

PR #3065, opened 2016-10-20
[BACKLOG-11693]Use default binary string conversion for Internet Address

hudak:
Removed override for binary string converter.
`InetAddress.getByAddress` expects a byte array representing the IP Address value, not an encoded String.

@matthewtckr -- this override was added with https://github.com/pentaho/pentaho-kettle/pull/2766/commits/162d73be33164980f9a3dcc5d8b7239d902f6d1d
I added more unit testing and checked the related cases for regression, everything seems to be okay. Is there anything else that should be tested?

### DIFF
--- a/core/src/org/pentaho/di/core/row/value/ValueMetaInternetAddress.java
+++ b/core/src/org/pentaho/di/core/row/value/ValueMetaInternetAddress.java
@@ -22,24 +22,23 @@
 
 package org.pentaho.di.core.row.value;
 
-import java.math.BigDecimal;
-import java.math.BigInteger;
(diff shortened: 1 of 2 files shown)


matthewtckr:
The removal looks ok to me, as ValueMetaBase.convertBinaryStringToNativeType() calls convertData(), which is still overridden in ValueMetaInternetAddress.

hudak:
@mbatchelor Please review, thanks!

pamval:
@hudak @mkambol this needs to go into 7.0.0.0.

PR #3067, opened 2016-10-24
[PPP-3571] Use of vulnerable component httpclient-4.2.5.jar - CVE-2014-3577

Advent51:
- version updated to 4.5.2


Advent51:
@rmansoor @mbatchelor please review

PR #3070, opened 2016-10-24
[BACKLOG-11862] - trapping a failed get input row for TableInput when…

kurtwalker:
… the step has already been cancelled


kurtwalker:
I know, no unit test.  I tried, but my mocks were getting so deep as to make the test totally meaningless

PR #3086, opened 2016-10-27
[BACKLOG-11771] PDI - renaming a folder in the repository home folder…

pedrofvteixeira:
… gets an error

- [CHERRY-PICK] https://github.com/pentaho/pentaho-kettle/pull/3085/commits/4d6149492f6994f2c578cc8dad30a272413684a8

	- PurRepository lost the ability to rename folder under /home because we now proxy-wrap it in a RepositorySessionTimeoutHandler
	- RepositoryExtended interface was created 8 months ago ( Feb 2016 ) to provide additional Repository methods on the next major release

- Validated in 7.0-SNAPSHOT built today ( Oct 27th )
- Note: while going through this issue on the 'rename' folder action, I noticed there's a mention that the same also occurs on the 'delete' folder action; looking at it, it was the exact same reason; this PR includes the (same) fix to the 'delete' folder action as well

### DIFF
--- a/engine/src/org/pentaho/di/repository/RepositoryExtended.java
+++ b/engine/src/org/pentaho/di/repository/RepositoryExtended.java
@@ -1,3 +1,25 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
+ * Copyright (C) 2002-2016 by Pentaho : http://www.pentaho.com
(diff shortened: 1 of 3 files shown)


pedrofvteixeira:
@dkincade , @kcruzada , @pamval please review

PR #3088, opened 2016-10-27
[BACKLOG-11939] - Drools upgrade bringing in vulnerable xstream jar

YuryBY:


pamval:
looks good, waiting for wingman

PR #3163, opened 2016-11-14
[BACKLOG-7056] Dataservices: Min() is not ignoring null values (same …

andrei-viaryshka:
…for max)

* switch off lazy conversion by default
* use valuemeta to check if value is null when lazy conversion is on


andrei-viaryshka:
@mkambol Could you please review?

engine/src/org/pentaho/di/trans/steps/csvinput/CsvInputMeta.java
```diff
@@ -125,7 +125,7 @@ public void setDefault() {
     delimiter = ",";
     enclosure = "\"";
     headerPresent = true;
-    lazyConversionActive = true;
+    lazyConversionActive = false;
```
mkambol: While I'd like to see the above become the default, I don't think we can make that change within the scope of this bug.

andrei-viaryshka:
@mkambol I've rolled back lazy conversion option change.

PR #3165, opened 2016-11-15
[PDI-15628] Results from jobs run in parallel should not be mixed in log files

AliaksandrShuhayeu:
[PDI-15628] Results from jobs run in parallel should not be mixed in log files

-added checking object name in LoggingObject equals()


AliaksandrShuhayeu:
I fixed checkstyle violations.

PR #3168, opened 2016-11-15
[PDI-14937] executors_output_step not cleared when a hop is deleted from the transformation executor step"

AliaksandrShuhayeu:
[PDI-14937] executors_output_step not cleared when a hop is deleted from the transformation executor step"

-added default implementation in interface


lucboudreau:
This PR is fine. Wingman was locked out by GitHub for spamming stats.

AliaksandrShuhayeu:

PR #3169, opened 2016-11-15
[SP-2993][PDI-14937] executors_output_step not cleared when a hop is deleted from the transformation executor step"

AliaksandrShuhayeu:
[SP-2993][PDI-14937] executors_output_step not cleared when a hop is deleted from the transformation executor step"
-added default implementation in interface


AliaksandrShuhayeu:

PR #3170, opened 2016-11-16
PDI-15515 - Changed implementation of encoder. Replaced legacy esapi …

ArtsiomYurhevich:
Changed implementation of encoder. Replaced legacy ESAPI encoder ( org.owasp.esapi ) by  org.owasp.encoder). Moved some tests for servlets from pentaho-kettle/test/org/pentaho/di/www/ to engine module.  In case the moved tests in pentaho-kettle/test/  aren't used they may be removed


core/src/org/pentaho/di/core/xml/XMLHandler.java
```diff
@@ -768,20 +768,19 @@ public static String getString() {
...
   public static String addTagValue( String tag, String val, boolean cr, String... attributes ) {
     StringBuilder value = new StringBuilder( "<" );
-    Encoder encoder = ESAPI.encoder();
-    value.append( tag );
+    value.append( Encode.forXml( tag ) );
 
     for ( int i = 0; i < attributes.length; i += 2 ) {
-      value.append( " " ).append( encoder.encodeForXMLAttribute( attributes[i] ) ).append( "=\"" ).append(
-        attributes[i + 1] ).append( "\" " );
+      value.append( " " ).append( Encode.forXml( attributes[i] ) ).append( "=\"" ).append(
```
mbatchelor: I believe you need forXmlAttribute here. The rules for an XML element are very different than the encoder rules for an XML attribute. You need to accommodate this difference in a Unit Test if it's not already covered. https://www.w3.org/TR/xml/#NT-AttValue - note you cannot have ampersand or open angle brackets, and you can't have the same character used to open the attribute name ( either ' or " depending on how the attribute is formed )

core/test-src/org/pentaho/di/compatibility/ValueTest.java
```diff
@@ -764,7 +764,7 @@ public void testGetXML() {
     vs6.setPrecision( 2 );
     result = vs6.getXML();
     assertEquals(
-      "<value><name>Name</name><type>Date</type><text>2006&#x2f;03&#x2f;01 17&#x3a;01&#x3a;02.005</text><length>-1</length><precision>2</precision><isnull>N</isnull></value>",
+      "<value><name>Name</name><type>Date</type><text>2006/03/01 17:01:02.005</text><length>-1</length><precision>2</precision><isnull>N</isnull></value>",
```
mbatchelor: So clearly, the old ESAPI encoder encoded colon( : ) and forward slash ( / ). I'm a bit worried about backward compatibility here if we have *any* persisted XML as it clearly won't be decoded back into colon and forward slash. Please get to the bottom of what in the whole system might have used this - if it's a method used in serializing a KTR/KJB, we could be creating a regression.

core/test-src/org/pentaho/di/core/xml/XMLHandlerUnitTest.java
```diff
@@ -148,7 +148,7 @@ public void addTagValueBoolean() {
 
   @Test
   public void addTagValueDate() {
-    String result = "2014&#x2f;12&#x2f;29 15&#x3a;59&#x3a;45.789";
+    String result = "2014/12/29 15:59:45.789";
```
mbatchelor: Same comment as above - was this method used to persist information, and if so, what happens when deserialized? I suspect it will be broken on deserialization.

mbatchelor:
Lots of test errors as well. Please fix up so I can process the PR

ArtsiomYurhevich:
#3215

mchen-len-son:
waiting for review and merge of #3215, which is fixing test failures.

pedrofvteixeira:
Merged https://github.com/pentaho/pentaho-kettle/pull/3215 , so I re-triggered wingman, so that we can get an accurate report.

@ArtsiomYurhevich, some questions/concerns in regards to this change: have we tested the impact of this change downstream ? 

1. with this PR, kettle-core now depends on _owasp.encoder_ instead of _owasp.esapi_ 👍 
2. But.. pentaho-platform/extensions depends on kettle-core ( with a default transitive=true ) 
2.1. https://github.com/pentaho/pentaho-platform/blob/7.0.0.1-R/extensions/ivy.xml#L233
3. pentaho-platform/assembly depends on extensions and all its _transitives_
3.1. https://github.com/pentaho/pentaho-platform/blob/7.0.0.1-R/assembly/ivy.xml#L20-L22
4. pentaho-platform-ee depends on all of the CE assemble **and** also declares _owasp.esapi_ as a dependency of its own
4.1. https://github.com/pentaho/pentaho-platform-ee/blob/7.0.0.1-R/bi-platform-ee-assembly/ivy.xml#L117

So... could platform EE assembly end up with both _owasp.encoder_ **and** _owasp.esapi_ in its WEB-INF/lib ? 🤔 

CC'ing @mbatchelor , @pamval , @duarteteixeira

ArtsiomYurhevich:
Hello @pedrofvteixeira ,
 The pull request with the encoder changes for pentaho-platform-ee has already been sent. See https://github.com/pentaho/pentaho-platform-ee/pull/1013 .But it isn't merged yet

pedrofvteixeira:
@ArtsiomYurhevich thanks.

I'm good with this PR.

Wingman triggered yesterday **after** https://github.com/pentaho/pentaho-kettle/pull/3215 got merged is also good.

@mbatchelor, care to do the honours ( should all look good to you too ) ? 

**Note**: this PR should be merged alongside https://github.com/pentaho/pentaho-platform-ee/pull/1013

CC'ing @duarteteixeira

PR #3172, opened 2016-11-18
PDI-15196 - Fixed skipping rows if json field is null, empty or if th…

ArtsiomYurhevich:
- Fixed skipping rows if json field is null, empty or if there are no matches with json fields and at the same time there are no additional not null row fields
- Corrected filtering out rows and avoiding excess processRow calls.

- Reworked process of result rows' storage and converting.

-  Changed 'Ignore Missing Path' behavior. Now it is unchecked it will throw an exception if there are at least one null json value in output row after parsing. Previously it failed only if  all the rows misses the same path.


ArtsiomYurhevich:
Please, do not merge it!

PR #3191, opened 2016-11-28
[PDI-15757] Add field "Ignore missing path" to step "Load file content in memory"

evere10:
Hi all,

I added field "Ignore missing path" to step "Load file content in memory", like similar field in step "JSON Input". Without this new option (or with this new option unchecked in step) step fails with such error:

2016/11/09 16:17:43 - Load file content in memory.0 - ERROR (version 6.1.0.1-196, build 1 from 2016-11-09 12.10.12 by everetennikov) : Couldn't open file #0 : file:///filename --> org.apache.commons.vfs2.FileSystemException: Could not determine the size of "file:///filename" because it is not a file.


mbatchelor:
Good evening Evgeny, thanks so much for the contribution. Unfortunately, I cannot process this PR until the corresponding test case is updated ( https://github.com/pentaho/pentaho-kettle/blob/master/engine/test-src/org/pentaho/di/trans/steps/loadfileinput/LoadFileInputMetaTest.java ). That test case validates that you can round-trip the settings from file to repository back to file again and vice-versa without losing any settings (loadSaveTester code).

evere10:
@mbatchelor I'm not quite sure, what should I do now. Seems like wingman doesn't complain about my commits in branch, but only about changes made in master during last two weeks? I didn't touch a lot of files listed in "Changed files".

PR #3201, opened 2016-11-28
[BACKLOG-12327] AVG function doesn't work for data services with CSV …

andrei-viaryshka:
…Input step with Lazy Conversion on


andrei-viaryshka:
@mkambol Could you review please?

mkambol:
@andrei-viaryshka - I notice some of the getXXX methods still just do a straight == null check, rather than isNull().   Is there a reason only Number, BigNumber and Date needed to be changed?

andrei-viaryshka:
@mkambol I've made changes/tests only for types that are used in other tests with lazy conversion.

andrei-viaryshka:
@mkambol Added the change for Integer as well. It looks like lazy conversion shouldn't be an issue for other types.

PR #3204, opened 2016-11-29
Sp 3004 - Backport of BACKLOG-7091 - Json Input step - Allow multiple paths to parse. At this time "We MUST have the same number of values for all paths."

ArtsiomYurhevich:


mchen-len-son:
Double-checked locally.
Tests are passing, checkstyle is OK.
Merging.

PR #3221, opened 2016-12-02
SP-3026 - Backport of PDI-15379 - Component failure "Input JSON" only in version 6.1.0.1-196. In the 5.2.0.0 version works correctly

ArtsiomYurhevich:
Skip checking equality of json results size if it's size is 0


mchen-len-son:
Checked locally. Looks good.
Waiting for wingman.

PR #3223, opened 2016-12-02
SP-3069 - Backport of PDI-15196 - JSON Input Step does not ignore missing paths

ArtsiomYurhevich:
Fixed skipping rows if json field is null, empty or if there are no matches with json fields and at the same time there are no additional not null row fields. Corrected filtering out rows


mchen-len-son:
Checked locally, PR looks good.
Waiting for wingman results.

ArtsiomYurhevich:
@mbatchelor Could you review please

PR #3230, opened 2016-12-05
[SP-3055][PDI-14797] Get Filenames Step issue with VFS and WinSSHD

AliaksandrShuhayeu:
- reverted because is not full fix


mchen-len-son:
Checked locally. Revert looks OK.
Merging

PR #3240, opened 2016-12-09
[PDI-14621] JSON Output creates empty file when it receives empty str…

stanislau-strelchanka:
[PDI-14621] JSON Output creates empty file when it receives empty stream (with option "Append" and "Do not create file...")


stanislau-strelchanka:
@mbatchelor, @matthewtckr, @pamval , @mchen-len-son, @tgf, @mdamour1976, @bmorrise please review

stanislau-strelchanka:
@mbatchelor, @matthewtckr, @pamval , @mchen-len-son, @tgf, @mdamour1976, @bmorrise please review

PR #3242, opened 2016-12-12
[PRD-5778]-JDBC connection details still remains even after removing from Options tab for MsSQL connections.JDBC connection details still remains

denisprotopopov:
- set using pluginId insteadof constant
 - add unit test

### DIFF
--- a/core/src/org/pentaho/di/core/database/DatabaseMeta.java
+++ b/core/src/org/pentaho/di/core/database/DatabaseMeta.java
@@ -2752,7 +2752,7 @@ public class DatabaseMeta extends SharedObjectBase implements Cloneable, XMLInte
     // This is also covered/persisted by JDBC option MS SQL Server / instancename / <somevalue>
     // We want to return <somevalue>
     // --> MSSQL.instancename
-    return getExtraOptions().get( "MSSQL.instance" );
+    return getExtraOptions().get( getPluginId() + ".instance" );
   }
 
   /**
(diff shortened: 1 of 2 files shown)


denisprotopopov:
@pamval @mchen-len-son please review the changes

tmorgner:
I dont think that code actually changes anything. It just takes the hardcoded plugin id "MSSQL" and replaces it with a non-hardcoded one. But the property in question is only valid for MSSQL database connections, so the code will now still work largely as before. I cannot see how it is going to fix the PRD problem.

As PRD stores the properties for connections outside of the database-meta interface, the error probably lies there instead. DatabaseMeta is only used for feeding the dialog, after that we extract all properties and set them as JDBC properties on the Driver-connection  definition. I would assume it is that code (in PRD) that does not handle the case where a property has gone away.

denisprotopopov:
As I understand user can create 2 types of connection for MS SQL (MS SQL SERVER and MS SQL SERVER Native). They have different pluginIds.
Connection for MS SQL Server (Native) has PluginId=MSSQLNATIVE so in https://github.com/pentaho/pentaho-reporting/blob/master/designer/datasource-editor-jdbc/source/org/pentaho/reporting/ui/datasources/jdbc/ui/XulDatabaseDialog.java#L294 
We received realkey='ce' instead of 'instance' for key=MSSQLNATIVE.instance

tmorgner:
Wow, ok, that is a connection so strange that I would not have foreseen it. I stand corrected.

PR #3262, opened 2016-12-17
PDI-15661 - Spoon Crashed when Logging is Removed. Since no one optio…

ArtsiomYurhevich:
…n is selected it created a non-sized array(length == 0) for the options and trying to display it in table in "logging" tab in "executions resuts" caused  java.lang.ArrayIndexOutOfBoundsException as it tried to get the first element from the array. And also fixed unexpected additional row appending when you click on checkbox "Include?" in Transformation properties in "logging" tab


ArtsiomYurhevich:
@mbatchelor , @pamval  Could you take a look please.

PR #3267, opened 2016-12-20
[ENGOPS-2667] Removing antcontrib targets(part II)

abaskakau:
@lgrill-pentaho @lucboudreau


lgrill-pentaho:
Don't push non-SP change into pending release branch.

PR #3269, opened 2016-12-20
[ENGOPS-2667] Removing antcontrib targets(part II)

abaskakau:
@lgrill-pentaho @lucboudreau


lgrill-pentaho:
Do not push into the pending release build branch.

PR #3270, opened 2016-12-20
[PDI-14650] In csv input when using multiple character delimiter and an

mchen-len-son:
enclosure character fields lose last character
 - fixed multi character delimiter processing in CsvInputStep
 - added unit tests


mchen-len-son:
@mbatchelor @e-cuellar @bmorrise @wseyler please review

PR #3271, opened 2016-12-20
[PDI-14449]  [PDI-14417] [PDI-14418] 

AliaksandrShuhayeu:
[PDI-14449] Filter Rows step and Kettle date variables,
[PDI-14417] Filter rows step behaviour with inputting dates - problem with EnterValueDialog,
[PDI-14418] Filter Rows step is not displaying date values after closing then re-opening a transformation

-added reading date in compatible format (how it was saved)
-replaced modify listener on focus lost listener
-added test
-added power mock dependency


AliaksandrShuhayeu:

PR #3278, opened 2016-12-22
[TEST] - Unit test for JoinRows step as a part of PDI-14994.

ArtsiomYurhevich:


engine/src/org/pentaho/di/trans/steps/joinrows/JoinRows.java
```diff
@@ -109,7 +109,11 @@ public void initialize() throws KettleException {
 
       for ( int i = 1; i < getInputRowSets().size(); i++ ) {
         String directoryName = environmentSubstitute( meta.getDirectory() );
-        data.file[i] = File.createTempFile( meta.getPrefix(), ".tmp", new File( directoryName ) );
+        File file = null;
+        if ( directoryName != null ) {
+          file = new File( directoryName );
+        }
+        data.file[i] = File.createTempFile( meta.getPrefix(), ".tmp", file );
```
ArtsiomYurhevich: Made it for test purposes , it doesn't affect the behaviour

engine/src/org/pentaho/di/trans/steps/joinrows/JoinRows.java
```diff
@@ -446,7 +450,9 @@ public void dispose( StepMetaInterface smi, StepDataInterface sdi ) {
     // Remove the temporary files...
     if ( data.file != null ) {
       for ( int i = 1; i < data.file.length; i++ ) {
-        data.file[i].delete();
+        if ( data.file[i] != null ) {
+          data.file[i].delete();
+        }
```
ArtsiomYurhevich: I have faced the issue when you don't define a temp directory or the path isn't found then the transformation hangs for ever. It is caused by NPE in finally block https://github.com/pentaho/pentaho-kettle/blob/master/engine/src/org/pentaho/di/trans/step/RunThread.java#L96

PR #3287, opened 2016-12-23
[SP-3136][BACKLOG-13476] Fix wrong chars around yen sign

mchen-len-son:


mchen-len-son:

This is a backport of #3259.

PR #3245, opened 2016-12-12
[PDI-15870] Incorrect error when verifying step

matthewtckr:

PR #3288, opened 2016-12-23
[SP-3134][PDI-15870] Incorrect error when verifying step

mchen-len-son:


mchen-len-son:

This is a backport of #3245.

PR #3293, opened 2016-12-23
[SP-3125][PRD-5778]-JDBC connection details still remains even after removing from Options tab for MsSQL connections.JDBC connection details still remains

denisprotopopov:
- set using pluginId insteadof constant
 - add unit test


mchen-len-son:
@denisprotopopov please, fix the checkstyle violations

denisprotopopov:
@mchen-len-son done

PR #3320, opened 2016-12-29
[PDI-13754] LDAP Input Step - Using variable for "Trust store path"

AliaksandrShuhayeu:
[PDI-13754] LDAP Input Step - Using variable for "Trust store path"
-added resolving variable for trust store path
-added resolving and decoding variable for Trust store password


pamval:
@AliaksandrShuhayeu, can we write an unit test to cover this change?

AliaksandrShuhayeu:
@mbatchelor @pamval I have added tests.
Could you please review and merge it?

PR #3247, opened 2016-12-13
[PDI-14420] - Excel Output with Append option is not working

YuryBY:
@mbatchelor, @duarteteixeira, could you please take a look?

PR #3330, opened 2016-12-30
[Sp-3172]

YuryBY:
@mbatchelor, @mchen-len-son, could you please take a look?


mchen-len-son:
This is a backport of #3247 and #3328.

PR #3331, opened 2016-12-30
Sp 3152

mchen-len-son:


mchen-len-son:
This one is made upon #3314, also containing checkstyle violation fixes.

PR #3332, opened 2016-12-30
[SP-3175][PDI-15909] PDI-15781 fix breaks remote execution on carte s…

mchen-len-son:
…erver


mchen-len-son:
@mbatchelor @pamval can you please take a look?

This is a backport of regression fix for PDI-15909 - #3327.

PR #2014, opened 2015-12-30
Pdi 14763

yshakhau:
Commit PDI-14724 must be merged before commit PDI-14763

PR #3338, opened 2017-01-05
[PDI-14763] - Disabling error handling hop doesn't disable error hand…

mbatchelor:
…ling of the step. Replaces first commit on #2014


mbatchelor:
Note - fixed checkstyle, TextFileInputMetaLoadSaveTester issue is a non-deterministic failure documented on PDI-15930

PR #3346, opened 2017-01-06
[PDI-12263] - MondrianInput should release resources as soon as possible

mbatchelor:
[PDI-14120] - MondrianInput should provide linenr logging


mbatchelor:
@lucboudreau - please review and merge!

PR #3348, opened 2017-01-06
[PDI-2286] Sort steps before saving to a file

HiromuHota:
This patch ensures steps are sorted before saving to a file to become VCS (like git) friendly.


mrjefflewis:
+1 for this PC. PDI has needed this fixed for years

PR #3349, opened 2017-01-09
fix configuration problem

weipeilun:
The pool parameters set in properties should be set as config of GenericObjectPool. Otherwise they won't work.
It also affects all versions before 7.0


mbatchelor:
@weipeilun - Please file a JIRA case at http://jira.pentaho.com to report the defect. Also, I need to get a test case on this before I can merge it.

mbatchelor:
No test case, doesn't compile, not responsive to requests.

PR #3351, opened 2017-01-09
[BACKLOG-13475] Fix misspelled MDI Key

AlexanderBuloichik:


mdamour1976:
@AlexanderBuloichik does this break previously saved KTR's with MDI?

AlexanderBuloichik:
@mdamour1976 Yes. You added MDI at 2016-05-26, i.e. before 7.0.
We don't have functionality for MDI migration, but we could think about it.

PR #3360, opened 2017-01-10
PDI-15648 - Group by step - Minimum aggregation doesn't work when null value in first row

ArtsiomYurhevich:
In case "minNullIsValued" is false, firstly, we need to set the first non-null value for further comparing


ArtsiomYurhevich:
@mbatchelor, @bmorrise , @e-cuellar , Could you, please, take a look?

ArtsiomYurhevich:

PR #3370, opened 2017-01-12
[SP-3171] - Backport of PDI-14420 - Excel Output with Append option i…

YuryBY:
…s not working (7.0 Suite)

 - additional fix


YuryBY:

PR #3371, opened 2017-01-12
[PDI-14420] - Excel Output with Append option is not working

YuryBY:
- additional fix


YuryBY:

PR #3373, opened 2017-01-12
[SP-3172] - Backport of PDI-14420 - Excel Output with Append option i…

YuryBY:
…s not working (6.1 Suite)

 - additional fix


YuryBY:

PR #3376, opened 2017-01-13
[PDI-13487]Excel Output field formats are ignored when "Split Every Rows" is used

AliaksandrShuhayeu:
[PDI-13487]Excel Output field formats are ignored when "Split Every Rows" is used….. Rows" is used

-added clearing formats in closing file, so for next file format will be reinitialize and write in next file
-added test

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/exceloutput/ExcelOutput.java
+++ b/engine/src/org/pentaho/di/trans/steps/exceloutput/ExcelOutput.java
@@ -681,7 +681,7 @@ public class ExcelOutput extends BaseStep implements StepInterface {
         }
 
       }
-      // data.formats.clear();
+      data.formats.clear();
       if ( log.isDebug() ) {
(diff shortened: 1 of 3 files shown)


AliaksandrShuhayeu:

engine/src/org/pentaho/di/trans/steps/exceloutput/ExcelOutput.java
```diff
@@ -681,7 +681,7 @@ private boolean closeFile() {
         }
 
       }
-      // data.formats.clear();
+      data.formats.clear();
```
mbatchelor: Important note - that line was commented out when the "Append in Excel Output" feature was created - SVN revision 6294 on 12/31/2007. So when testing, you have to test all aspects of this step including appending.

AliaksandrShuhayeu: @mbatchelor I double-checked this step Excel Output. My fix does not affect options: append and append to Excel template. After merging PR I going to add this affected area in jira and QA will check it one more time.

PR #3377, opened 2017-01-13
Created API for reporting metrics and status

hudak:
This is the first pass at a progress reporting API, please send your comments!


pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/IExecutionContext.java
```diff
@@ -1,17 +1,26 @@
...
   ITransformation getTransformation();
 
+  /**
+   * Transformation parameters are the preferred way of controlling execution.
+   * Maybe arguments can be rolled into environment?
+   */
+  @Deprecated
   String[] getArguments();
+
+  CompletableFuture<IExecutionResult> execute();
```
pentaho-nbaker: just Future right? should only be referenced as CompletableFuture internally

pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/IOperation.java
```diff
@@ -11,9 +12,9 @@
  * which is associated with the behavior specific to
  * an IEngine.
  */
-public interface IOperation extends Serializable {
+public interface IOperation extends IReportingEventSource {
 
-  String getId();
+  @Override String getId();
```
pentaho-nbaker: Can probably remove here I would think

mkambol: @pentaho-nbaker - remove getId()?  Don't we need some way of identifying the operation?

pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/reporting/IProgressReporting.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.function.Consumer;
+
+/**
+ * Created by hudak on 1/11/17.
+ */
+public interface IProgressReporting {
+  <S extends IReportingEventSource, D extends Serializable>
+  Publisher<IReportingEvent<S, D>> eventStream( S source, Class<D> type );
+
+  Collection<IReportingEventSource> getReportingSources();
```
mkambol: Could this be a Set?

pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/IExecutionContext.java
mkambol: Should IExecutionContext have .execute()?  Or should an IEngine be given a context in which to execute?

pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/IOperation.java
pentaho-nbaker: IReportingEventSource already has getId();

pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/IExecutionContext.java
hudak: I prefer the Java 8 CompletableFuture, since it allows for callbacks in addition to blocking. Could switch to a CompletionStage if you prefer
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html

pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/IOperation.java
hudak: This is the getId() required by IReportingEventSource

pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/reporting/IProgressReporting.java
hudak: Yes, could be. Didn't want to assume the implementations had save `hashCode` and `equals` implementations, though we could prescribe that.

pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/IExecutionContext.java
hudak: @mkambol Execute was originally on IEngine, but I figured you would never want to execute on a context from a different engine, so this is safer. In the engines we were casting back into concrete implementations.

This could also allow allow for a builder-style fluent syntax `engine.prepare(someStuff).setParameters(yadaYada).execute() `

pentaho-nbaker: I'm good with both changes.

You shouldn't ever execute anything but the ExecutionContext given by the engine in prepare, so this makes sense.

pdi-execution-engine/api/src/main/java/org/pentaho/di/engine/api/IOperation.java
hudak: Ah, see what you mean. getId is inherited so can be removed from this interface, will do

PR #3378, opened 2017-01-16
BACKLOG-11920 - Using DET with a trans loaded via open URL

flbrino:
If connected to a repository must import/save file before using DET


flbrino:
@mkambol for validation. Thks.

PR #3379, opened 2017-01-16
[PDI-13930]Get Files from Header Row doesn't load more than 256 fields in Excel Input step

AliaksandrShuhayeu:
[PDI-13930]Get Files from Header Row doesn't load more than 256 fields in Excel Input step

-extract method for processing workbook
-deleted hardcoded limit 256
-added test


AliaksandrShuhayeu:

PR #3380, opened 2017-01-16
all databases are equal?

Nirkus:
works for "them".. *SCNR*


Nirkus:
@mbatchelor I am curious, though - did I overlook something or did you?

All databases are equal when compared by this `euqals()` code, aren't they?
```
public boolean equals( Object obj ) {
  Database other = (Database) obj;
  return other.databaseMeta.equals( other.databaseMeta );
}
```

PR #3394, opened 2017-01-18
[BACKLOG-14128] Solved - Error runing DET opening a transformation vi…

flbrino:
…a Open URL before connecting to repository & Unit Tests

If connecting to a repository and selected the option not close tabs, then mark the transformation tabs as changed for saving on repository before executing.
Line 8160: Solved similiar situation when opening last used files at PDI start... If last used files were local and connecting to a repository at PDI start same error occured


flbrino:
@mkambol please review. Thks.

PR #3402, opened 2017-01-20
[BACKLOG-14131] Part 1- Return false when cancel button pressed, solu…

flbrino:
…tion for Cancel button when prompted for save doesn't cancel the operation


flbrino:
@mkambol BACKLOG-14131 Part I, please reaview. Thks.

PR #3405, opened 2017-01-21
PDI-15877 - Enclosure character is not removed from output stream

ArtsiomYurhevich:
https://github.com/pentaho/pentaho-kettle/blob/master/engine/src/org/pentaho/di/trans/steps/csvinput/CsvInput.java#L536 Here the facts that enclosure character(s) may be represented not only as 1 byte and that enclosure can consist of more than one character weren't taken into account


ArtsiomYurhevich:

PR #3406, opened 2017-01-21
[BACKLOG-14131] Part 2 - If the operation cancelled return

flbrino:


flbrino:
@mkambol BACKLOG-14131 Part II, please reaview. Thks.

PR #3407, opened 2017-01-21
PDI-13841 - Step "Get table names" fails for names that must be enclo…

ArtsiomYurhevich:
…sed in double-quotes (with special characters)


ArtsiomYurhevich:

PR #3424, opened 2017-01-24
[BACKLOG-14037]  Wiring "Step Metrics" tab to engine events.

mkambol:
@CodeOnCoffee @hudak @ccaspanello


hudak:
Using TransAdapter strategy instead

PR #3425, opened 2017-01-25
[SP-3249][PDI-13930]Get Files from Header Row doesn't load more than …

mchen-len-son:
…256 fields in Excel Input step

-extract method for processing workbook
-deleted hardcoded limit 256
-added test


mchen-len-son:
@pamval @mbatchelor please review.
thank you.

mchen-len-son:
@kolinus please take a look

PR #3432, opened 2017-01-25
[SP-3247][PDI-13487]Excel Output field formats are ignored when "Split Every ... Rows" is used

AliaksandrShuhayeu:
[SP-3247][PDI-13487]Excel Output field formats are ignored when "Split Every ... Rows" is used

-added clearing formats in closing file, so for next file format will be reinitialize and write in next file
-added test


mchen-len-son:
Looks ok. The failing test is a non-deterministic one.

PR #3441, opened 2017-01-27
[BACKLOG-14076,BACKLOG-14389] Hide the browser instead of closing it when leaving DET

afrjorge:
- refresh the browser after it starts loading the most recently set URL


afrjorge:

PR #3442, opened 2017-01-27
BACKLOG-14326 Solved prompted to overrwite on save & entering DET

flbrino:


flbrino:
@mkambol please review. Thks

flbrino:
Testing if the transformation/job exists on the repository if doesn't exists must clean the file path so doesn't try to use local path [BACKLOG-11870]. If exists just save the transformation.

PR #3445, opened 2017-01-27
[BACKLOG-13507] Fix cluster refresh of steps that don't refresh when …

tkafalas:
…cluster is changed.


tkafalas:
@rmansoor The test failure is pre-existing and had been appearing since build 1064 on Jan 23.  I notified Mike and Ben is working on a fix.

PR #3446, opened 2017-01-29
PDI-14840 - Having two syslog steps in transformation makes it running forever

ArtsiomYurhevich:
"running forever" was caused by an npe exception in dispose method in finally block which prevented transformation from proper finishing. Shutdown method in 0.9.34 in syslog api isnt thread safe and is not intended to be used multiple times . the method in 0.9.46 is also non fully thread safe. https://github.com/pentaho/pentaho-kettle/blob/master/engine/src/org/pentaho/di/trans/steps/syslog/SyslogMessage.java#L187


kolinus:
@mdamour1976, @pamval  could you please take a look

PR #3460, opened 2017-01-31
PDI-12261 - Microsoft Excel Writer - Appending lines to existing sheet

ArtsiomYurhevich:
The problem was : getLastRowNum() returns row number of the last row in sheat beginning from 0, so the first row( header ) has 0 number. And then https://github.com/pentaho/pentaho-kettle/blob/master/engine/src/org/pentaho/di/trans/steps/excelwriter/ExcelWriterStep.java#L827-L829  we dont increment the index to the next row.


kolinus:
@mdamour1976 @pamval could you please review

PR #3472, opened 2017-01-31
[BACKLOG-14307]Remote engine client

hudak:
@CodeOnCoffee


pentaho-nbaker:
How is this registered? I don't see annotations or an addition to the kettle-engines.xml

PR #3478, opened 2017-02-01
PDI-14745

vgarciabnz:
Add PATCH method in REST-Client


vgarciabnz:
Hi! This pull request just adds a minor improvement (but important for us). The change is so small that I thought no further tests were needed.
If there is something I can do to complete the pull request, please let me know.

PR #3479, opened 2017-02-02
[PDI-14334] Copy Files - overwrites files if subfolders used

SergeyTravin:
find file in subfolder, relative to destination directory
add unit and integration tests
rename to camel case


mchen-len-son:
Looks good from my side.

@mbatchelor can you please take a look?
thanks.

PR #3480, opened 2017-02-02
[PDI-13103] Flag wasStarted in MappingData never set to true

AliaksandrShuhayeu:
-added setting flag in true ( in consistency with SimpleMapping )
-added correct processing this flag in dispose() ( in consistency with SimpleMapping )
-added removing from collection active transformation not only in case waiting sub transformation


kolinus:
@mkambol @mdamour1976 could you please review

PR #1664, opened 2015-08-25
[BACKLOG-3826] - Implemented universal directory variable and created fallbacks for legacy variables.

mdamour1976:

PR #3481, opened 2017-02-02
[CDA-91] - As a Developer I can't invoke Jobs or Transformations from…

YuryBY:
… "Initial Transformation"

### DIFF
--- a/engine/src/org/pentaho/di/trans/steps/transexecutor/TransExecutorMeta.java
+++ b/engine/src/org/pentaho/di/trans/steps/transexecutor/TransExecutorMeta.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2016 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2017 by Pentaho : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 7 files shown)


engine/src/org/pentaho/di/trans/steps/transexecutor/TransExecutorMeta.java
```diff
@@ -608,120 +599,7 @@ public void getFields( RowMetaInterface inputRowMeta, String name, RowMetaInterf
...
-
-    // Pass some important information to the mapping transformation metadata:
-    //
-    mappingTransMeta.copyVariablesFrom( space );
-    mappingTransMeta.setRepository( rep );
-    mappingTransMeta.setMetaStore( metaStore );
-    mappingTransMeta.setFilename( mappingTransMeta.getFilename() );
-
-    return mappingTransMeta;
+    return loadMappingMeta( executorMeta, rep, null, space );
```
mbatchelor: I've compared all the load*Meta methods you replaced with the one in this new super-class, and all of the others compare well with your new method. This one here though is vastly different than the others. Line 618/619 (on the left) aren't seemingly done in the new class and neither are the lines in the FILENAME: case. Not saying that it isn't correct getting them all to do the same thing, but I'm concerned about regressions here.

YuryBY: @mbatchelor, yes you are right. But as far as I understand code above makes the same thing like was done by  Mike here: https://github.com/pentaho/pentaho-kettle/pull/1664/files#diff-e56f34b7521e57126c305cd757b1c332R633. I have moved it all to one place in order to avoid duplications. lets wait for @mdamour1976, to be sure that I did not miss something.

kolinus: @mdamour1976 could you please review?

YuryBY:
@mdamour1976, any updates?

mdamour1976:
Generally, this is a pretty good cleanup and it will provide benefit to r2d2 as we look at the mapping step and simple mapping steps.

I am curious about two things, mostly to address Marc's concerns.  First, there are not any unit tests, and second, I'd like to know how much regression testing we've done with samples.

Have we addressed all of the wingman issues?  (looks like some - yes)  I didn't pull it down locally, but wingman was not compiling..

YuryBY:
@mbatchelor, @mdamour1976, the PR was reworked. Now https://github.com/pentaho/pentaho-kettle/pull/3481/files#diff-317482073dea594569a507d4525ff757R55 contains your, Mike, changes. Thanks.

PR #3483, opened 2017-02-02
[BACKLOG-14287] - Add new property to the transformation to be loaded…

marcoslarsen:
…/written from/to the ktr that will be used for det state persistence.

@pentaho/kashyyyk @mdamour1976 @graimundo @pentaho/r2d2 @pamval Please review.


bmorrise:
Is there any reason we can't use the attributesMap from the AbstractMeta: https://github.com/pentaho/pentaho-kettle/blob/master/engine/src/org/pentaho/di/base/AbstractMeta.java#L149

PR #3488, opened 2017-02-06
PDI-14840 - Having two syslog steps in transformation makes it runnin…

ArtsiomYurhevich:
…g forever


mchen-len-son:
concurrent test failure is not related to the changes.

PR #3496, opened 2017-02-07
[BACKLOG-14601] - Change URL and modify text position

eliofreitas:


eliofreitas:

annapopova1:
It means to change doc URL we need rebuild kettle instead of just change properties file. Is it correct?

eliofreitas:
Yes, the idea is to eliminate the redundant url that was added in the properties file and instead use the function getDocUrl to obtain it

PR #3504, opened 2017-02-08
[PDI-15934]-Table Output step generated DDL without data type (MS SQLServer)

denisprotopopov:
- for BigNumber, Number shown Float(53), Binary shown VARBINARY(MAX), InternetAddress shown UNKNOWN
 - added testcase for Binary


denisprotopopov:

PR #3513, opened 2017-02-08
[BACKLOG-14363,DOC] converted comment into javadoc

afrjorge:


afrjorge:

annapopova1:
@afrjorge do you know how I can reproduce the case when setUrl works incorrectly and we need use another variable to store the value (lastNavigateURL)?

afrjorge:
Revert e68db70322d3a2daab6f6de209f2b7921a66ab32, explore two steps in PDI, copy the second step URL and open in an external browser, check that the URL for the second or consecutive steps are always equal to the one of the first step.

annapopova1:
FYI can't reproduce it on Windows. Probably OS specific behavior.

afrjorge:
Updated comment to reflect that.

PR #3514, opened 2017-02-10
[PDI-15929] Salesforce Input step throws a Incompatible class exception when run on server in 7.0

AliaksandrShuhayeu:
-set classloader for current thread when lib force-wsc.jar was called.

Library force-wsc.jar ( method com.sforce.ws.bind.TypeMapper#load ) loads class with  classloader of thread. The thread for server case has web class loader but Salesforce classes are loaded by plugin class loader. So we have a problem here com.sforce.ws.bind.TypeMapper#readSingle.
For fixing I used pentaho standart approach like here https://github.com/pentaho/pentaho-hadoop-shims/blob/master/common/hadoop-shim/src/main/java/org/pentaho/hadoop/shim/common/HadoopShimImpl.java#L72-L82. When SalesforceInput calls force-wsc.jar we set classloader for thread what SalesforceInput was loaded.

### DIFF
--- a/plugins/salesforce/src/org/pentaho/di/trans/steps/salesforceinput/SalesforceInput.java
+++ b/plugins/salesforce/src/org/pentaho/di/trans/steps/salesforceinput/SalesforceInput.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2016 by Pentaho : http://www.pentaho.com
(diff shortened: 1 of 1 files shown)


matthewtckr:
Couldn't the old jar just be pulled from the server platform?

AliaksandrShuhayeu:
Sent new PR https://github.com/pentaho/pentaho-platform/pull/3418

PR #3516, opened 2017-02-10
[BACKLOG-14633]Updated remote API for client-server model

hudak:
@CodeOnCoffee


pdi-engine-api/api/src/main/java/org/pentaho/di/engine/api/remote/Execution.java
```diff
@@ -0,0 +1,36 @@
...
+  ExecutionRequest getRequest();
+
+  /**
+   * Update a {@link LogicalModelElement} from this request.
+   * This will be routed to through the event pipe to the client.
+   *
+   * @param sourceId {@link LogicalModelElement#getId()}
+   * @param value    Serialized value of {@link PDIEvent#getData()}
+   */
+  void update( String sourceId, byte[] value );
```
pentaho-nbaker: public interface Execution<T extends Serializable>
update( T payload )

Then the client impl can bind it to TunneledPayload ??

PR #3522, opened 2017-02-14
[BACKLOG-14733] Fix job and transformation labels

AlexanderBuloichik:


engine/src/org/pentaho/di/job/entries/trans/messages/messages_en_US.properties
```diff
@@ -55,7 +55,7 @@ JobTrans.RunTransInCluster.Label=Run this transformation in a clustered mode?
 JobTrans.LogfileExtension.Label=Extension\:
 JobTrans.SelectTransRep.Tooltip=Select a transformation to run from the current repository.
 JobTrans.Error.ChekingParentLogFolderTitle=Error
-JobTrans.Fields.Argument.Label=Argument
+JobTrans.Fields.Argument.Label=Arguments
```
bmorrise: We probably need to add an additional property because JobTrans.Fields.Argument.Label is used for the table heading and it should still be singular.

AlexanderBuloichik:
@bmorrise fixed

PR #3543, opened 2017-02-17
[PDI-16029] fix setting absolute path if file is zip

SergeyTravin:


SergeyTravin:
@mchen-len-son, please review PR. Thanks!

SergeyTravin:
@mchen-len-son, @mbatchelor  please review

PR #3566, opened 2017-02-20
SP-3257-fix - Backport of PDI-15877 - Enclosure character is not removed from output stream 

ArtsiomYurhevich:


ArtsiomYurhevich:
SP cases for PDI-15922 ( SP-3273 and SP-3274 ) should be merged after PDI-15877 SP cases( SP-3257 and SP-3258 ) but SP-3273 and SP-3274 were already merged before SP-3257 and SP-3258 . Solution for PDI-15922 is based on PDI-15877 ,so, providing additional pull request with the required changes for this sp case

PR #3567, opened 2017-02-20
[BACKLOG-14180] shared object synchronization with repository and name changes

tgf:


tgf:
@bmorrise please review

PR #3572, opened 2017-02-21
[CDA-91] As a Developer I can't invoke Jobs or Transformations from "Initial Transformation"

AliaksandrShuhayeu:
[CDA-91] As a Developer I can't invoke Jobs or Transformations from "Initial Transformation"

-fixed variable with class name for subclasses
-fixed message for exception loading sub transformation


AliaksandrShuhayeu:

PR #3575, opened 2017-02-21
[BACKLOG-14832] missing objects from legacy repositories

tgf:


tgf:
@bmorrise please review

PR #3586, opened 2017-02-22
[BACKLOG-11126][BACKLOG-11425] - Updated ExpandedContentManger for OSX

wseyler:
This fixes an SWT browser issue on the Mac.  Now when you on the mac when the user goes into DET if there is an existing execution panel open it is hidden so that artifacts from it won't "come through" to the DET screen.  When leaving DET it is set visible again.  The additional code is only active on OS X. @mdamour1976 @bmorrise


wseyler:
Fixed unit test.  @mdamour1976 @bmorrise

PR #3590, opened 2017-02-23
[BACKLOG-14634] Restoring CompletableFuture to API now that RSA has f…

hudak:
…ixed compatability


hudak:
Use https://github.com/pentaho/pentaho-kettle/pull/3591 instead

PR #3594, opened 2017-02-24
[BACKLOG-14733] Fix job and transformation table header label

AlexanderBuloichik:


bmorrise:
I discussed this bug with UX and to keep consistency with other table headers we are going to leave this as "Argument" to keep it consistent with the rest of the tables as they have singular column headers.

PR #3599, opened 2017-02-24
[BACKLOG-14873] - Provide a default mask when none is provided to prevent a NPE

rmansoor:


rmansoor:
@davidmsantos90 @aminmkhan , do you folks see any issues with this default mask. Please provide your feedback and review this PR

core/src/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -1113,7 +1113,9 @@ public String getFormatMask() {
 
     if ( !Utils.isEmpty( this.conversionMask ) ) {
       mask = this.conversionMask;
-    }
+    } else {
```
davidmsantos90: This function is at the *base class* so there is no guarantee that the default mask to apply should be the DEFAULT_DATE_MASK.

To this specific case we could override this method in the String meta value class to provide the default date mask if it was doing the String -> Date conversion

rmansoor:
@davidmsantos90 thanks for the review. I will  look into that option

PR #3614, opened 2017-03-02
[PDI-15973]   Salesforce Input step no longer allows for relationship…

tim-ryzhov:
… queries

* Proper selection of child elements value
* Unit - test


mbatchelor:
Failing test was fixed in PDI-15900.

PR #3616, opened 2017-03-02
[BACKLOG-11769] PDI - Pan is unable to run a zip File exported as

pdesai16:
"File>Export>Linked Resources to XML" returns [error: can't open file]
    Currently the code uses the regular expression to identify if there is
    "zip" in the name. Now we check to see if the fileName starts with
    protocol like "Zip". If is does not, checks to see if the path is relative
    -Added a utility method in KettleVFS
    -Added test case for new method in KettleVFS


pdesai16:
@pentaho/rogueone  Please review

pentaho-mracine:
Looks good to me!

pdesai16:
@lucboudreau  code coverage drop does not look related to the change. Is it ok to merge?

PR #3618, opened 2017-03-02
[BACKLOG-14834] Add a DET visual indicator to the rendering of a step…

ddiroma:
… on the canvas

@bmorrise

### DIFF
--- a/engine/src/org/pentaho/di/trans/TransPainter.java
+++ b/engine/src/org/pentaho/di/trans/TransPainter.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2016 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2017 by Pentaho : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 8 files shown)


core/src/org/pentaho/di/core/extension/KettleExtensionPoint.java
```diff
@@ -39,6 +39,7 @@
     TransPainterStep( "TransPainterStep", "Draw additional information on top of a transformation step icon" ),
     TransPainterFlyout( "TransPainterFlyout", "Draw step flyout when step is clicked" ),
     TransPainterFlyoutTooltip( "TransPainterFlyoutTooltip", "Draw the flyout tooltips" ),
+    TransPainterDetIndicator( "TransPainterDetIndicator", "Draw indicator for det" ),
```
bmorrise: This is unnecessary, we should just use the TransPainterStep entry instead.

engine/src/org/pentaho/di/trans/TransPainter.java
```diff
@@ -973,6 +979,13 @@ private void drawStep( StepMeta stepMeta ) {
       LogChannel.GENERAL.logError( "Error calling extension point(s) for the transformation painter step", e );
     }
 
+    try {
```
bmorrise: This is unnecessary, we should just use the TransPainterStep entry instead.

PR #3621, opened 2017-03-06
[BACKLOG-14619] Fix help URLs

AlexanderBuloichik:


mdamour1976:
@AlexanderBuloichik for all help URL's that point to a "versioned" mindtouch link, you can just use the portion of the URL that identifies it, kettle will add the base url to that when we call getDocUrl.  The property that method uses needs to be updated to 7.1 if it hasn't already.

AlexanderBuloichik:
@mdamour1976 okay, will try

AlexanderBuloichik:
@mdamour1976 I added ability to use doc url without prefix in the xml files, and replaced "https://help.pentaho.com/Documentation/7.0/" prefix into "https://help.pentaho.com/Documentation/7.1/"

PR #3634, opened 2017-03-09
[BACKLOG-11076] alow closing tabs when reppository connection lost with unsaved changes

tgf:


tgf:
@bmorrise please review

PR #3637, opened 2017-03-10
[BACKLOG-15186]: PMR and Hadoop Job Executor job entries fail because of they can not find "LogWriter" class (NoClassDefFoundError)

TatsianaKasiankova:
Hadoop Job Executor job entry uses org.pentaho.di.core.logging.LogWriter. This is located in kettle5-log4j-plugin-xxx.jar.
This jar was pulled into <PENTAHO>\data-integration\lib transitively from pentaho-reporting-engine-classic-extensions-kettle project.
After GAV changes for reporting projects (https://github.com/pentaho/pentaho-kettle/commit/8136fd585deb77007d06a89a0df369cae83285c1) this jar stopped to be pulling out transitively and disappeared from <PENTAHO>\data-integration\lib.

But we need to have it for big-data-project.
So the fix is to receive this dependency directly.

### DIFF
--- a/assembly/ivy.xml
+++ b/assembly/ivy.xml
@@ -11,11 +11,11 @@
     <conf name="dataservice-jdbc" />
     <conf name="pentaho-system" />
   </configurations>
-  
+
   <publications>
     <artifact name="${ivy.artifact.id}" type="zip"/>
   </publications>
-  
+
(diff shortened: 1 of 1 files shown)


TatsianaKasiankova:
@duarteteixeira , @mbatchelor Hi, please review.

PR #3642, opened 2017-03-14
[BACKLOG-14959] extra column popping in single threader params

tgf:


tgf:
@bmorrise please review

PR #3647, opened 2017-03-15
[BACKLOG-15326] - Unable to save output to local drive - add 301 HTTP code handling

AndreyBurikhin:
@mdamour1976 , @bmorrise  please review. 
Please, read BACKLOG-15326-related email before merge!


AndreyBurikhin:
Should be treated as a separate case.

PR #3657, opened 2017-03-16
[SP-3379][PDI-14500]-Null if step does not use format when converting dates

denisprotopopov:
- now it uses the format (conversion mask) from the meta data (previous step)
 - added unit test


denisprotopopov:

PR #3659, opened 2017-03-16
[PDI-15991] Spoon very slow to start running transforms with lots of steps

kcaswick:
Improves performance by eliminating excessive clearing of the step fields cache


mbatchelor:
Thanks for the contribution. The fixes were integrated into other fixes that are committed to master. Thanks again!

PR #3669, opened 2017-03-20
[SP-3334][PDI-15989]   Problem saving "Streaming source step" and "St…

tim-ryzhov:
…reaming target step" properties on "ETL Metadata Injection" step (7.0 Suite)

* Saving streamSourceStepname from UI


mchen-len-son:
Double-checked locally.
Looks OK.
Merging.

PR #3673, opened 2017-03-20
[BACKLOG-15349] dont try to use standalone sessions for repository connection

tgf:


tgf:
@bmorrise please review

PR #3675, opened 2017-03-21
[PDI-15801]  [Check if column exists in a table] Multiple errors in s…

tim-ryzhov:
…tep result and logs

* Fix step execution logic
* Unit tests


tim-ryzhov:
@mbatchelor, according to lines 228-229,  initially there is an error and result = false (just not to put it everywhere). Reasons can be different: an empty table name, no columns, bad connection, table not exists, etc. The result is only true if all columns exist (line 298). So there is nothing should be added at line 279

mchen-len-son:
@mbatchelor @bmorrise @e-cuellar can you please take a look?
thanks.

PR #3678, opened 2017-03-21
[PDI-15932] Use PluginRegistry.java to synchronise callExtensionPoint

befc:
Isolated callExtensionPoint for cycle through ExtensionPointMap's HashBasedTable to make use of lock already inside PluginRegistry, since it's already locking access to the ExtensionPointMap

### DIFF
--- a/core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
+++ b/core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2013 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2017 by Pentaho : http://www.pentaho.com
  *
  *******************************************************************************
  *
@@ -90,7 +90,10 @@ public class ExtensionPointMap {
 
   /**
    * Add the extension point plugin to the map
-   * 
+   *
+   * Warning: Do not call this method directly through ExtensionPointMap, it may lead to concurrency problems. Call
+   * PluginRegistry.getInstance().registerPlugin() instead.
+   *
    * @param extensionPointPlugin
    */
   public void addExtensionPoint( PluginInterface extensionPointPlugin ) {
@@ -101,6 +104,9 @@ public class ExtensionPointMap {
 
   /**
    * Remove the extension point plugin from the map
+   *
+   * Warning: Do not call this method directly through ExtensionPointMap, it may lead to concurrency problems. Call
+   * PluginRegistry.getInstance().registerPlugin() instead.
    * 
    * @param extensionPointPlugin
    */
--- a/core/src/org/pentaho/di/core/extension/ExtensionPointHandler.java
+++ b/core/src/org/pentaho/di/core/extension/ExtensionPointHandler.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2013 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2017 by Pentaho : http://www.pentaho.com
  *
  *******************************************************************************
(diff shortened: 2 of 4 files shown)


befc:
Need to increase timeout of the concurrency test, 5 seconds not enough in the wingman build

core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
```diff
@@ -90,7 +90,10 @@ public static ExtensionPointMap getInstance() {
...
   /**
    * Add the extension point plugin to the map
-   * 
+   *
+   * Warning: Do not call this method directly through ExtensionPointMap, it may lead to concurrency problems. Call
+   * PluginRegistry.getInstance().registerPlugin() instead.
+   *
    * @param extensionPointPlugin
    */
   public void addExtensionPoint( PluginInterface extensionPointPlugin ) {
```
pamval: Shouldn't we change the visibility of this method to make it visible inside the package only? 

Or if the class is not to be used from outside, maybe it should be a private class in PluginRegistry?

core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
```diff
@@ -101,6 +104,9 @@ public void addExtensionPoint( PluginInterface extensionPointPlugin ) {
 
   /**
    * Remove the extension point plugin from the map
+   *
+   * Warning: Do not call this method directly through ExtensionPointMap, it may lead to concurrency problems. Call
+   * PluginRegistry.getInstance().registerPlugin() instead.
    * 
    * @param extensionPointPlugin
    */
```
pamval: Same here.

core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
pedrofvteixeira: Ideally, yes.

However, this is public-facing code, and if we were to do such a visibility change on a minor release, we might:

- cause compiler errors on any customer/community custom code out there that may be accessing it ( i.e. custom Kettle plugins )
- cause the same issue to occur on previous versions, should this fix ever be included in a SP case

Safest approach seems to be to mark this as deprecated and leave a comment with a recommended alternative.

If you still feel otherwise, please let us know and we'll do the change.

befc:
@mbatchelor, could you please have a look

core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
pamval: I don't see how this change addresses the issue. Anyone caling this method will potentially hit the concurrency issue. Access to the map is still not synchronized.

befc: @pamval, this PR approach is based on the fact that PluginRegistry, in registerPlugin() and removePlugin() methods already protects the ExtensionPoint map inside it's own lock. So by adding the callExtensionPoint() in PluginRegistry, we can use the same lock and protect this access.

Yes, HashBasedTable can be accessed by someone else and we will have the same problem. Other approach that can fix this access, without going back to the solution in the first PR, is to move the pieces of code in PluginRegistry that call ExtensionPointMap from inside it's own lock, to avoid dead lock problems. And guarantee all accesses to ExtensionPointMap are protected from concurrent access by using "synchronized" or "reentrantreadwritelock".

befc: But, even with this solution, if someone calls ExtensionPointMap.get() or ExtensionPointMap.getMap() and iterates over them, they will end up with the ConcurrentModificationException

befc:
@mbatchelor, @pamval, @pedrofvteixeira, Will close this pull request and create a new one with this solution. 

Side note: To completely protect the ExtensionPointMap need to remove the getMap() method as well.

PR #3681, opened 2017-03-21
[BACKLOG-14865] As a PDI user, I want the rows streamed back from spa…

andrei-viaryshka:
…rk to be converted into a form PDI can use (back to RowMeta+Object[])

Move RowConversionManager to separate plugin.

Related to https://github.com/pentaho/pentaho-ee/pull/97.


andrei-viaryshka:
@mkambol Could you review please?

andrei-viaryshka:
The test fails because it uses pdi-engine-api SNAPSHOT which does not contains required changes.

PR #3696, opened 2017-03-24
[PDI-15898]   Salesforce Input Step "Get Fields" button returns wrong…

tim-ryzhov:
… field names when using custom SOQL query
* Do not duplicate fields 
* Fixed to the same behavior as Axis - Salesforce SOAP Api sends IDs always in the response, even if we don't request it in SOQL query and the id's value is null in this case. So, do not add this Id to the fields list
* Unit tests


mchen-len-son:
@mbatchelor @bmorrise @e-cuellar  can you please take a look?
thanks.

PR #3697, opened 2017-03-27
[BACKLOG-14878] Adding Logging Event, Serializer and hooks into the A…

pentaho-nbaker:
…dapter


pentaho-nbaker:
Wingman won't be able to build this. There are changes to pdi-engine-api which are needed in Engine.

PR #3699, opened 2017-03-28
[PDI-15932] Protect extensionPointPluginMap against ConcurrentModificationException...

befc:
… with reentrant lock and isolate the ExtensionPointHandler iteration inside ExtensionPointMap.callExtensionPoint()

Based on the solution proposed by @mbatchelor:

- [x] Remove the get( id ) method from the map: 
\- Having grepped all the code we ship, the only classes that use ExtensionPointMap are in the same package.
\- It's not a public API nor does it implement an interface
> Removed getMap() as well to prevent any iteration of the map from the outside that may lead to a similar problem

- [x] Create a new package-protected method called callExtensionPoint that receives a log channel, id, and object (like the method in ExtensionPointHandler only not public static)
- [x] In ExtensionPointMap do the iteration

- [x] Fix the couple of tests that use the map directly (if there are any)
> Had to create two new methods that expose just the needed values for validation in UTs: getTableValue() and getNumberOfRows()

**Extra steps:** 
- In PluginRegistry class I had to remove the blocks that make calls to ExtentionPointMap via PluginTypeListener from the blocks that are locked. The reason for this was the possibility of having dangerous concurrency issues, like dead locks.
- Added new UT, testExtensionPointMapConcurrency(), to check concurrency access in ExtensionPointMap

### DIFF
--- a/core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
+++ b/core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2013 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2017 by Pentaho : http://www.pentaho.com
  *
  *******************************************************************************
  *
@@ -24,8 +24,8 @@ package org.pentaho.di.core.extension;
 import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.collect.HashBasedTable;
-import com.google.common.collect.Maps;
 import com.google.common.collect.Table;
+import org.pentaho.di.core.exception.KettleException;
 import org.pentaho.di.core.logging.LogChannel;
 import org.pentaho.di.core.logging.LogChannelInterface;
 import org.pentaho.di.core.plugins.PluginInterface;
@@ -33,7 +33,7 @@ import org.pentaho.di.core.plugins.PluginRegistry;
 import org.pentaho.di.core.plugins.PluginTypeListener;
 
 import java.util.List;
-import java.util.Map;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 /**
  * This class maintains a map of ExtensionPointInterface object to its name.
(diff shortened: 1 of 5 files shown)


befc:
@mbatchelor, @pamval, @pedrofvteixeira, here is the new PR based on the comments from the last one. Could you please check.

**Note**: I checked about coverage issue but seems there is not much that can be done. 
I could call reInitialize() in the concurrency test to increase coverage in ExtensionPointMap, or add a new test for this method, since it's the only one that is not covered but seems like this method is not called anywhere, maybe could even be removed?

core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
```diff
@@ -105,38 +103,91 @@ public void addExtensionPoint( PluginInterface extensionPointPlugin ) {
...
+   * Call the extension point(s) corresponding to the given id
+   *
+   * This iteration was isolated here to protect against ConcurrentModificationException using PluginRegistry's lock
+   *
+   * @param log     log channel to pass to extension point call
+   * @param id      the id of the extension point interface
+   * @param object  object to pass to extension point call
+   */
+  public void callExtensionPoint( LogChannelInterface log, String id, Object object ) throws KettleException {
+    lock.writeLock().lock();
```
pamval: Since you're not changing the map here, shouldn't this get the read lock instead of the write lock?

befc: Yes, before, as I was using the get( id ) method that was now deleted, the supplierFunction used by this method had some operations inside of the google guava code that had problems when running them concurrently. But as you mentioned this now, I tried to replace with a readLock and it behaves well with the new code. 

Thank you, will pull this changes into the PR

core/src/org/pentaho/di/core/extension/ExtensionPointMap.java
```diff
@@ -105,38 +103,91 @@ public void addExtensionPoint( PluginInterface extensionPointPlugin ) {
...
+   * Call the extension point(s) corresponding to the given id
+   *
+   * This iteration was isolated here to protect against ConcurrentModificationException using PluginRegistry's lock
+   *
+   * @param log     log channel to pass to extension point call
+   * @param id      the id of the extension point interface
+   * @param object  object to pass to extension point call
+   */
+  public void callExtensionPoint( LogChannelInterface log, String id, Object object ) throws KettleException {
+    lock.readLock().lock();
```
befc: Changed from writeLock to readLock, according to @pamval's comment

PR #3701, opened 2017-03-28
[BACKLOG-11076] reset connection on repo loss detection

tgf:


tgf:
@bmorrise please review

PR #3707, opened 2017-03-30
[PDI-15059] Spoon startup: Multiple and different slf4j-log4j jar fil…

mchen-len-son:
…es in the class path

 - downgraded slf4j to 1.7.7 to be consistent across all the Pentaho modules


mchen-len-son:
@pamval @mbatchelor @duarteteixeira please take a look

PR #3712, opened 2017-03-30
[SP-3407][PDI-15871]-Repository Explorer: Database Connections tab sh…

mchen-len-son:
…ows "Unable to initialize security tab" dialog

 - added verification modifiedDate != null
 - added unit test


mchen-len-son:
This are exactly the same changes as in #3710.
The only difference is in commit message - it contains the right JIRA number.

PR #3716, opened 2017-04-03
[PDI-16048] LDAP Input protocol not substituted

AliaksandrShuhayeu:
-added resolving a variable for a protocol of LDAP step
-added a test


AliaksandrShuhayeu:

PR #3726, opened 2017-04-05
[BACKLOG-15249] Exception handling modifications so Spark App messages get sent to Spoon

ccaspanello:
@hudak / @CodeOnCoffee / @mkambol / @rfellows


pdi-engine/api/src/main/java/org/pentaho/di/engine/api/remote/Execution.java
```diff
@@ -27,6 +27,7 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Serializable;
+import java.util.concurrent.*;
```
hudak: Is the correct ExecutonException being used?

PR #3732, opened 2017-04-07
[BACKLOG-15646] AEL: when running transformation with the same name a…

andrei-viaryshka:
…s step in it , ktr hangs & fails

Change Transformation Id to random UUID.

### DIFF
--- a/engine/src/org/pentaho/di/trans/ael/adapters/TransMetaConverter.java
+++ b/engine/src/org/pentaho/di/trans/ael/adapters/TransMetaConverter.java
@@ -26,6 +26,7 @@ package org.pentaho.di.trans.ael.adapters;
 
 import com.google.common.base.Throwables;
 import org.pentaho.di.core.exception.KettleException;
+import org.pentaho.di.core.util.Utils;
 import org.pentaho.di.engine.api.model.Operation;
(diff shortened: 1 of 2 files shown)


andrei-viaryshka:
@mkambol @hudak Could you review please?

engine/src/org/pentaho/di/trans/ael/adapters/TransMetaConverter.java
```diff
@@ -32,18 +32,20 @@
...
 
   public static final String TRANS_META_CONF_KEY = "TransMeta";
+  public static final String TRANS_META_NAME_CONF_KEY = "TransMetaName";
   public static final String STEP_META_CONF_KEY = "StepMeta";
 
 
   public static Transformation convert( TransMeta transMeta ) {
     final org.pentaho.di.engine.model.Transformation transformation =
-      new org.pentaho.di.engine.model.Transformation( transMeta.getName() );
+      new org.pentaho.di.engine.model.Transformation( UUID.randomUUID().toString() );
```
mkambol: Thinking we should probably keep transMeta.getName(), but append a string to make it unique.  That way logging can provide some idea of what the source transformation is.

PR #3733, opened 2017-04-07
[PPP-3692] - Use of vulnerable component commons-beanutils: 1.8.0 - C…

YuryBY:
…VE-2014-0114

- remove redundant dependency commons-collections-core

@pamval, @mbatchelor,  @graimundo   could you please take a look?

engine receives commons-beanutils and commons-collections transitively

### DIFF
--- a/engine/ivy.xml
+++ b/engine/ivy.xml
@@ -7,6 +7,7 @@
         <conf name="runtime" />
         <conf name="test" extends="runtime" visibility="private" />
         <conf name="source" />
+        <conf name="provided" extends="runtime" visibility="private" />
   </configurations>
 
   <publications>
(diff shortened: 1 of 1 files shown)


hbfernandes:
As far as i know, the way to get a "provided like" behaviour is to use a configuration with a private visibility. Same as it's already defined on that ivy.xml for the test dependencies. So the quick fix is to either set the dependency configuration to **conf="test->default"** or to add a new configuration to the file in all similar to the test one and call it "provided".

**<conf name="provided" extends="runtime" visibility="private" />**

and then use

**<dependency org="commons-beanutils"   name="commons-beanutils-core" rev="1.8.0"  conf="provided->default"  transitive="false" />**

The later option is the most expressive of intent, so I would advise that one.

YuryBY:
@pamval, @mbatchelor, @graimundo the PR is updated. Thanks.

PR #3736, opened 2017-04-10
[PDI-14598] When creating roles with the existing names but with lowercase letters the roles are added but they cannot be removed

Advent51:
- fixed role creation - now users cannot create roles with the same name but with different case.


Advent51:
**Please don't merge before my request**

Advent51:
@mbatchelor please review

Advent51:

PR #3737, opened 2017-04-10
[BACKLOG-15159] - DET - If a user deletes a field after having run a transformation and then inspects the data in a DET flat table viz an error will occur - CSV file input fix

AndreyBurikhin:
@bmorrise , @mdamour1976 please review


AndreyBurikhin:
Closed as we do not want to bring this to 7.1. This PR should be reopened after creating the 7.1 brunch.

PR #3739, opened 2017-04-11
[BACKLOG-15706]-HDP 25 - Hive with PMR job fails-Cannot instantiate class: org.osjava.sj.SimpleContextFactory

stepanovdg:
* introduce new lettle compatibility property


rmansoor:
@stepanovdg PR looks good. We have check style violations. I will issue another PR to fix the checkstyle issues.

PR #3740, opened 2017-04-11
[BACKLOG-15879] Upgrading POI to 3.15 due to vulnerabilities

e-cuellar:


hbfernandes:
Are we sure commons collections is not needed for anything else other than poi? Because version 3 is not compatible with 4, so we'd break any usage outside of poi. Also, on the engine ivy.xml, seems like the poi-ooxml dependency is duplicated, i believe we could delete one.

PR #3751, opened 2017-04-12
[BACKLOG-15748] Adding pentaho-kerberos-jaas capability only if

pdesai16:
ael-security marker feature is installed.


pdesai16:
@CodeOnCoffee @bmorrise @pentaho/rogueone  - Please review

pdesai16:
There will be separate PR to add the ael-security karaf feature

PR #3764, opened 2017-04-14
[BACKLOG-15685] As a Chris, I want to be able to disable steps when r…

pavel-sakun:
…unning on Spark

Added case when there are both types of incoming hops: enabled and
disabled.

@mkambol


mkambol:
Another thing I noticed- If an input step has a disabled output, it will still be present in the trans and execute.  Those should also be removed from the generated transformation.

PR #3775, opened 2017-04-18
[SP-3390][BACKLOG-12164] Change default 'ignore missing path' value

AliaksandrShuhayeu:


AliaksandrShuhayeu:
It contains trivial changes and can not call a maven problem.

PR #3786, opened 2017-04-20
[PDI-13208] - Fixed input sample is broken

mbatchelor:


mbatchelor:
Ignore the diff - the sample itself is broken. The dialog says to omit the linefeed in the length, but the length given (and the sum of the field lengths) is 64 and the line is clearly only 63. Changed the line length to 63 and the length of the last field to 37 which is correct. Sample works with the input file having \n or \r\n line endings.

mbatchelor:
@pedrofvteixeira - could you review this ?

PR #3787, opened 2017-04-20
[SP-3441][PDI-15932] Protect extensionPointPluginMap against Concurre…

mchen-len-son:
…ntModificationException with reentrant lock and isolate the ExtensionPointHandler iteration inside ExtensionPointMap.callExtensionPoint()


mchen-len-son:

this is a backport of #3699.

PR #3788, opened 2017-04-20
[PDI-14732] "The PostgreSQL bulk loader" can't inserts datetime field only date, that makes unavailable insert large datasets with datetime fields into PostgresSQL

Advent51:
- Proposed fix was implemented.


Advent51:

Advent51:
@mbatchelor @mchen-len-son @pamval @duarteteixeira please review

PR #3798, opened 2017-04-24
[PDI-15835] - VFS DefaultFileSystemManager is not threadsafe - a Thre…

alina-ipatina:
…adLocal should be used to protect the FileSystemManager

- add threadsafe wrapper for StandartFileSystemManager
- add concurrency tests


alina-ipatina:
@mchen-len-son @mbatchelor @pamval please review.

alina-ipatina:
@mchen-len-son @mbatchelor @pamval could you please take a look?

pamval:
Why did we choose to subclass FileSystemManager instead of protecting access to it using ThreadLocal, as suggested in the Jira?

Also, how does this address Amazon VFS registration (the other issue mentioned in the jira)?

PR #3799, opened 2017-04-24
[BACKLOG-16118] Fixed issue with field lengths in text input

bmorrise:


bmorrise:
@pentaho/r2d2

dkincade:
A unit or integration test on this would have been nice.

PR #3816, opened 2017-04-25
[BACKLOG-13032] Mavenize elasticsearch-bulk-insert

e-cuellar:


smaring:
dependency.pentaho-xul.version  does not exist in https://github.com/pentaho/build-resources/blob/8.0-qat/version.properties  also, scm is probably not required at this level

PR #3826, opened 2017-04-27
[BACKLOG-16179] Fix String to Numeric conversion behaviour that was…

davidmsantos90:
…causing it to throw a KettleException

@pentaho/millenniumfalcon please review

Converting from string to a numeric must rely on the Value metadata and not on the default mask of the numeric type we are converting to.

Also did a small tweak in the sample so that it gives the exact result after the conversion (fix works without this)

### DIFF
--- a/assembly/package-res/samples/transformations/JavaScript - String to Number to String conversion.ktr
+++ b/assembly/package-res/samples/transformations/JavaScript - String to Number to String conversion.ktr
@@ -471,10 +471,10 @@
         <type>Number</type>
         <format/>
         <currency/>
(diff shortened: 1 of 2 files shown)


graimundo:
@davidmsantos90 don't forget the cherry pick PRs for 7.1 and 7.1.0.0

davidmsantos90:
@graimundo cherry-picks:
 - https://github.com/pentaho/pentaho-kettle/pull/3832
 - https://github.com/pentaho/pentaho-kettle/pull/3833

PR #3827, opened 2017-04-27
[PDI-16178] arrayIndexOutOfBoundsException when collapsing a folder in "Open File" dialog

Advent51:
- check was added


Advent51:

PR #3828, opened 2017-04-27
[PDI-16192] handle lost repository on schedule perspective and explorer

tgf:


tgf:
@bmorrise please review

PR #3834, opened 2017-04-28
[PDI-14781] Empty log file gets created when job runs on carte server

alina-ipatina:
- write logs to file from carte server


alina-ipatina:
@mchen-len-son @mbatchelor @pamval please review.

PR #3848, opened 2017-05-02
[SP-3448][PDI-16045]-Rest Client step does not encode parameter list properly

denisprotopopov:
- fix for case when parameters are defined in the "Matrix Parameters" tab


mchen-len-son:
Double-checked locally - the tests are passing.

PR #3857, opened 2017-05-05
[PDI-7090] Only one user can make changes in a repository at any time. (Database Repository is locked)

j-a-n:
When using MySQL as repository database every SELECT statement has to be followed by a COMMIT to avoid locking issues in parallel sessions. For this reason the current code contains the method closeReadTransaction() which also needs to be called after SELECTs in getNextTableID(), getIDs() and getStrings().

The patch fixes the issue described in  [[PDI-7090]](http://jira.pentaho.com/browse/PDI-7090).


mbatchelor:
Thanks for the contribution! PR was reviewed and merged.

PR #3860, opened 2017-05-08
[PDI-16196] - Fix inconsistencies when closing tabs with the same name

joaormf:


pamval:
@joaormf please submit a cleanup pr fixing that checkstyle violation.

I scrolled right past it when accepting the pr....

joaormf:
@pamval Will fix those checkstyle and the cobertura issues in a new PR

PR #3869, opened 2017-05-09
[BACKLOG-16377] AEL in client assembly revision

lgrill-pentaho:
* Cleanup the leftover pdi-daemon-assembly-<version> directory
* Set exec flag perms on AEL binary files
* Bit of additional cleanup for kettle assembly


rfellows:
Looks good. be sure to keep track of your commits so we can cherry-pick them over to the 7.1 branches once we verify them in 8.0

PR #3880, opened 2017-05-11
[BACKLOG-16413] - AEL is generally missing from pdi-ce

rfellows:


rfellows:
@dkincade @mdamour1976 please review

PR #3887, opened 2017-05-12
[BACKLOG-16444] - Removing the transitive dependencies.

rmansoor:


rmansoor:
@hudak , @pentaho-nbaker please review

hudak:
We may need to add rxjava as an assembly
``` 
   <dependency>
      <groupId>io.reactivex.rxjava2</groupId>
      <artifactId>rxjava</artifactId>
      <version>${rxjava.version}</version>
    </dependency>
```

PR #3894, opened 2017-05-15
[PDI-15982] Incomplete/Missing logs when Main Job with sub jobs are s…

alina-ipatina:
…cheduled

- add internal buffer for LogChannelFileWriter
- do not remove channels from LoggingRegistry that are write logs to files


alina-ipatina:
@mchen-len-son @mbatchelor @e-cuellar please review.

PR #3895, opened 2017-05-15
[BACKLOG-16013] Set internal Kettle variables

andrei-viaryshka:


andrei-viaryshka:
@hudak @mkambol Could you review please?

PR #3898, opened 2017-05-15
[COMP-5831] Exclude drools-reteoo

lucboudreau:
Exclude drools-reteoo. This is a legacy engine which is not used by default. There is no need for us to include it.


lucboudreau:
@pentaho-nbaker @CodeOnCoffee Hi Nick. For review. Looks to me like we don't need this rules engine. Will close the component case once it goes away.

PR #3903, opened 2017-05-16
[PDI-14906]-Check Db connections step timeout minutes calculated wrong

denisprotopopov:
- refactored unit test to make execution shorter


mchen-len-son:
Apparently, Wingman didn't catch the last changes - checkstyle is OK.
Merging.

PR #3921, opened 2017-05-19
[SP-3526][PDI-14781] Empty log file gets created when job runs on car…

alina-ipatina:
…te server

- write logs to file from carte server


lijingjian001:
image

Target "unzip-swtlibs" does not exist in the project "Kettle CE Assembly".

who can help me?

PR #3926, opened 2017-05-23
[PDI-14734]-Dialog's table discards input after losing focus

denisprotopopov:
- fixed saving dialog
 - added test for save/load


denisprotopopov:
@mbathchelor please review the changes

PR #3927, opened 2017-05-23
[SP-3562][PDI-16305] - Dialog uses DI Repository instead of Pentaho R…

mchen-len-son:
…epository


mchen-len-son:
@mbatchelor please take a look. 
the failing test is not related to the changes.

mchen-len-son:
@pamval can you please take a look?
thanks.

PR #3936, opened 2017-05-23
[PDI-16352] Merge Rows Diff check row layout

matthewtckr:


matthewtckr:
Will rebase for http://jira.pentaho.com/browse/PDI-16352

engine/src/org/pentaho/di/trans/steps/mergerows/MergeRowsMeta.java
```diff
@@ -334,6 +335,35 @@ public void check( List<CheckResultInterface> remarks, TransMeta transMeta, Step
...
       remarks.add( cr );
     }
+
+    RowMetaInterface referenceRowMeta = null;
+    RowMetaInterface compareRowMeta = null;
+    try {
+      referenceRowMeta = transMeta.getPrevStepFields( referenceStream.getStepname() );
+      compareRowMeta = transMeta.getPrevStepFields( compareStream.getStepname() );
+    } catch ( KettleStepException kse ) {
+      // investigate causes
```
mbatchelor: Swallowed exception here - in the validation, shouldn't that exception be captured and provided as a check result?

engine/src/org/pentaho/di/trans/steps/mergerows/MergeRows.java
```diff
@@ -257,7 +257,7 @@ public boolean init( StepMetaInterface smi, StepDataInterface sdi ) {
    * @throws KettleRowException
    *           in case there is a compatibility error.
    */
-  protected void checkInputLayoutValid( RowMetaInterface referenceRowMeta, RowMetaInterface compareRowMeta ) throws KettleRowException {
+  static void checkInputLayoutValid( RowMetaInterface referenceRowMeta, RowMetaInterface compareRowMeta ) throws KettleRowException {
```
mbatchelor: Why change this to a static? It would be better if you just used "this" in the check method and change the method back to protected.

matthewtckr: The `protected` was removed so that it can be referenced in the Meta class, and it was made `static` as the method does not make use of any state of the class, only the two arguments, so it didn't need to run in the context of the instance object.

engine/src/org/pentaho/di/trans/steps/mergerows/MergeRowsMeta.java
matthewtckr: Thanks, I'll look into this.

matthewtckr: The PR has been updated to add an error check result when it catches here.

PR #3938, opened 2017-05-24
[SP-3491]-[PDI-13182] - Fix invalid message

PashkovichAndrei:


mchen-len-son:
Minor change. Checkstyle is OK.

PR #3939, opened 2017-05-24
[SP-3559][PDI-16196]-Saving a transformation with a duplicate name makes PDI work inconsistently

denisprotopopov:
- Fix inconsistencies when closing tabs with the same name
 - Improve cobertura tests coverage


mchen-len-son:
Checked locally.
Tests passed and checkstyle is OK.
Merging.

PR #3943, opened 2017-05-24
[SP-3551][PDI-16281] - Catch exception during encrypt

stepanovdg:


stepanovdg:
@mbatchelor @pamval @mchen-len-son

PR #3952, opened 2017-05-26
[SP-3565]Commit open transactions

AliaksandrShuhayeu:
Commit open transactions after select statement to avoid locking of parallel sessions when using MySQL as repository database.


AliaksandrShuhayeu:

PR #3962, opened 2017-05-30
[PDI-16222] Unzip File Job Entry - Move files After Extraction Failed

AliaksandrShuhayeu:
[PDI-16222] Unzip File Job Entry - Move files After Extraction Failed
-moving and deleting correct fileObject

### DIFF
--- a/engine/src/org/pentaho/di/job/entries/unzip/JobEntryUnZip.java
+++ b/engine/src/org/pentaho/di/job/entries/unzip/JobEntryUnZip.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2016 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2017 by Pentaho : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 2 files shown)


engine/src/org/pentaho/di/job/entries/unzip/JobEntryUnZip.java
```diff
@@ -789,45 +789,7 @@ public boolean includeFile( FileSelectInfo info ) {
...
-        } finally {
-          if ( destFile != null ) {
-            try {
-              destFile.close();
-            } catch ( IOException ex ) { /* Ignore */
-            }
-          }
-        }
-      }
+      doUnzipPostprocessing( sourceFileObject, movetodir, realMovetodirectory );
```
mbatchelor: Would prefer the if ( afterunzip == 1 ) { to be around the call to doUnzipPostprocessing so that an unnecessary call is avoided.

engine/src/org/pentaho/di/job/entries/unzip/JobEntryUnZip.java
```diff
@@ -839,6 +801,51 @@ public boolean includeFile( FileSelectInfo info ) {
     return retval;
   }
 
+  /**
+   * Moving or deleting source file if it is enabled.
+   */
+  private void doUnzipPostprocessing( FileObject sourceFileObject, FileObject movetodir, String realMovetodirectory ) throws FileSystemException {
```
mbatchelor: Method should be called doUnzipPostProcessing ( capitalize Processing please ).

AliaksandrShuhayeu:
The reason of this issue was using fileObject instead of sourceFileObject only. They both are input arguments of the method unzipFile(...). They are identical if it was called for a file, but for a folder they are differents (see call hierarchy).
I could only changed it but I got problems with testing it. The method unzipFile(...) contains calling static (KettleVFS.getFileObject(...), KettleVFS.getInstance().getFileSystemManager() ). So I created new method only for testing it. I did't do aditional changing (for example: calling this method for afterunzip == 0) that my changing was understandable.

I fixed code review issues. @mbatchelor Could you please review and merge it?

AliaksandrShuhayeu:
I fixed code review issues. @mbatchelor Could you please review and merge it?

PR #3970, opened 2017-05-31
[PDI-14764] Jaro algorithm glitch

stanislau-strelchanka:
- exclude log4j dependency pull


stanislau-strelchanka:
@lucboudreau @mchen-len-son  can you please take a look?

PR #3971, opened 2017-05-31
[PDI-16224] Job/Transformation returns "ORA-00911: invalid character"error

AliaksandrShuhayeu:
-added correct processing backslash for Oracle DB
refactored previous fix


AliaksandrShuhayeu:
-added correct processing backslash for Oracle DB
refactored previous fix
@lucboudreau @mbatchelor Could you please review and merge it?

lucboudreau:
Great job on the refactoring. This solution is much better.

PR #3972, opened 2017-05-31
[BACKLOG-16115] Add logic to stop transformation

ccaspanello:
@hudak / @mkambol


ccaspanello:
Ran command on my machine.  Frist time it failed; but then I manually put the Maven dependency into the ivy cache folder and ran the build again; worked fine.  This is obviously a build/dependency issue.

PR #3984, opened 2017-06-06
[PSW-229] Script files don't have an header with Copyright data

Advent51:
- license header was added


Advent51:

PR #3991, opened 2017-06-07
[CLEANUP] fix GAVs for mavenized plugins

smaring:
@pminutillo @hbfernandes


hbfernandes:
@smaring I rebased the base branch so this PR has all sorts of conflicts now, i'll update the gav's myself so no need for this anyway.

PR #3993, opened 2017-06-07
[BACKLOG-16769] - removing newline before any cdata.  removing compil…

kurtwalker:
…e dependency on wstx.


kurtwalker:
don't know why it's complaining about copyright.  They have 2017

PR #4032, opened 2017-06-16
[PDI-16312] JSON Input step ignores the trim type

Advent51:
- trim type now sets.
- line separators in test were changed to the system default.


Advent51:

Advent51:
@mbatchelor @pamval @mchen-len-son @duarteteixeira please review

PR #4034, opened 2017-06-16
[PDI-16275] Output File incorrect after meta-injection due to derived value not being updated

Advent51:
- the newline is now updates in the file format setter
- the "injection" annotation was moved on the setter in order to implement the above thing in metadata injection


Advent51:

Advent51:
@mbatchelor @pamval @mchen-len-son @duarteteixeira please review

PR #4065, opened 2017-06-22
[SP-3604][PDI-16198] Multiple Preview dialogs when previewing a step …

mchen-len-son:
…using "Number of Copies" > 1

In rare cases in rare environments sometimes buffer can contain additional null row value.
It was not possible reproduce it in debugging with breakpoints on modification points of buffer.
Fix NPE. See additional PDI-16198.


mchen-len-son:
@mbatchelor @pamval @stepanovdg please take a look.

PR #4070, opened 2017-06-23
Base classes for input steps

AlexanderBuloichik:


AlexanderBuloichik:
Will be changed for mavenization

PR #4074, opened 2017-06-23
Revert "[SP-3564]Commit open transactions after select statement to avoid locking of parallel sessions when using MySQL as repository database."

stepanovdg:
This reverts commit 25b5b63bf472151d73ab126a581e35783bc27b60.


stepanovdg:
@mchen-len-son @duarteteixeira Could you please take a look?

PR #4079, opened 2017-06-24
[PDI-16388] - SelectValues type change not using default conversion mask

mbatchelor:


mbatchelor:
@lucboudreau , @pedrofvteixeira  , @pamval , @mchen-len-son - can any of you review this PR please? Note that it should increase code coverage as I'm testing things that haven't been tested before. Additionally, without setting the new compatibility flag, the system will work exactly as it did before this change. You only get the new behavior if you set the KETTLE_COMPATIBILITY_SELECT_VALUES_TYPE_CHANGE_USES_TYPE_DEFAULTS flag (either as a variable, or in the kettle.properties) to Y.

engine/src/org/pentaho/di/trans/steps/selectvalues/SelectValuesMeta.java
```diff
@@ -384,13 +384,22 @@ public void getDeleteFields( RowMetaInterface inputRowMeta ) throws KettleStepEx
   }
 
   public void getMetadataFields( RowMetaInterface inputRowMeta, String name ) throws KettlePluginException {
```
lucboudreau: Should probably be annotated with @Deprecrated.

mbatchelor:
Incorporated Luc's feedback.

PR #4084, opened 2017-06-27
[SP-3624][PDI-16360] - Fix NPE in KettleGenericFileSystemConfigBuilde…

mchen-len-son:
…r (#4008)


mchen-len-son:
@stepanovdg please review

PR #4085, opened 2017-06-27
[SP-3586] Backport of PDI-15982 - Incomplete/Missing logs when Main Job with sub jobs are scheduled (7.1 Suite)

Advent51:
backport of #3906


Advent51:

PR #4096, opened 2017-06-29
[BACKLOG-17069] Fixes clearing recent searches at startup AND avoids …

e-cuellar:
…displaying recent searches empty dropdown


e-cuellar:
@ddiroma Please Sir!...

PR #4099, opened 2017-06-29
[BACKLOG-17014]  Provide key driven Mulit-provider MetastoreLocator

tkafalas:


rmansoor:
@pentaho/2-1b, @smaring . This is something we are seeing after the mavenization. tests keeps on failing. This has cause our team to reverts other commits. Now, I am going ahead and processing the PR since I know this commit has not caused this test to break. We do need to look at the test taking long time after being switched to maven.

PR #4103, opened 2017-06-30
[UNIT-TEST]-FIX Saxon dependency

stepanovdg:


hbfernandes:
This is not the correct way to fix the problem and it would have implications on the assemblies if the transitive dependencies change, so that would also need updating.

PR #4104, opened 2017-06-30
[BACKLOG-15721] Fixes for the following...

ddiroma:
From: https://docs.google.com/spreadsheets/d/17QboBxK0V064iFzd3GXdpYGZ5I2bg-yWP5OOCYyC9u0/edit#gid=1166424294

Lines: 43/62/82 (same issue), 50, 66, 69, 70, 80 (no fix needed), 83, 84.
43/62/82: Public, Home, don't have proper capitalization in the breadcrumb area.
50: Date/Time text color should be using #999999 in the list view. (card view is correct currently, just not the list view)
66: in the save dialog, clicking a folder name in the right pane is populating the "File name" text box. It should only populate in that box when a file is selected.
69: The breadcrumb truncated menu text color should be #333333.
70: "The breadcrumb truncated menu drop shadow and border doesnt match the flyout we're using in the tab container at the bottom of DET.
-- See this page for styling under ""Tab list"". http://ux.pentaho.com/DET/#g=1&p=tab-container"
80: ALREADY WORKING...DID NOT REPRODUCE ISSUE (using chrome browser)
83: "Per UX request, update the selected state color for the tree.

Main color will remain primary blue in tree.
If an item is selected in the right hand pane, the tree selected folder will be dark blue, while the selected item on right is primary blue."
84: Per UX request, update "File folder" to "Folder"


ddiroma:
@bmorrise 
@e-cuellar 

Removed unused imports from RepositoryDirectory.java. Please review and merge.

PR #4108, opened 2017-07-03
[BACKLOG-16971] Truncate recent search text at the edge of the flyout…

e-cuellar:
… container with an ellipsis and use tooltip.


e-cuellar:
@ddiroma Please Sir!

PR #4110, opened 2017-07-04
[BACKLOG-17053] enable trans graph logging before execution

tgf:


tgf:
@bmorrise please review

PR #4113, opened 2017-07-04
[PDI-16255] Filename Field Name Changes in XML Input Stream (StAX) St…

stanislau-strelchanka:
[PDI-16255] Filename Field Name Changes in XML Input Stream (StAX) Step 7.x when there is an incoming Hop

   - add property for 6.1 compatibility for StAX step

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/Const.java
+++ b/core/src/main/java/org/pentaho/di/core/Const.java
@@ -1062,6 +1062,11 @@ public class Const {
   public static final String KETTLE_GLOBAL_LOG_VARIABLES_CLEAR_ON_EXPORT = "KETTLE_GLOBAL_LOG_VARIABLES_CLEAR_ON_EXPORT";
 
   /**
+   * Set this variable to true to disable receiving file names from another step for XML Input Stream (StAX) step.
+   */
+  public static final String KETTLE_IGNORE_RECEIVING_INPUT = "KETTLE_IGNORE_RECEIVING_INPUT";
+
+  /**
    * Compatibility settings for {@link org.pentaho.di.core.row.ValueDataUtil#hourOfDay(ValueMetaInterface, Object)}.
    *
    * Switches off the fix for calculation of timezone decomposition.
--- a/engine/src/main/resources/kettle-variables.xml
+++ b/engine/src/main/resources/kettle-variables.xml
@@ -457,6 +457,12 @@
   </kettle-variable>
 
   <kettle-variable>
+    <description>Set this variable to true to disable receiving file names from another step for XML Input Stream (StAX) step.</description>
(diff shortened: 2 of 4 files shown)


stanislau-strelchanka:
@mbatchelor @mchen-len-son please review

stanislau-strelchanka:
@mbatchelor @mchen-len-son please review

stanislau-strelchanka:

stanislau-strelchanka:

pamval:
@stanislau-strelchanka why are we introducing a configuration to fix a regression? In which circumstances do we need to keep the behavior as it exists now?

stanislau-strelchanka:
@pamval according to http://jira.pentaho.com/browse/PDI-629 this behavior is not a regression.
It is consistent with similar steps, CSV input for example. But as Chantel Brathwaite explained in the comment section it is not backward compatible with transformations that are coming from 6.1 or earlier versions of kettle.

stanislau-strelchanka:

stanislau-strelchanka:

kolinus:
@pedrofvteixeira  @mdamour1976 could you please take a look

stanislau-strelchanka:
@mchen-len-son @mbatchelor @pamval @pedrofvteixeira @mdamour1976 
please review

stanislau-strelchanka:
@mchen-len-son @mbatchelor @pamval @pedrofvteixeira @mdamour1976
please review

core/src/main/java/org/pentaho/di/core/Const.java
```diff
@@ -1062,6 +1062,11 @@ public String getMessage() {
   public static final String KETTLE_GLOBAL_LOG_VARIABLES_CLEAR_ON_EXPORT = "KETTLE_GLOBAL_LOG_VARIABLES_CLEAR_ON_EXPORT";
 
   /**
+   * Set this variable to true to disable receiving file names from another step for XML Input Stream (StAX) step.
+   */
+  public static final String KETTLE_IGNORE_RECEIVING_INPUT = "KETTLE_IGNORE_RECEIVING_INPUT";
```
mbatchelor: Strelchanka, I don't like the change done this way because it becomes an all or nothing enterprise (you always ignore the receiving input or you don't). Instead, I think I'd rather see the step behave as it did before if there's already a value in the field irrespective of whether there's an incoming hop.

stanislau-strelchanka: @mbatchelor, I've created property to make it backward compatible with this issue #826.
Just to be clear regarding "make it behave as it did before" did you mean totally disable influence of incoming hops to the step( only use entered manually variables ) or give a priority to entered manually variables over incoming hops?

PR #4120, opened 2017-07-06
[PDI-16318] REST Client Step: Http Method DELETE results in 401 where GET, PUT succeed. Postman works fine with OK/200

Advent51:
- if the entity is an empty string, the DefaultApacheHttpMethodExecutor will not set headers due to the null com.sun.jersey.client.property.chunkedEncodingSize property.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/rest/Rest.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/rest/Rest.java
@@ -169,10 +169,10 @@ public class Rest extends BaseStep implements StepInterface {
       }
 
       ClientResponse response = null;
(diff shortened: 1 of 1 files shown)


Advent51:

Advent51:
To see the problem you can look at https://github.com/jersey/jersey-1.x-old/blob/6dd59a568eced3d5dda18d0484c07c0b2d311de6/jersey/contribs/jersey-apache-client/src/main/java/com/sun/jersey/client/apache/DefaultApacheHttpMethodExecutor.java#L120
If the `Entity` is an empty string and the `chunkedEncodingSize` property is null the method `writeOutBoundHeaders(...)` will not invoked.

PR #4121, opened 2017-07-06
[PDI-16221] "file" option of Kitchen and Pan do not work in the file system

Advent51:
- All possible options are covered now


Advent51:

PR #4127, opened 2017-07-06
[BACKLOG-16971] Fixes displaying tooltips only when search text overflows past the width of the dropdown menu.

e-cuellar:


e-cuellar:
@ddiroma Please Sir!

PR #4128, opened 2017-07-07
[BACKLOG-17332] file sorting on open/save dialog

tgf:


tgf:
@ddiroma @bmorrise please review

plugins/file-open-save/core/src/main/javascript/app/components/files/files.component.js
```diff
@@ -183,14 +184,12 @@ define([
       vm.sortState = vm.sortField === field ? vm.sortState : 0;
       switch (vm.sortState) {
```
ddiroma: @tgf This does not allow the user to clear all sorting. Meaning, 1st click will sort ascending. 2nd click will sort descending. 3rd click will sort ascending again and so forth. Does UX want the 3rd click to clear all sorting? If so, this will need to be fixed. If not, then we are good.

ddiroma: @tgf @bmorrise ...got clarification from UX. Wants same behavior as DET. DET toggles asc/desc after 1st click. We are good. :)

PR #4129, opened 2017-07-07
[BACKLOG-16971] Fixes the magnifying glass icon is not turning blue on focus when there is no previous search

e-cuellar:


e-cuellar:
@ddiroma Please Sir!

PR #4134, opened 2017-07-09
[BACKLOG-16971] Fixes if the user clicks to the left or right side of…

e-cuellar:
… the search box the search box does not become active


e-cuellar:
@bmorrise @ddiroma Please review....

PR #4137, opened 2017-07-10
[BACKLOG-17332] file open/save sort: was breaking when same name; also sort tree folders

tgf:


tgf:
@bmorrise please review

PR #4138, opened 2017-07-11
[BACKLOG-17588] NPE parsing property

tgf:


tgf:
@bmorrise please review

PR #4139, opened 2017-07-11
[BACKLOG-17332] sort tree folders with proper hierarchy

tgf:


tgf:
@bmorrise please review

PR #4143, opened 2017-07-12
[PDI-16337] - Error reading from Oracle Number(38) column

ArtsiomYurhevich:
Added new advanced database connection option "Strict number(38) interpretation (Oracle specific) ".

 If true Oracle NUMBER with 38 precision is interpreted as BigDecimal

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
+++ b/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
@@ -4728,7 +4728,8 @@ public class ValueMetaBase implements ValueMetaInterface {
 
           if ( databaseMeta.getDatabaseInterface() instanceof OracleDatabaseMeta ) {
             if ( precision == 0 && length == 38 ) {
-              valtype = ValueMetaInterface.TYPE_INTEGER;
+              valtype = ( (OracleDatabaseMeta) databaseMeta.getDatabaseInterface() )
+                .strictBigNumberInterpretation() ? TYPE_BIGNUMBER : TYPE_INTEGER;
             }
             if ( precision <= 0 && length <= 0 ) {
               // undefined size: BIGNUMBER,
--- a/dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
+++ b/dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
@@ -43,6 +43,7 @@ import org.pentaho.di.core.database.DatabaseInterface;
 import org.pentaho.di.core.database.DatabaseMeta;
 import org.pentaho.di.core.database.GenericDatabaseMeta;
 import org.pentaho.di.core.database.MSSQLServerNativeDatabaseMeta;
+import org.pentaho.di.core.database.OracleDatabaseMeta;
 import org.pentaho.di.core.database.PartitionDatabaseMeta;
 import org.pentaho.di.core.exception.KettleException;
 import org.pentaho.di.core.logging.LogChannel;
@@ -203,6 +204,8 @@ public class DataHandler extends AbstractXulEventHandler {
   XulCheckbox upperCaseIdentifiersCheck;
 
   XulCheckbox preserveReservedCaseCheck;
+
+  XulCheckbox strictBigNumberInterpretaion;
 
   XulCheckbox useIntegratedSecurityCheck;
 
@@ -690,6 +693,11 @@ public class DataHandler extends AbstractXulEventHandler {
       meta.setPreserveReservedCase( preserveReservedCaseCheck.isChecked() );
     }
 
+    if ( strictBigNumberInterpretaion != null && meta.getDatabaseInterface() instanceof OracleDatabaseMeta ) {
+      ( (OracleDatabaseMeta) meta.getDatabaseInterface() )
+        .setStrictBigNumberInterpretation( strictBigNumberInterpretaion.isChecked() );
+    }
+
     if ( preferredSchemaName != null ) {
       meta.setPreferredSchemaName( preferredSchemaName.getValue() );
     }
@@ -831,7 +839,6 @@ public class DataHandler extends AbstractXulEventHandler {
     setOptionsData( meta.getExtraOptions() );
 
     // Advanced panel settings:
-
     if ( supportBooleanDataType != null ) {
       supportBooleanDataType.setChecked( meta.supportsBooleanDataType() );
     }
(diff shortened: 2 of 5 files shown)


kolinus:
@pedrofvteixeira @pamval @mdamour1976  could you please take a look

pamval:
Do we really need to alter the base interface and class with a property that is specific to the OracleDatabaseMeta?

ArtsiomYurhevich:
@pamval Otherwise we would have this checkbox available for checking/unchecking for all types of connections , but it wouldn't affect anything and it may confuse someone. If we shouldn't  change interface, i will remove the logic.

ArtsiomYurhevich:
@pamval What do you think?

core/src/main/java/org/pentaho/di/core/database/DatabaseMeta.java
```diff
@@ -2321,13 +2321,30 @@ public boolean preserveReservedCase() {
   }
 
   /**
+   * @return true if using strict number(38) interpretation ( Oracle specific )
```
pamval: No need to change DatabaseMeta. Remove these methods

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -4728,7 +4728,7 @@ public ValueMetaInterface getValueFromSQLType( DatabaseMeta databaseMeta, String
 
           if ( databaseMeta.getDatabaseInterface() instanceof OracleDatabaseMeta ) {
             if ( precision == 0 && length == 38 ) {
-              valtype = ValueMetaInterface.TYPE_INTEGER;
+              valtype = databaseMeta.strictBigNumberInterpretation() ? TYPE_BIGNUMBER : TYPE_INTEGER;
```
pamval: You already know it's a OracleDatabaseMeta otherwise you would've not entered this conditional branch. Cast databaseMeta to OracleDatabaseMeta so you can call the method.

dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
```diff
@@ -856,6 +862,17 @@ private void setInfo( DatabaseMeta meta ) {
       preserveReservedCaseCheck.setChecked( meta.preserveReservedCase() );
     }
 
+    if ( strictBigNumberInterpretaion != null ) {
+      //check if Oracle
+      if ( meta.getDatabaseInterface() instanceof OracleDatabaseMeta ) {
+        strictBigNumberInterpretaion.setVisible( true );
+        strictBigNumberInterpretaion.setChecked( meta.strictBigNumberInterpretation() );
```
pamval: Same as above, you already know it's a OracleDatabaseMeta, cast it.

dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
```diff
@@ -856,6 +862,17 @@ private void setInfo( DatabaseMeta meta ) {
...
+      } else {
+        strictBigNumberInterpretaion.setVisible( false );
+        strictBigNumberInterpretaion.setChecked( false );
```
pamval: Remove this line as it is not needed

dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
```diff
@@ -1320,6 +1337,18 @@ private void setConnectionSpecificInfo( DatabaseMeta meta ) {
...
 
+    // Strict Number(38) interpretation
+    if ( strictBigNumberInterpretaion != null ) {
+      // Check if oracle
+      if ( meta.getDatabaseInterface() instanceof OracleDatabaseMeta ) {
+        strictBigNumberInterpretaion.setVisible( true );
+        strictBigNumberInterpretaion.setChecked( meta.strictBigNumberInterpretation() );
+      } else {
+        strictBigNumberInterpretaion.setVisible( false );
+        strictBigNumberInterpretaion.setChecked( false );
```
pamval: Remove this line as it is not needed.

dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
```diff
@@ -1320,6 +1337,18 @@ private void setConnectionSpecificInfo( DatabaseMeta meta ) {
       indexTablespaceBox.setValue( meta.getIndexTablespace() );
     }
 
+    // Strict Number(38) interpretation
+    if ( strictBigNumberInterpretaion != null ) {
+      // Check if oracle
+      if ( meta.getDatabaseInterface() instanceof OracleDatabaseMeta ) {
+        strictBigNumberInterpretaion.setVisible( true );
+        strictBigNumberInterpretaion.setChecked( meta.strictBigNumberInterpretation() );
```
pamval: Cast meta to OracleDatabaseMeta and call the method

dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
ArtsiomYurhevich: At this point we do not know whether meta is Oracle or not. Xul checkboxes in advanced options tab are defined in https://github.com/pentaho/pentaho-kettle/blob/master/dbdialog/src/main/resources/org/pentaho/ui/database/databasedialog.xul#L93-L109 and are same for all types of connections. So checking the "strictBigNumberInterpretaion" for doesn't  guarantee that the meta is Oracle.

dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
ArtsiomYurhevich: We shouldn't remove these lines because we should always reset the state of the checkbox. For example we select oracle, check out checkbox and then switch to another type of connection. So, the state of the checkbox will stay True for the non-oracle connection

dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
ArtsiomYurhevich: the same

core/src/main/java/org/pentaho/di/core/database/DatabaseMeta.java
ArtsiomYurhevich: Removed

ArtsiomYurhevich:
Removed changes from databasemeta. Some notes: From one side we do not want to add any specific methods to our base classes and interfaces but from other side the point overloads our ui logic. Ideally DataHandler do not know about OracleDatabaseMeta at all

ArtsiomYurhevich:
Also it can be simplified by implementing dynamic xul loading with advanced checkboxes for specific connections. For now they are static for all types of connections. But if implementing the logic because of only one type of connection and only one checkbox may exceed the costs and code load compared with just checking instance of in datahandler

PR #4160, opened 2017-07-14
[BACKLOG-17645] - Create a new default property to be able to disable the environment+browser check

eliofreitas:


ui/src/main/java/org/pentaho/di/ui/core/PropsUI.java
```diff
@@ -1221,6 +1221,17 @@ public int getSupportedVersion( String property ) {
     return Integer.parseInt( properties.getProperty( property ) );
   }
 
+  /**
+   * isBrowserEnvironmentCheckDisabled
```
afrjorge: you don't need to add the function name here

ui/src/main/java/org/pentaho/di/ui/util/EnvironmentUtils.java
```diff
@@ -152,6 +152,17 @@ protected int getSupportedVersion( String property ) {
   }
 
   /**
+   * isBrowserEnvironmentCheckDisabled
```
afrjorge: same here

ui/src/main/resources/org/pentaho/di/ui/core/default.properties
```diff
@@ -8,6 +8,7 @@ BackgroundColorB=255
 BackgroundColorG=255
 BackgroundColorR=255
 DefaultPreviewSize=1000
+DisableBrowserEnvironmentCheck=false
```
afrjorge: Are you sure about this? All other values seem to be `Y`/`N` instead of `true`/`false`.

ui/src/main/java/org/pentaho/di/ui/core/PropsUI.java
```diff
@@ -1212,7 +1212,7 @@ public void setCanvasGridSize( int gridSize ) {
   }
 
   /**
-   * getSupportedVersion
+   * gets the supported version of the requested software
```
afrjorge: This sentence should be formatted according to the other descriptions.

ui/src/main/java/org/pentaho/di/ui/core/PropsUI.java
```diff
@@ -1221,6 +1221,15 @@ public int getSupportedVersion( String property ) {
     return Integer.parseInt( properties.getProperty( property ) );
   }
 
+  /**
+   * Ask if the browsing environment checks are disabled.
+   *
+   * @return 'true' if disabled 'false' otherwise.
+   */
+  public boolean isBrowserEnvironmentCheckDisabled() {
+    return "Y".equalsIgnoreCase( properties.getProperty( "DisableBrowserEnvironmentCheck", "N" ) );
```
afrjorge: "DisableBrowserEnvironmentCheck" should be a `private static final String`

PR #4161, opened 2017-07-14
[BACKLOG-15718] File Open Save fixes for the following...

ddiroma:
7 - Right now the text in the cards (name and date info) are not vertically aligned.
11 - Before vertical scrollbar appears, maintain 20px gutter to right, currently looks bigger than 20px.
12 - When vertical scrollbar appears, maintain 20px gutter to right of card between card edge and scrollbar (dev to see if thats possible).
14 - At the end of the 12 recent files, there is no 20px of space under the last row of cards. Right now they stick to the container edge.
22 - In the recents view, the cards are not displaying in 2 columns wide by default. They are stacking in one column
23 - "You haven't opened anything recently." not centered on Ubuntu
63 - When renaming a file/folder the styling of the row shifts
71 - in the save dialog, clicking a folder name in the right pane is populating the "File name" text box. It should only populate in that box when a file is selected.
90 - Scrolling bar is cutting the last file in the list view (see screen shot column)
109 - Search label should be using open sans regular, its using arial currently.
115 - The open dialog width should be 930px, it looks bigger
118 - Space in footer of dialog


ddiroma:
@bmorrise

PR #4163, opened 2017-07-15
Enhance kettle gpload plugin to support unicode (non-printable ASCII …

float2net:
# Enhance kettle gpload plugin to support unicode(non-printable ASCII) delimiter
Use case of non-printable ASCII character delimiter is very common. however the current gpload plugin doesn't support unicode (non-printable ASCII character) delimiter correctly, this enhancement will fix it. 

With this PR, unicode delimiter can be accepted in UI with prefix of **\u** , e,g, *\u001b*, as shown below:
gpload_ui_unicode

Later on running the plugin will produce gpload control file having a unicode 'DELIMITER' field, quoted with **double** quote (instead of single quote on common printable delimiter), as shown below:
control0_better

As GPLOAD command line utility usage stated below:
gpload_cmd_gramma

A produced sample gpload load0.data file (with ESC(\u001b) delimiter):
load0_unicode_delimiter

And the file shall be successfully imported into Greenplum database:
db


float2net:
Test delimiter with the format $ [value] , e.g. $ [01], it doesn't work. the gpload control file do not produce.
There is error on console: "A delimiter must be specified"
Screen capture follow:
env-sub-ui
env-sub-err

float2net:
The environment substitute on '$[1b]' produce empty:
gpload_debug_code
gpload_debug

PR #4164, opened 2017-07-16
[PDI-6834] Fix saving of transformation log database and table in database repos…

j-a-n:
…itory

I configured transformation logging by setting two environment variables in kettle.properties:

KETTLE_TRANS_LOG_DB=logdb
KETTLE_TRANS_LOG_TABLE=translog

Kettle should use them as default for every transformation if you do not set the values in the logging dialog of the transformation.
This is working als long as you do not save the transformation to a database repository.

In that case kettle stores some values in R_TRANS_ATTRIBUTE:
TRANS_LOG_TABLE_CONNECTION_NAME = NULL
TRANS_LOG_TABLE_TABLE_NAME = NULL
which is correct.

In addition kettle sets ID_DATABASE_LOG and TABLE_NAME_LOG in R_TRANSFORMATION.
I do not know why this is required.

logTable.getDatabaseMeta() is used to determine the value for TABLE_NAME_LOG.
getDatabaseMeta() calls getActualConnectionName() which returns the value set in the transformation respectively the value set in environment.

In my opinion this is wrong for two reasons:
Firstly this returns the name of the log database connection not the log database name.
Secondly this value should be null if not explicitly set in the transformation.

When the transformation is loaded from the database repository the value stored in TABLE_NAME_LOG is set via BaseLogTable.setTableName().
This value would be overwritten by R_TRANS_ATTRIBUTE.TRANS_LOG_TABLE_TABLE_NAME if it wouldn't be NULL.

As the result the log table name is set to the log database connection name in transformation logging properties, which is definitely wrong.
The log connection is set to the right value, but with the value explicitly set, a change via the environment variable KETTLE_TRANS_LOG_DB is no longer possible.

My patch changes this behaviour.
ID_DATABASE_LOG and TABLE_NAME_LOG are only set if connectionName respectively tableName are explicitly set, and the right values are used.

The patch fixes the issue described in [PDI-6834](http://jira.pentaho.com/browse/PDI-6834)


j-a-n:
This pull request is open for a very long time now.
Is there something I can do to get it merged?

mbatchelor:
@j-a-n - Thanks for the contribution, and sorry this has languished so long. So before we can process the PR, we need a unit test that covers the change in behavior. Please update the PR with unit test for the changes, and then assign the JIRA case to Triage. Thanks!

j-a-n:
@mbatchelor Sorry, I really need some help with the unit test.

PR #4168, opened 2017-07-17
[TEST] locale loadsave could fail for default and have wrong error message

tgf:


tgf:
@bmorrise please review

PR #4169, opened 2017-07-17
[BACKLOG-15720] Fixes the Home and Public label in the search box are appearing lower case.

e-cuellar:


e-cuellar:
@ddiroma Please Sir!

PR #4172, opened 2017-07-18
[PDI-9323] dealing with decimals in checksum, comparisons etc not according to value

Advent51:
- checksum calculates on native data types
- added tests


Advent51:
@mbatchelor @pamval @mchen-len-son @mdamour1976 please review

PR #4176, opened 2017-07-19
[PDI-14442]-PDI Import scripts are executed in parallel by multiple users

stepanovdg:


stepanovdg:
IT Tests for Pur repository are still skipped http://jira.pentaho.com/browse/BACKLOG-17551

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepository.java
```diff
@@ -2999,9 +3000,21 @@ protected void saveTransOrJob( ISharedObjectsTransformer objectTransformer, Repo
           new RepositoryFile.Builder( file ).title( RepositoryFile.DEFAULT_LOCALE, element.getName() ).createdDate(
               versionDate != null ? versionDate.getTime() : new Date() ).description( RepositoryFile.DEFAULT_LOCALE,
               Const.NVL( element.getDescription(), "" ) ).build();
-      file =
+      try {
```
mbatchelor: Need to see a test case before this can be merged.

stepanovdg:
For reproducing was used importParallel.sh script (see atached). 
Steps:
- in PDI - create connection to repository with name Admin
- in repository create dir test in Public catalog (/Public/test)
- in import-rules.xml - disable all checks (false value in tag enabled)
- copy design-tools/data-integration/samples/transformations/Getting Started Transformation.ktr to design-tools/data-integration/
- copy importParallel.sh to design-tools/data-integration/
- set run permissions "chmod +x ./importParallel.sh"
- run script ./importParallel.sh
- See results in import.log file.

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepository.java
stepanovdg: Added reproduce scenario.

PR #4179, opened 2017-07-20
[SP-3645][BAD-613]-Reg: PMR Hive job fails on secured cluster if principal is set by variable

stepanovdg:


stepanovdg:
@mchen-len-son @mbatchelor Could you please take a look?

PR #4180, opened 2017-07-20
[SP-3691] Backport of PDI-16318 - REST Client Step: Http Method DELETE results in 401 where GET, PUT succeed. Postman works fine with OK/200 (7.1 Suite)

Advent51:


Advent51:

PR #4181, opened 2017-07-20
[SP-3661] Backport of PDI-16312 - JSON Input step ignores the trim type (7.1 Suite)

Advent51:


Advent51:

PR #4183, opened 2017-07-20
[SP-3694] Backport of PDI-16221 - "file" option of Kitchen and Pan do not work in the file system (7.1 Suite)

Advent51:


Advent51:

PR #4185, opened 2017-07-20
[SP-3676][PDI-16388] - SelectValues type change not using default conversion mask

stepanovdg:


stepanovdg:

stepanovdg:
@pamval @duarteteixeira Could you please take a look?

PR #4187, opened 2017-07-20
[PDI-9388] Deleting a column from the Datagrid step moves column names as heading for wrong data

AliaksandrShuhayeu:
[PDI-9388] Deleting a column from the Datagrid step moves column names as heading for wrong data

added synchronization between Fields and Data tables


AliaksandrShuhayeu:

pamval:
@AliaksandrShuhayeu isn't the tableview the widget used everywhere in Spoon? Don't we need to create listeners in other places to keep current functionality elsewhere (for instance, fields table in pretty much every step) ?

AliaksandrShuhayeu:
@pamval Yes, tableview is used everywhere in Spoon. 
No, we dont need create aditional listeners. It has dummy default realization (TableView.java L194).
We need add realization in case when we need to control modification (for example our component: 2 synchronized tabs ).

PR #4200, opened 2017-07-24
[BACKLOG-17332] folders on bottom when reversed; secondary sort by name

tgf:


tgf:
@bmorrise @ddiroma please review

PR #4202, opened 2017-07-25
[PDI-14346] Error handling in Insert/Update from sub transformation

AliaksandrShuhayeu:
-added returning false in error case


AliaksandrShuhayeu:

PR #4213, opened 2017-07-27
[PMD-679]-Variable symbols displaying in the database dialog; PME does not support variables

denisprotopopov:
-  replaced the variable-aware fields with normal fields and removed the
 variable indicators next to the fields


denisprotopopov:
@mbatchelor @pamval please review the changes

PR #4214, opened 2017-07-27
[PDI-13416] Adding a duplicate method in UDJC causes exception and crashes Spoon

AliaksandrShuhayeu:
added running compilation customer's class in separate thread for avoid crash the application in case of an unchecked compilation error

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/userdefinedjavaclass/UserDefinedJavaClassMeta.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/userdefinedjavaclass/UserDefinedJavaClassMeta.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2016 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2017 by Pentaho : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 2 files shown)


AliaksandrShuhayeu:
@pamval Could you please review and merge it?

engine/src/main/java/org/pentaho/di/trans/steps/userdefinedjavaclass/UserDefinedJavaClassMeta.java
```diff
@@ -139,9 +136,37 @@ public UserDefinedJavaClassMeta() {
       "org.pentaho.di.trans.steps.userdefinedjavaclass.*", "org.pentaho.di.trans.step.*",
       "org.pentaho.di.core.row.*", "org.pentaho.di.core.*", "org.pentaho.di.core.exception.*" } );
 
-    cbe.cook( new Scanner( null, sr ) );
```
pamval: If the issue is to protect the compilation from throwing unchecked errors, maybe having a try {} catch (Throwable) {} around this line is enough instead of introducing a new Thread

engine/src/main/java/org/pentaho/di/trans/steps/userdefinedjavaclass/UserDefinedJavaClassMeta.java
```diff
@@ -139,9 +136,37 @@ public UserDefinedJavaClassMeta() {
...
+        finishedCorrect[ 0 ] = true;
+      } catch ( Exception e ) {
+        checkedException[ 0 ] = e;
+      }
+    };
+
+    Thread thread = new Thread( runnable );
+    thread.start();
+
+    while ( true ) {
```
pamval: If we really must compile in a separate thread, please void this synchronization code. You should use Executors/Future (or wait/notify along with a monitor object if you really feel like writing synchronization code)

AliaksandrShuhayeu:
@pamval Thanks for your correction. I'm sorry. I was wrong. Could you please review and merge it?

PR #4215, opened 2017-07-27
[SP-3678][PDI-16352] Merge Rows Diff check row layout

mchen-len-son:


mchen-len-son:
@stepanovdg please take a look

PR #4218, opened 2017-07-28
[SP-3587] Backport of PDI-15982 - Incomplete/Missing logs when Main Job with sub jobs are scheduled (6.1 Suite)

Advent51:
- backport of #3906


Advent51:
@mchen-len-son @pamval please review

PR #4223, opened 2017-07-30
[SP-3687][PDI-15858] Repository Manager dialog is blank - not able to…

mchen-len-son:
… create any repository

 - added meta tag to repositories index page forcing it to IE11 mode


mchen-len-son:
@pamval @stepanovdg @mbatchelor please take a look.

PR #4225, opened 2017-07-31
[PDI-16337]-Spoon db dialog cache doesn't update value of database interface after changing db type

ArtsiomYurhevich:
Now after selecting and changing database type, db cache will hold and update the correct value of database interface


ArtsiomYurhevich:
@pamval Could you ,please, take a look ?

PR #4245, opened 2017-08-03
[BACKLOG-17671] TableView column resize

tgf:
TableView: allow disabling automatic column expansion for contents
ColumnsResizer: account for vscrollbar when resizing, option to refit
after manual column resize

### DIFF
--- a/ui/src/main/java/org/pentaho/di/ui/core/widget/ColumnInfo.java
+++ b/ui/src/main/java/org/pentaho/di/ui/core/widget/ColumnInfo.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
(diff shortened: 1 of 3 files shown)


tgf:
@ddiroma please review

roboguy:
@tgf - Is there any plan to allow user control of this behavior ? For example in the `TransPreviewDelegate` that controls the preview table .. we usually have long strings that do not fit well in this table. We would like to be able to set autoResize to false.

tgf:
@roboguy if you mean user control from the ui, then no; it's just meant to be set when creating the table. otherwise this can be used in preview or anything else using tableview

roboguy:
@tgf - thank you for replying.

PR #4246, opened 2017-08-03
[BACKLOG-17366] - I want to support multiple topic ingestion from a s…

Aliaksandr-Kastenka:
…treaming ingestion step(insert variables icon moved next to the topics table)

"Insert variables" icon moved accoroding to UX recommendation.
Bug with the lack of changing font of transformation name(tab name) to bold in case of adding topic with variables, or changing subtrans path addressed.
Goes with https://github.com/pentaho/pentaho-ee/pull/476


Aliaksandr-Kastenka:
@DFieldFL @kurtwalker 
Please review.

PR #4249, opened 2017-08-05
[PDI-2301]-Slave server monitor refresh freezes Spoon

ArtsiomYurhevich:
-Optimized LoggingBuffer( by using concurrent collections and streams)
-Implemented local-temp-file cache for keeping statuses of stopped or finished transformations or jobs which significantly decreases carte response time and cpu usage ( especially for medium and large logs)
-The saved log content will be alive during specified by user properties in kettle.properties ( By default 1 day)
-The changes remove restrictions to saving limited number of logs ( It was 5000 lines by default before ) since statuses are saved as temp files, not in memory
-Optimized SpoonSlave:
  -Disabled excess events( which requested ktr/kjb status many times at one moment )
  -Fixed memory leak. SpoonSlave saved all the carte ktr/kjbs logs in map (id,log) and never cleared it( even if it was removed ) . In case of large logs we may catch OOME and Spoon crash.
  -Update slave view time is configurable now via kettle.properties (SPOON_CARTE_VIEW_UPDATE_TIME)
-Fixed critical memory leak which concerns xml parsing( XMLHandlerCache ). The implementation of it is incomplete. XMLHandler ,which we use every time we need to parse xml, caches nodes(which may be large ones) and never clears it. Sooner(if we parse large xml files) or later it may cause spoon crash (oome) . Reimplemented the cache with using WeakHashMap. You may face with it when running ktrs/kjbs with large logs on carte via spoon and after some time spoon will crash. Slave tab in spoon periodically sends multiple request to carte for status updates, parses xml responses  and then in profiler you could see large objects representing not cleared nodes. Note that it does not concern only slave tab but everywhere we use XmlHandler .

### DIFF
--- a/engine/src/main/java/org/pentaho/di/www/PrepareExecutionTransServlet.java
+++ b/engine/src/main/java/org/pentaho/di/www/PrepareExecutionTransServlet.java
@@ -39,6 +39,7 @@ import org.pentaho.di.i18n.BaseMessages;
 import org.pentaho.di.trans.Trans;
 import org.pentaho.di.trans.TransConfiguration;
 import org.pentaho.di.trans.TransExecutionConfiguration;
+import org.pentaho.di.www.cache.CarteStatusCache;
 
 
 public class PrepareExecutionTransServlet extends BaseHttpServlet implements CartePluginInterface {
@@ -46,6 +47,7 @@ public class PrepareExecutionTransServlet extends BaseHttpServlet implements Car
 
   private static final long serialVersionUID = 3634806745372015720L;
   public static final String CONTEXT_PATH = "/kettle/prepareExec";
+  private CarteStatusCache cache = CarteStatusCache.getInstance();
 
   public PrepareExecutionTransServlet() {
   }
@@ -214,6 +216,7 @@ public class PrepareExecutionTransServlet extends BaseHttpServlet implements Car
 
         try {
           trans.prepareExecution( null );
+          cache.remove( trans.getLogChannelId() );
 
           if ( useXML ) {
             out.println( WebResult.OK.getXML() );
--- a/engine/src/main/java/org/pentaho/di/www/RemoveJobServlet.java
+++ b/engine/src/main/java/org/pentaho/di/www/RemoveJobServlet.java
@@ -29,6 +29,7 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.owasp.encoder.Encode;
 import org.pentaho.di.core.Const;
 import org.pentaho.di.core.util.Utils;
@@ -36,6 +37,7 @@ import org.pentaho.di.core.logging.KettleLogStore;
 import org.pentaho.di.core.xml.XMLHandler;
 import org.pentaho.di.i18n.BaseMessages;
 import org.pentaho.di.job.Job;
+import org.pentaho.di.www.cache.CarteStatusCache;
 
 
 public class RemoveJobServlet extends BaseHttpServlet implements CartePluginInterface {
@@ -44,6 +46,9 @@ public class RemoveJobServlet extends BaseHttpServlet implements CartePluginInte
   private static final long serialVersionUID = -2051906998698124039L;
 
   public static final String CONTEXT_PATH = "/kettle/removeJob";
+
+  @VisibleForTesting
+  private CarteStatusCache cache = CarteStatusCache.getInstance();
 
   public RemoveJobServlet() {
   }
@@ -184,6 +189,7 @@ public class RemoveJobServlet extends BaseHttpServlet implements CartePluginInte
 
     if ( job != null ) {
 
+      cache.remove( job.getLogChannelId() );
       KettleLogStore.discardLines( job.getLogChannelId(), true );
       getJobMap().removeJob( entry );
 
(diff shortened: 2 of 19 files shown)


ArtsiomYurhevich:
@mbatchelor Could you please take a look?

mchen-len-son:
@pamval @bmorrise can you please take a look?
thanks.

kolinus:
@mdamour1976 could you please take a look

kolinus:
@mbatchelor could you please review

engine/src/main/java/org/pentaho/di/www/PrepareExecutionTransServlet.java
```diff
@@ -39,13 +39,15 @@
...
 import org.pentaho.di.trans.TransExecutionConfiguration;
+import org.pentaho.di.www.cache.TransJobStatusCache;
 
 
 public class PrepareExecutionTransServlet extends BaseHttpServlet implements CartePluginInterface {
   private static Class<?> PKG = PrepareExecutionTransServlet.class; // for i18n purposes, needed by Translator2!!
 
   private static final long serialVersionUID = 3634806745372015720L;
   public static final String CONTEXT_PATH = "/kettle/prepareExec";
+  private TransJobStatusCache cache = TransJobStatusCache.getInstance();
```
mbatchelor: Since servlets are singletons, I don't think you should do this.

engine/src/main/java/org/pentaho/di/www/RemoveJobServlet.java
```diff
@@ -44,6 +45,7 @@
   private static final long serialVersionUID = -2051906998698124039L;
 
   public static final String CONTEXT_PATH = "/kettle/removeJob";
+  private TransJobStatusCache cache = TransJobStatusCache.getInstance();
```
mbatchelor: As above

engine/src/main/java/org/pentaho/di/www/RemoveTransServlet.java
```diff
@@ -45,6 +46,8 @@
 
   public static final String CONTEXT_PATH = "/kettle/removeTrans";
 
+  private TransJobStatusCache cache = TransJobStatusCache.getInstance();
```
mbatchelor: Ditto

engine/src/main/java/org/pentaho/di/www/PrepareExecutionTransServlet.java
ArtsiomYurhevich: @mbatchelor Marc, do you mean that we should TransJobStatusCache.getInstance() each time we need cache? Could you please clarify we we should do so?

ArtsiomYurhevich:
@mbatchelor Could you please take a look?

ArtsiomYurhevich:
@mbatchelor Review please

mchen-len-son:
@pamval @mbatchelor can you please take a look?
thanks.

engine/src/main/java/org/pentaho/di/www/PrepareExecutionTransServlet.java
mbatchelor: normally, instance variables in a servlet are no-nos. In this case it's likely OK.

engine/src/main/java/org/pentaho/di/www/cache/TransJobStatusCache.java
```diff
@@ -0,0 +1,130 @@
...
+import java.time.LocalDate;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+
+public class TransJobStatusCache {
```
mbatchelor: There needs to be an interface for this so that on the server, we could use the built-in CacheManager for managing these entries.

engine/src/main/java/org/pentaho/di/www/cache/CachedItem.java
```diff
@@ -0,0 +1,64 @@
...
+public class CachedItem {
+
+  private String logId;
+  private File file;
+  private LocalDate exceedTime;
+
+  public CachedItem( String logId, File file ) {
+    this.logId = logId;
+    this.file = file;
+    exceedTime = LocalDate.now().plusDays( 1 );
```
mbatchelor: This needs to be configurable

engine/src/main/java/org/pentaho/di/www/cache/TransJobStatusCache.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+    return instance;
+  }
+
+  public TransJobStatusCache() {
+    removeService.scheduleAtFixedRate( () -> cachedMap.forEach( ( k, v ) -> {
+      if ( LocalDate.now().isAfter( v.getExceedTime() ) ) {
+        remove( k );
+      }
+    } ), 1, 1, TimeUnit.DAYS );
```
mbatchelor: The TimeUnit.DAYS bit needs to be configurable. Consider an extremely busy server that builds up these temp files rapidly and they need to clean this out every couple of hours.

engine/src/main/java/org/pentaho/di/www/cache/CachedItem.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ *
+ ******************************************************************************/
+
+package org.pentaho.di.www.cache;
+
+import java.io.File;
+import java.time.LocalDate;
+
+public class CachedItem {
```
mbatchelor: For overriding the cache below, this class should implement Serializable

mchen-len-son:
@mbatchelor @bmorrise please have a look
thanks.

PR #4250, opened 2017-08-07
[PDI-14721] Problem with MySQL and BLOBs - ValueMetaInterface.TYPE_BINARY is being set to ValueMetaInterface.STRING

AliaksandrShuhayeu:
-there was deleted setting wrong type

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
+++ b/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
@@ -4791,8 +4791,6 @@ public class ValueMetaBase implements ValueMetaInterface {
             length = rm.getColumnDisplaySize( index );
           } else if ( databaseMeta.isMySQLVariant()
               && ( type == java.sql.Types.VARBINARY || type == java.sql.Types.LONGVARBINARY ) ) {
-            // set the data type to String, see PDI-4812
-            valtype = ValueMetaInterface.TYPE_STRING;
             // PDI-6677 - don't call 'length = rm.getColumnDisplaySize(index);'
             length = -1; // keep the length to -1, e.g. for string functions (e.g.
             // CONCAT see PDI-4812)
--- a/core/src/test/java/org/pentaho/di/core/database/DatabaseUnitTest.java
+++ b/core/src/test/java/org/pentaho/di/core/database/DatabaseUnitTest.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2016 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2017 by Pentaho : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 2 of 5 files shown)


AliaksandrShuhayeu:
It is fix for PDI-14721 , ESR-5021 . 
Also I tested PDI-4812 . Current changes don't broke it.
With this fix we don't need part of fix PDI-8373 . My commit contains removing a part of fix PDI-8373 with a workaround. I tested it too. Current changes don't broke PDI-8373 . 
I tested connected PDI-8290 . PPDI-8290 is not duplicate and has a little other case in the transformation. PDI-8373 and PDI-8290 work correct with my changes. 
PDI-8373 has attachment 42321.zip. It doesn't work correct in master but my changes fix it too.
Short.
Before fix (master now): 
PDI-14721 , ESR-5021 There is this issue 
PDI-4812 work correct 
PDI-8290 work correct 
PDI-8373 attachment 42321.zip have a customer problem. It was not fixed.
Current changes: 
PDI-14721 , ESR-5021 , PDI-4812 , PDI-8290 , PDI-8373 work correct

AliaksandrShuhayeu:
@mbatchelor Could you please review it?

core/src/main/java/org/pentaho/di/core/database/Database.java
```diff
@@ -2858,11 +2858,7 @@ public void setProcLookup( String proc, String[] arg, String[] argdir, int[] arg
   }
 
   public Object[] getLookup( boolean failOnMultipleResults ) throws KettleDatabaseException {
-    return getLookup( failOnMultipleResults, false );
-  }
-
-  public Object[] getLookup( boolean failOnMultipleResults, boolean lazyConversion ) throws KettleDatabaseException {
```
mbatchelor: Why was this method removed ? It's a public method in Database, so it could be called by lots of other things. Don't believe you can do this.

core/src/main/java/org/pentaho/di/core/database/Database.java
```diff
@@ -2873,17 +2869,12 @@ public void setProcLookup( String proc, String[] arg, String[] argdir, int[] arg
...
-
-  public Object[] getLookup( PreparedStatement ps, boolean failOnMultipleResults, boolean lazyConversion )
-    throws KettleDatabaseException {
     ResultSet res = null;
     try {
       log.snap( Metrics.METRIC_DATABASE_GET_LOOKUP_START, databaseMeta.getName() );
       res = ps.executeQuery();
 
-      Object[] ret = getRow( res, lazyConversion );
+      Object[] ret = getRow( res );
```
mbatchelor: Why was the specification of lazy conversion removed here ? I think it needs to be put back.

AliaksandrShuhayeu:
All these strings were added in scope of  PDI-8373. All these strings are a workaround in scope of PDI-8373. 
Object[] ret = getRow( res, lazyConversion );
We could get second argument yes only in org.pentaho.di.trans.steps.databaselookup.DatabaseLookup L=141 . I removed it as not actual workaround.
I deleted my changes in Database.java. But these strings are not used in kettle more.

AliaksandrShuhayeu:
@mbatchelor Could you please review it?
All these strings were added in scope of PDI-8373. All these strings are a workaround in scope of PDI-8373.
Object[] ret = getRow( res, lazyConversion );
We could get second argument yes only in org.pentaho.di.trans.steps.databaselookup.DatabaseLookup L=141 . I removed it as not actual workaround.
I deleted my changes in Database.java. But these strings are not used in kettle more.

PR #4256, opened 2017-08-08
[PDI-16365]-Spaces in connection name can cause unexpected behavior

denisprotopopov:
- regression fix: changing name of connection into new (non existing)


denisprotopopov:
@pamval please review the changes

PR #4257, opened 2017-08-08
[PDI-16221] "file" option of Kitchen and Pan do not work in the file system

Advent51:
- fixed definition of the absolute path


Advent51:
@pamval @stepanovdg @mbatchelor please review

PR #4263, opened 2017-08-09
[BACKLOG-18135] - Can't run a transformation on Carte from spoon

YuryBY:


YuryBY:
@pamval, could you please take a look?

PR #4291, opened 2017-08-15
[BACKLOG-18167] Renaming a name in Open dialog doesn't refresh the tr…

ddiroma:
…ansformation/job tab


ddiroma:
@bmorrise 
Please review. I am unsure if the KettleExceptions I am using in RepositoryBrowserController.java are the correct ones.

Also, in RepositoryBrowserEndpoint.java, I am not 100% sure if I should use Response.Status.NOT_ACCEPTABLE when catching those exceptions.

Please offer input. Thanks.

PR #4292, opened 2017-08-16
[PDI-16430]-The PostgreSQL bulk loader" truncate action exists an error

denisprotopopov:
- applied fix (https://github.com/pentaho/pentaho-kettle/pull/4255), added unit test


denisprotopopov:
@pamval @mbatchelor please review the changes

zfruc:
There is a small problem from line 100 to line 104 in engine/src/main/java/org/pentaho/di/trans/steps/pgbulkloader/PGBulkLoader.java, you have separated truncate operation and copy command, so this code block should be deleted. Otherwise it would raise an error, because truncate operation was sent to PGCopyOutputStream(). In addition, line 86 can be deleted.

PR #4316, opened 2017-08-23
[BACKLOG-18361] - registering Get Rows From Result a second time as G…

kurtwalker:
…et Rows from Stream


kurtwalker:
closing until we get the new icon

PR #4147, opened 2017-07-12
[PDI-13227] join rows step - after changing the steps on incoming hops, step gets into undefined / not working state

Advent51:
- the extra restrictions were removed - now the field positions updates every time

PR #4324, opened 2017-08-28
[SP-3700] Backport of PDI-13227 - join rows step - after changing the steps on incoming hops, step gets into undefined / not working state (7.1 Suite)

Advent51:
- backport of #4147


Advent51:

PR #4326, opened 2017-08-28
[PDI-16422] runTrans web service does not apply transformation parameters

AliaksandrShuhayeu:
[PDI-16422] runTrans web service does not apply transformation parameters

clearing parameters was removed


mchen-len-son:
@AliaksandrShuhayeu please add SP case number to the commit message

PR #4328, opened 2017-08-28
[SP-3701] Backport of PDI-13227 - join rows step - after changing the steps on incoming hops, step gets into undefined / not working state (6.1 Suite)

Advent51:
- backport of #4147


Advent51:
@pamval @mchen-len-son @mbatchelor @stepanovdg please review

PR #4329, opened 2017-08-28
[SP-3711][PDI-14346] Error handling in Insert/Update from sub transformation

AliaksandrShuhayeu:
[SP-3711][PDI-14346] Error handling in Insert/Update from sub transformation

-added returning false in error case


AliaksandrShuhayeu:

PR #4330, opened 2017-08-28
BACKLOG-18540 SSL support for the websocket daemon

flbrino:


aminmkhan:
@flbrino looking good to me ...

PR #4339, opened 2017-08-29
[SP-3709][PDI-9388] Deleting a column from the Datagrid step moves column names as heading for wrong data

AliaksandrShuhayeu:
[SP-3709][PDI-9388] Deleting a column from the Datagrid step moves column names as heading for wrong data

added synchronization between Fields and Data tables


AliaksandrShuhayeu:

PR #4348, opened 2017-08-30
[BACKLOG-18314] update subfolders when deleting parent; parent was set to path when renaming

tgf:


tgf:
@bmorrise please review

PR #4350, opened 2017-08-30
[BACKLOG-18653] PDI: The error "Unknown scheme "nc" in URI" occurs on…

andrei-viaryshka:
… attempt to copy or delete files on HDFS


andrei-viaryshka:
@rmansoor Could you please take a look? Related PR is https://github.com/pentaho/big-data-plugin/pull/1000.

PR #4352, opened 2017-08-31
[BACKLOG-18441] - Pentaho slave server: Old configuration is used for…

YuryBY:
… running job on Pentaho Server after changing Username and saving changes


YuryBY:
@mbatchelor, @mchen-len-son, could you please take a look?

rmansoor:
@YuryBY validated this change  testing scenarios 
-  Runnning a job with a transformation.
- Runnin a job with correct carte cluster credentials and then changing it to be bad and running it again.

Both test were done successfully. Merging the PR

PR #4369, opened 2017-09-06
[PDI-16559] - Multiple step-meta classes generate AIOOB after MDI

mbatchelor:


mbatchelor:
@mchen-len-son - can you or @kolinus review this please? Thanks.

PR #4371, opened 2017-09-07
[PDI-16323] Ability to publish data source over SSL to the BA server

Advent51:
- added the use of connection properties


Advent51:
@mbatchelor @pamval @mchen-len-son @stepanovdg please review

PR #4377, opened 2017-09-08
[BACKLOG-18294] Websocket Daemon Client Session with kerberos Auth

flbrino:


flbrino:
@mkambol for validation. Thks

PR #4379, opened 2017-09-08
BACKLOG-18754 - fix conversion between storage and metadata when comp…

joaormf:
…aring two values

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
+++ b/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
@@ -23,6 +23,32 @@
 
 package org.pentaho.di.core.row.value;
 
+import org.pentaho.di.compatibility.Value;
+import org.pentaho.di.core.Const;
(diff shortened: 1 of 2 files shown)


joaormf:
In Kettle a transformation has a CSVInput step with a DATE field with the conversion mask of "MM/dd/yyyy HH:mm". That step is visualized in DET.

On DET in stream view that DATE field is used in a bar chart as the horizontal axis, and then a Keep Only filter is applied. Without this change (including the ones in [det PR 1000](https://www.github.com/pentaho/pentaho-det/pull/1000) and [det-ee PR 495](https://github.com/pentaho/pentaho-det-ee/pull/495/files) ), the query execution fails with conversion exceptions between the formats.

Notice that because of [BACKLOG-5847](http://jira.pentaho.com/browse/BACKLOG-5847) date columns in a data service always use the format YYYY-MM-dd.

joaormf:
@pentaho/kashyyyk @pentaho/millenniumfalcon @pentaho/bobafett can you please review this change.
Thanks

PR #4380, opened 2017-09-10
[BACKLOG-18592] - Search an replace all hadoop cluster url to the file system url

rmansoor:
- Only resolve hc url with a cluster information that is defined.


rmansoor:
@mkambol @tkafalas please review.

@tkafalas I had a horrible time trying to setEmbeddedNamedCluster in the transmeta for testing. Is there any other alternative.

mkambol:
@rmansoor this looks good

Couple alternatives for mocking the embeddednamedcluster in the transmeta:
1)  Use a mockito spy.  (in general spys are evil, but sometimes is the best way with legacy code).
2)  Pass the named cluster in to the convert method.

rmansoor:
@mkambol thanks for the feedback. Updated the pr with your suggestions. Please review

PR #4383, opened 2017-09-12
[BACKLOG-17671] table resize stutter in windows

tgf:


tgf:
@bmorrise please review

PR #4385, opened 2017-09-13
[BACKLOG-17959] Setting the operation and the transformation with err…

flbrino:
…or logic after receiving websocket daemon error events


flbrino:
@mkambol @aminmkhan ready for validation/merge

PR #4387, opened 2017-09-13
[BACKLOG-18316] use current dir in save as

tgf:


tgf:
@bmorrise please review

PR #4402, opened 2017-09-18
7.1 Provide Support for Apache Phoenix

xfzyrsidney:
Phoenix tables doesnt have a syntax of "insert into table " instead they have a ysntax of "upsert into table".So any data insertion/upsertion into phoenix tables through PDI fails. Rest all should be fine as phoenix provides a JDBC driver similar to many other RDMS.
Can this taken into account by the pentaho PDI team and provide support for phoenix?


mbatchelor:
Hey there,

Thanks for the contribution, but this PR has several problems and cannot be processed:
1- It's a merge commit with 216 files changed.
2- We do these kinds of things with a specific database dialect
3- You submitted a PR to the 7.1 branch instead of master.

Closing this PR.

PR #4406, opened 2017-09-18
[BACKLOG-18537] Removing the RSA daemon from the assembly.

mkambol:
Missed one.


mkambol:
@ccaspanello @flbrino @aminmkhan

PR #4408, opened 2017-09-18
[BACKLOG-18937] rename update issue

tgf:


tgf:
@ddiroma please review

PR #4409, opened 2017-09-19
[BACKLOG-18937] fix renaming reg with subfolders and files

tgf:


tgf:
@ddiroma please review

PR #4414, opened 2017-09-20
[BACKLOG-18612] Add functionality for subtrans logging within AEL/Spark

DFieldFL:
This will allow the subtrans running within AEL/Spark to bubble up to
logging to the Client just like it does in the kettle classic engine.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/ael/websocket/MessageEventService.java
+++ b/engine/src/main/java/org/pentaho/di/trans/ael/websocket/MessageEventService.java
@@ -24,6 +24,7 @@
 package org.pentaho.di.trans.ael.websocket;
 
 import org.pentaho.di.engine.api.events.PDIEvent;
+import org.pentaho.di.engine.api.model.ModelType;
 import org.pentaho.di.engine.api.remote.Message;
 import org.pentaho.di.engine.api.remote.RemoteSource;
 import org.pentaho.di.engine.api.remote.StopMessage;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/pentaho/di/trans/ael/websocket/MessageEventService.java
```diff
@@ -176,10 +177,14 @@ public int compare( Message o1, Message o2 ) {
           //others messages from the Daemon to Client are PDIEvent <RemoteSource, ?>
           RemoteSource keyRemoteSource = ( (RemoteSource) ( (PDIEvent) o1 ).getSource() );
           RemoteSource remoteSource = ( (RemoteSource) ( (PDIEvent) o2 ).getSource() );
+
+          // Note: `== ModelType.TRANSFORMATION` is to allow all transformation type logging to come through to the
+          // client specifically for AEL subtrans/subjob logs
           if ( keyRemoteSource.getModelType() == remoteSource.getModelType()
```
kurtwalker: Is there any refactoring we can do to pull these conditions into a method with a name to help understand what we're checking?

DFieldFL: Can do.

DFieldFL: I personally don't know all the use cases. I would have to discuss with @mkambol to give them all appropriate names.

PR #4418, opened 2017-09-22
[BACKLOG-18831] - Data services return Null values when filtering num…

Aliaksandr-Kastenka:
…bers

For (null < -10) condition result should be 'false'.
Check for null added to FUNC_SMALLER and FUNC_SMALLER_EQUAL, similarly to FUNC_REGEXP( https://github.com/pentaho/pentaho-kettle/blob/master/core/src/main/java/org/pentaho/di/core/Condition.java#L460 ).
Other FUNC's work correctly in case of null filed.


Aliaksandr-Kastenka:
@kurtwalker @DFieldFL 
Please review.

PR #4419, opened 2017-09-25
Automatically reconnect when connection was closed

martijnlentink:
I often encountered that steps took too long to push their rows to the table output step, where they arrive too late to prevent a timeout from the server. In our set up we aren't able to change the timeout. An automatic reconnect could help fix this when new rows arrive.


mbatchelor:
Thanks for the contribution!! Could you please create an improvement JIRA case at http://jira.pentaho.com and include this Pull Request in the description. Would love to process, but need to have JIRA case for the change.

PR #4427, opened 2017-09-26
[CLEANUP] javadoc: namedcluster saved in embedded metastore

DFieldFL:


DFieldFL:
@pentaho/ackbar @tkafalas

DFieldFL:
I can't reproduce the test error locally.

pavel-sakun:
AFAIR the test is failing like this for years

tmcsantos:
The failure happens when the system is on high load, failing any test with timeouts.

PR #4432, opened 2017-09-27
[PDI-15268] PDI locks up when changing Job - Properties - Parameters.

AliaksandrShuhayeu:
[PDI-15268] PDI locks up when changing Job - Properties - Parameters.
added calculation wrong focus event (a workaround for SWT bug)

### DIFF
--- a/ui/src/main/java/org/pentaho/di/ui/core/widget/TableView.java
+++ b/ui/src/main/java/org/pentaho/di/ui/core/widget/TableView.java
@@ -563,6 +563,10 @@ public class TableView extends Composite {
     lsFocusText = new FocusAdapter() {
       @Override
       public void focusLost( FocusEvent e ) {
+        if ( isWrongLostFocusEvent() ) {
+          return;
+        }
+
         final Display d = Display.getCurrent();
 
         if ( table.isDisposed() ) {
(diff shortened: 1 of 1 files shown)


AliaksandrShuhayeu:

ui/src/main/java/org/pentaho/di/ui/core/widget/TableView.java
```diff
@@ -626,6 +630,18 @@ public void run() {
...
+      /**
+       * This is a workaround for SWT bug (see PDI-15268).
+       * Calling a context menu should be ignored in
+       * SWT org.eclipse.swt.widgets.Control#gtk_event_after
+       *
+       * @return true if it is wrong event
+       */
+      private boolean isWrongLostFocusEvent() {
+        Control controlGotFocus = Display.getCurrent().getCursorControl();
+        return Const.isLinux() && ( controlGotFocus == null || text.equals( controlGotFocus ) );
```
mbatchelor: Should there not be a null check on the text variable ?
return Const.isLinux() && ( controlGotFocus == null || ( text != null && text.equals( controlGotFocus ) ) ;

AliaksandrShuhayeu: @mbatchelor 
It's an event for a text Widget "text". It will be despose in line 599 ftext.dispose();. Before my changes we don't have NPE in L599.

PR #4444, opened 2017-09-29
[SP-3776][PDI-14721] Problem with MySQL and BLOBs - ValueMetaInterface.TYPE_BINARY is being set to ValueMetaInterface.STRING

AliaksandrShuhayeu:
[SP-3776][PDI-14721] Problem with MySQL and BLOBs - ValueMetaInterface.TYPE_BINARY is being set to ValueMetaInterface.STRING

-there was deleted setting wrong type


AliaksandrShuhayeu:

PR #4451, opened 2017-09-29
[BACKLOG-18969] Transformations no longer run on carte after 7.1 upgrade

andrei-viaryshka:
Changed to create run configuration with remote server info for cases when Run configuration does not exist for Transformation job entry but slave_server_name is specified.


andrei-viaryshka:
@rmansoor Could you take a look please?

PR #4465, opened 2017-10-02
[BACKLOG-18766] current directory changes in job entries, use auth context w register package

tgf:


tgf:
@bmorrise please review

PR #4485, opened 2017-10-09
[SP-3781][PDI-16559] - Multiple step-meta classes generate AIOOB afte…

mchen-len-son:
…r MDI


mchen-len-son:
@stepanovdg please review and merge
thanks

PR #4487, opened 2017-10-10
[BACKLOG-18677] Kettle plugin classloader support for multiple plugin…

tmcsantos:
… jars

@pentaho/2-1b pls review


tmcsantos:
You're probably right about logging something on those empty catch blocks, at this point I just trying to maintain previous "log free" behaviour.

PR #4496, opened 2017-10-11
[BACKLOG-18845] Remove the need for VERSION_FOR_LICENSE replacement

Kyrrui:


plugins/file-open-save/core/src/main/java/org/pentaho/repo/dialog/RepositoryOpenSaveDialog.java
```diff
@@ -94,6 +98,19 @@ public void open( String directory, String state, String filter, String origin )
     }
   }
 
+  private static String getClientPath(){
+    Properties properties = new Properties();
+    try {
+      InputStream inputStream =
+        FileHelper.class.getClassLoader().getResourceAsStream("project.properties");
```
hbfernandes: What's this FileHelper? I'd say use the current class to get the classloader from.

plugins/file-open-save/core/src/main/resources/project.properties
```diff
@@ -0,0 +1,2 @@
+PROJECT_VERSION=8.0-SNAPSHOT
```
hbfernandes: Given the single usage this has, i'd add a single property with the full client path and fetch it in one go:

CLIENT_PATH=/${project.artifactId}/${project.version}/index.html

Then add this file to src/main/resources-filtered so that it gets dynamically replaced.

plugins/file-open-save/core/src/main/project.properties
```diff
@@ -0,0 +1 @@
+CLIENT_PATH=/8.0-SNAPSHOT/file-open-save/index.html
```
tmcsantos: duplicated?

plugins/file-open-save/core/src/main/resources/project.properties
```diff
@@ -0,0 +1 @@
+CLIENT_PATH=/8.0-SNAPSHOT/file-open-save/index.html
```
tmcsantos: the path is wrong, and this must be filtered. move to `resources-filtered`
`CLIENT_PATH=/${project.artifactId}/${project.version}/index.html`

PR #4500, opened 2017-10-12
[PDI-16501]-Update step: The row in "The key(s) to look up the value(s)" is missing after exporting the repository.

ArtsiomYurhevich:
Fixed


ArtsiomYurhevich:
@bmorrise, @mbatchelor, @pamval Could you please take a look?

PR #4502, opened 2017-10-12
[PDI-16480] Salesforce: Datetime fields are not being set with Pentaho 7.1 version

AliaksandrShuhayeu:
[PDI-16480] Salesforce: Datetime fields are not being set with Pentaho 7.1 version

-added saving hh:mm:ss for saving datetime in Salesforce
-created one method with one rules for normalizing data for all Salesforce modify steps

### DIFF
--- a/plugins/salesforce/core/src/main/java/org/pentaho/di/trans/steps/salesforce/SalesforceStep.java
+++ b/plugins/salesforce/core/src/main/java/org/pentaho/di/trans/steps/salesforce/SalesforceStep.java
@@ -22,9 +22,12 @@
 
 package org.pentaho.di.trans.steps.salesforce;
 
+import com.google.common.primitives.Ints;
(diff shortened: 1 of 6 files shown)


AliaksandrShuhayeu:
Now we use hh:mm:ss when sending date. DateTime is sent correct in salesforce for salesforce date and salesforce date time fields. Salesforce Input step works correct before changes. It displays date and datetime as Hitachi Vantara date type with time format. 
These changes fixes Salesforce Insert, Update and Upset steps. 
These changes affect PDI-16434 , PDI-10836 . They should work correctly for all Salesforce modify steps: Insert, Update and Upset. 
Before it we didn't have fixing PDI-16434 for Salesforce Update step and we didn't have fixing PDI-10836 for Salesforce Update and Upset steps.

PR #4504, opened 2017-10-13
[PDI-16731] TextOutput for empty not null values are not enclosed.

dshvedchenko:
When field length is not specified at Fields Tab, but quote is enforced -> empty string  is left unquoted, so it is undistiguishable from NULL.

Similar code I suspect need to be propogated to S3 Text Output plugin

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/textfileoutput/TextFileOutput.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/textfileoutput/TextFileOutput.java
@@ -430,7 +430,7 @@ public class TextFileOutput extends BaseStep implements StepInterface {
         }
       }
 
-      if ( str != null && str.length > 0 ) {
+      if ( str != null && (str.length > 0 || meta.isEnclosureForced() ) ) {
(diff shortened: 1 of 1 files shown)


mbatchelor:
@dshvedchenko - Would love to merge this but we need a couple things:
1- A Jira defect filed at http://jira.pentaho.com
2- A test case demonstrating the issue is fixed so it doesn't regress.

dshvedchenko:
Will try to collect and prepare all info asap

jbleuel:
We need to solve this as an improvement and with a new UI element in the Text File Output step.

mbatchelor:
@dshvedchenko - thanks again for the contribution and the JIRA case. I unfortunately have to close this PR because the only way to process a change in behavior like this is with a UI update allowing people to specify that they want null strings enclosed in quotes. So this requires the following bits:
1- A UI change that as you suggest, also has to flow to the HBaseOutput/S3Output/etc (all subclasses of TextFileOutput)
2- A unit test to make sure that the capability doesn't regress inadvertently.

PR #4510, opened 2017-10-17
[PDI-16010] In Excel Input Step, trimming of space doesn`t happen whe…

mchen-len-son:
…n Spoon is opened in Japanese preferred language.

 - translated trim type strings in jp resource file to fix mapping between ui values set and trim type code inside the step


mchen-len-son:
@mbatchelor can you please take a look.
I've comented in the case about the changes.
Translations a copied from the similar resource file in ui module.

PR #4514, opened 2017-10-17
[BACKLOG-19342] make FastJsonReader fast again

tgf:


tgf:
@bmorrise please review

PR #4517, opened 2017-10-18
[PSW-229] Script files don't have an header with Copyright data

Advent51:
- added license to the command file


Advent51:
@mbatchelor @pamval @mchen-len-son @stepanovdg please review

PR #4532, opened 2017-10-24
[PDI-16535]-Errors creating transformations and jobs when the user connects to Pentaho Repository using a different case of the user name created

denisprotopopov:
- fixed. Now user searching in the list of users


denisprotopopov:
@bmorrise please review the changes

PR #4534, opened 2017-10-24
[BACKLOG-19496] allow canceling perspective activation

tgf:


tgf:
@bmorrise please review

PR #4537, opened 2017-10-25
[PDI-14615] - Invalid error message

mbatchelor:


mbatchelor:
IntelliJ was replacing my imports with .* after I saved my changes. I've fixed the issue in IntelliJ and squashed the fix.

PR #4541, opened 2017-10-26
[BACKLOG-18996] PDI - Kafka Producer Step - Running a transformation with the Kafka Producer step in Spark or AEL produces a plugin class error

pavel-sakun:
Throw exception in case ClassLoadingPlugin doesn't have requested class registered.
@pentaho/ackbar @mkambol @DFieldFL


core/src/main/java/org/pentaho/di/core/plugins/PluginRegistry.java
```diff
@@ -463,7 +463,14 @@ private String createSupplemantalKey( String pluginName, String id ) {
     }
 
     if ( plugin instanceof ClassLoadingPluginInterface ) {
-      return ( (ClassLoadingPluginInterface) plugin ).loadClass( pluginClass );
+      T aClass = ( (ClassLoadingPluginInterface) plugin ).loadClass( pluginClass );
+      if ( aClass == null ) {
+        throw new KettlePluginClassMapException( BaseMessages
+            .getString( PKG, "PluginRegistry.RuntimeError.NoValidClassRequested.PLUGINREGISTRY002", plugin.getName(),
+                pluginClass.getName() ) );
+      } else {
```
DFieldFL: else is not needed here.

pavel-sakun: Can't agree. Else branch is how it worked before (return plugin.loadClass), without it we're not getting services plugin provides.

DFieldFL: Sorry didn't see the code below. Thanks 👍

DFieldFL:
Looks like I can't merge...PluginRegistryUnitTest.java has a conflict. Would yo be able to fix this?

pavel-sakun:
@DFieldFL done.

PR #4551, opened 2017-10-30
[SP-3833] Backport of PDI-15803 - Need to update the tooltip "Run on …

mchen-len-son:
…a Data Integration (DI) Server" in the dialog Run Options

 - fixed tooltips


mchen-len-son:
@stepanovdg please review

PR #4553, opened 2017-10-30
[SP-3827] Backport of PDI-16010 - In Excel Input Step, trimming of sp…

mchen-len-son:
…ace doesn`t happen when Spoon is opened in Japanese preferred language

 - translated trim type strings in jp resource file to fix mapping between ui values set and trim type code inside the step


mchen-len-son:
@stepanovdg please review

PR #4554, opened 2017-10-30
[SP-3812] Backport of PDI-14492 - Copy Table Wizard_UI Issue

mchen-len-son:
- The page.createControl( shell ) will be called in Wizard.createPageControls(Composite), so should not be called in Spoon.copyTableWizard()


mchen-len-son:
@stepanovdg please review

PR #4556, opened 2017-10-30
[SP-3806][PDI-16434]Salesforce Upsert (and possibly insert) fails when upserting an integer into account object

AliaksandrShuhayeu:
[SP-3806][PDI-16434]Salesforce Upsert (and possibly insert) fails when upserting an integer into account object

-fixing sending Pentaho Integer value to Salesforce (SalesforceInsert and SalesforceUpsert steps)


AliaksandrShuhayeu:

mchen-len-son:
Merging.

PR #4557, opened 2017-10-30
[SP-3824][PDI-15268] PDI locks up when changing Job - Properties - Parameters.

AliaksandrShuhayeu:
[SP-3824][PDI-15268] PDI locks up when changing Job - Properties - Parameters.
added calculation wrong focus event (a workaround for SWT bug)


AliaksandrShuhayeu:

mchen-len-son:
Merging.

PR #4559, opened 2017-10-30
[SP-3818][PDI-16459] Salesforce Insert - sObject type 'string' is not supported

AliaksandrShuhayeu:
[SP-3818][PDI-16459] Salesforce Insert - sObject type 'string' is not supported

removed wrong using RelationshipName instead of Name


AliaksandrShuhayeu:

mchen-len-son:
Merging.

PR #4570, opened 2017-10-31
[SP-3817]-Backport of PDI-16501 - Update step: The row in "The key(s)…

ArtsiomYurhevich:
… to look up the value(s)" is missing after exporting the repository


ArtsiomYurhevich:
@stepanovdg, @mchen-len-son

PR #4575, opened 2017-11-01
[TEST-cleanup]-cleanup test org.pentaho.di.core.logging.LogChannelFileWriterTest#testStartStopLogging

ArtsiomYurhevich:


mchen-len-son:
The test was always passing, having the fix from #4428 applied or not.
This change is making it dependent on the actual conditions it's was created for.

PR #4580, opened 2017-11-01
[BACKLOG-19744] Remove the source fields from XML Join preview generation

DFieldFL:


DFieldFL:
@pentaho/ackbar please review

mbatchelor:
This looks really close to the "fix" that was originally done which caused a performance regression on PDI-16721. transMeta.getStepFields calls the step's getFields method. Can you validate that PDI-16721 is not reintroduced with that fix ?

DFieldFL:
will do

PR #4583, opened 2017-11-02
[SP-3837]-Backport of PDI-15891 - Problem running a zipped file with Pan.bat

ArtsiomYurhevich:


mchen-len-son:
Merging.

PR #4584, opened 2017-11-02
[SP-3837]-ETL Metadata Injection step fails doesn't find a subtransfo…

ArtsiomYurhevich:
…rmation after importing an exported zip(using export to linked resources)


mchen-len-son:
This is strange. Rows wingman is referring to as a checkstyle violating are good and have correct indentations.
Merging.

PR #4590, opened 2017-11-04
[PPP-3888] - Use of vulnerable component com.fasterxml.jackson.core :…

YuryBY:
… jackson-databind-2.5.4.jar ,com.fasterxml.jackson.core : jackson-databind-2.3.3.jar,com.fasterxml.jackson.core : jackson-databind-2.5.3.jar

### DIFF
--- a/plugins/pom.xml
+++ b/plugins/pom.xml
@@ -20,6 +20,9 @@
     <!-- Test running configuration -->
     <maven-surefire-plugin.reuseForks>false</maven-surefire-plugin.reuseForks>
     <maven-failsafe-plugin.reuseForks>false</maven-failsafe-plugin.reuseForks>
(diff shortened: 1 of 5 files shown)


plugins/pom.xml
```diff
@@ -20,6 +20,9 @@
     <!-- Test running configuration -->
     <maven-surefire-plugin.reuseForks>false</maven-surefire-plugin.reuseForks>
     <maven-failsafe-plugin.reuseForks>false</maven-failsafe-plugin.reuseForks>
+    <!-- jackson-databind depends on jackson-core and jackson-annotations
+	so that this artifacts have the same version to prevent compile errors -->
+    <jackson.version>2.9.2</jackson.version>
```
pamval: Why is this property named differently?

PR #4596, opened 2017-11-07
[PDI-16590] - Replace in String: Step processes regular expression even when "Use Regex" option set to N

AndreyBurikhin:
@bmorrise , @mbatchelor, @mchen-len-son  please review.


mchen-len-son:
@bmorrise , @mbatchelor, @pamval please review.

PR #4600, opened 2017-11-09
[PPP-3876]-Use of vulnerable component spring-security-core 4.1.9 cve-2017-4995

stepanovdg:


stepanovdg:
@bmorrise @mbatchelor @pamval Could you please take a look?

PR #4603, opened 2017-11-11
[PDI-8934]-BaseStep rowListeners - Thread Safety Issue. Removed synch…

ArtsiomYurhevich:
…ronized blocks and changed implementation of step listeners to CopOnWriteArrayList

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/ael/adapters/StepInterfaceEngineAdapter.java
+++ b/engine/src/main/java/org/pentaho/di/trans/ael/adapters/StepInterfaceEngineAdapter.java
@@ -118,13 +118,9 @@ public class StepInterfaceEngineAdapter extends BaseStep {
         // no way to convert this row to a RowMetaInterface.
         return;
       }
-      synchronized ( rowListeners ) {
-        for ( int i = 0; i < rowListeners.size(); i++ ) {
-          RowListener rowListener = rowListeners.get( i );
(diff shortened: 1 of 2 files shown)


ArtsiomYurhevich:
@mbatchelor , @pamval Could you please take a look?

pamval:
Maybe there was a discussion about this case that I missed, but the changes here do not follow the suggestions by @mbatchelor  on the jira case.

Not particularly happy about using CopyOnWriteArrayList since it's not clear to me how many writes we have on that list compared to reads. Who writes to that list?

ArtsiomYurhevich:
@pamval Adding row listeners is usually performed only during initialization/preparing stage of  trans/job execution and then during execution stage we only read from the collection. So, I think, CopyOnWriteArrayList is appropriate for this purposes

ArtsiomYurhevich:
@pamval The changes do not exactly follow the description of the issue because the issue was created long ago in 2012

pamval:
@mbatchelor please review this change.

mbatchelor:
I searched the code for mutations of the row listeners list (who calls either addRowListener or removeRowListener) and as far as I can see, the CopyOnWriteArrayList should not be a performance burden on normal execution. It does run the risk of making tests run slower though because we have a *huge* amount of unit and integration tests that rely on the row listeners for validating results. That said, any case we avoid any kind of synchronization without impacting usual runtime performance is a win.

PR #4615, opened 2017-11-15
[PDI-14296] Load File Content in Memory corrupts data read from binary files

Advent51:
- LoadFileInput step was modified in order to work with file content as a binary array, not as a String
- filecontent field in the LoadFileInputData became byte[]


Advent51:
@mbatchelor @pamval @mchen-len-son @bmorrise please review

PR #4617, opened 2017-11-15
[PDI-16580] - Webservices auth failed when using parameters for usern…

pentaho-lmartins:
…ame and password.

@duarteteixeira


pentaho-lmartins:
@pamval

PR #4621, opened 2017-11-15
[PDI-16783] Transformations are no longer run on Carte server after u…

mchen-len-son:
…pgrade from 6.1 executed by scheduled jobs

 - added job opening extensions point call


mchen-len-son:
@mbatchelor @pamval can you please take a look?
this was already tested on 7.1

PR #4626, opened 2017-11-16
[PDI-16589] - Running in parallel would introduce additional rows and…

pentaho-lmartins:
…/or invalid data in the output.

@duarteteixeira @pamval


pamval:
@bmorrise @mbatchelor @AlexanderBuloichik  @tgf please review

PR #4649, opened 2017-11-28
[PDI-16818] - Fix for run window taking ages to appear

mbatchelor:


mbatchelor:
I fixed the test case and updated the PR.

PR #4664, opened 2017-12-01
[PDI-16066] Transformation Executor Step does not Inherit Parameters

AliaksandrShuhayeu:
-fixed a broken option "Inherit all variables from the transformation"
-extracted a common behavior for activation params in one methods for children of StepWithMappingMeta
-overwriting parameters and the next logic
When the child parameter does not exist in the parent parameters, keep it.
When the child parameter does exist in the parent parameters, overwrite the child parameter by the parent parameter.
All other parent parameters need to get copied into the child parameters  (when the 'Inherit all variables from the transformation?' option is checked)

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/transexecutor/TransExecutor.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/transexecutor/TransExecutor.java
@@ -38,6 +38,7 @@ import org.pentaho.di.core.row.RowDataUtil;
 import org.pentaho.di.core.row.RowMeta;
 import org.pentaho.di.i18n.BaseMessages;
 import org.pentaho.di.job.DelegationListener;
+import org.pentaho.di.trans.StepWithMappingMeta;
 import org.pentaho.di.trans.Trans;
 import org.pentaho.di.trans.TransExecutionConfiguration;
 import org.pentaho.di.trans.TransMeta;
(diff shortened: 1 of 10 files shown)


mchen-len-son:
@mbatchelor @bmorrise can you please take a look?
thanks.

AliaksandrShuhayeu:
@bmorrise  Could you please review and merge it?
thanks.

engine/src/main/java/org/pentaho/di/trans/steps/transexecutor/TransExecutor.java
```diff
@@ -268,43 +265,16 @@ Trans createInternalTrans() throws KettleException {
...
-    String[] parameterNames = internalTrans.listParameters();
-    for ( int i = 0; i < parameters.getVariable().length; i++ ) {
-      String variable = parameters.getVariable()[ i ];
-      String fieldName = parameters.getField()[ i ];
-      String inputValue = parameters.getInput()[ i ];
-
-      String value;
-      // Take the value from an input row or from a static value?
-      if ( !Utils.isEmpty( fieldName ) ) {
-        int idx = getInputRowMeta().indexOfValue( fieldName );
```
roboguy: @AliaksandrShuhayeu - could you have a look at PDI-17045, I belive this code change has caused some change in how this step functions. I would appreciate your input. Thank you.

AliaksandrShuhayeu: @roboguy I answered you in PDI-17045. A problem with the Fields to replace existed  before this changes.

sudhanshu-Tango: Thank you @AliaksandrShuhayeu for the quick reply and taking a look at the ticket! 
As you mentioned in point 2:

"The UI does not populate the Fields to replace (see attached screenshots)"
This problem was not fixed in scope of PDI-16066. And this problem existed before a correction PDI-16066. 

Is there an open issue that deals with the UI problem ? If not would you be able to create it ?

Thank you again for taking a look at this.

PR #4680, opened 2017-12-04
[BACKLOG-19677] - getting substatus through same interface as kettle.…

kurtwalker:
…  sub-steps will always have same status as parent


kurtwalker:
@pentaho/ackbar

PR #4684, opened 2017-12-05
[BACKLOG-19589] let admin delete others' trash

tgf:


tgf:
@bmorrise please review

PR #4685, opened 2017-12-06
[BACKLOG-20155] - creates StepDialog annotation so we can have a clea…

tmcsantos:
…n way to separate plugins UI from implementation logic.


tmcsantos:
@pentaho/2-1b pls review

I'm waiting for wingman to finish the tests first, to then request a full stack build.

PR #4686, opened 2017-12-06
[PDI-16591] Able to import Job/Trans to /<root>

AliaksandrShuhayeu:
[PDI-16591] Able to import Job/Trans to /<root>
don't allow to import a file to /<root> of the pentaho repository


AliaksandrShuhayeu:

AliaksandrShuhayeu:

agJCraig:
Unfortunately, this approach to preventing an ETL job/transformation from being loaded into the / <root> directory also prevents _any_ XML file exported from the PDI Client (Spoon) from being imported by the PDI Import utility because, in an exported XML file, the <directory> tag values are always full paths (such as /public/MyETL/Job1). To import such a job, you have to specify / as the target directory.

To solve the underlying issue without rendering the Import utility incapable of importing XML files exported by Spoon, the checking needs to occur at the point where the <directory> tag is evaluated. That is, a fix that preserves the utility of the Import tool and still prevents a job/trans from being imported into the root directory needs to examine the <directory> tag values and reject a job/trans having its directory tag's value as / only _when the target directory on the command line is root_.

PR #4690, opened 2017-12-08
[BACKLOG-17271] First pass at a generic api and base impl for streaming

mkambol:
This commit includes
1) the StreamSource and StreamWindow interfaces, which
define contracts for receiving data and processing it within windowed
chunks.
2) Common base classes for stream step, meta, and dialogs
3) A Fixed time / batch size window implementation
4) A sample "tail -f" style step using all of the above.
The "plugins/file-stream" sample is not a child module only intended to
be built manually.


kurtwalker:
license headers fixed.  no need to wait for wingman to finish a second time.

PR #4694, opened 2017-12-12
[BACKLOG-19589] confirmation dialog appearance

tgf:


tgf:
@bmorrise please review

luckyxiang2017:
HTTP encode does not work，who can solve it

PR #4697, opened 2017-12-14
[BACKLOG-20407] - Replace cache version creation

plagoa:
- added a getCacheVersion method to the TransMeta class responsible for calculating the cache version considering only the relevant transformation properties


plagoa:
@marcoslarsen @mkambol @kurtwalker please review

related to:
https://github.com/pentaho/pdi-dataservice-server-plugin/pull/272

PR #4704, opened 2017-12-15
[PDI-16819]-Copying a Timestamp to result cannot be de-serialized whe…

ArtsiomYurhevich:
…n running on Slave Server


ArtsiomYurhevich:
@pamval , @mbatchelor Could you please take a look?

PR #4706, opened 2017-12-18
[PDI-16884] - Cassandra Output step does not work with apache-cassand…

YuryBY:
…ra-2.1.0

@mbatchelor, could you please take a look?

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/plugins/KettleURLClassLoader.java
+++ b/core/src/main/java/org/pentaho/di/core/plugins/KettleURLClassLoader.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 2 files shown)


luckyxiang2017:
HTTPPOST，line 286，please add encode
if ( entity != null ) {
                            String encode = data.realEncoding;
                            if (StringUtil.isEmpty(encode)) {
                                ContentType contentType = ContentType.get(entity);
                                encode = contentType == null ? "UTF-8" : contentType.getCharset().name();
                            }

                            body = new String(EntityUtils.toByteArray(entity), encode);
                        } else {
                            body = "";
                        }

luckyxiang2017:
You upgrade and change the original function，it's bad

core/src/main/java/org/pentaho/di/core/plugins/KettleURLClassLoader.java
```diff
@@ -104,7 +104,7 @@ public String getName() {
   protected synchronized Class<?> loadClass( String arg0, boolean arg1 ) throws ClassNotFoundException {
     try {
       return loadClassFromThisLoader( arg0, arg1 );
-    } catch ( ClassNotFoundException | NoClassDefFoundError e ) {
+    } catch ( ClassNotFoundException | NoClassDefFoundError | SecurityException e ) {
```
mbatchelor: Would prefer logging here - need to log the exception somewhere. Swallowing it is not a good fix.

YuryBY:
@mbatchelor, PR have updated

PR #4710, opened 2017-12-19
[BACKLOG-20516][8.0]-YARN: There is no output when running YARN job using "Clustered" location in Run configuration

stepanovdg:


stepanovdg:
@bmorrise @mchen-len-son Could you please take a look?

PR #4711, opened 2017-12-19
[BACKLOG-20516]-YARN: There is no output when running YARN job using "Clustered" location in Run configuration

stepanovdg:


stepanovdg:
@bmorrise  @duarteteixeira Could you please take a look?

PR #4713, opened 2017-12-19
[PDI-16660] Metadata Injection Sample is broken

vasilii-komarov:
- mapping has been fixed in use_metainject_step.ktr


vasilii-komarov:
@bmorrise  @pamval  @mchen-len-son  @mdamour1976  please review

PR #4714, opened 2017-12-20
[PDI-16707] Date format is not retained after performing Memory Group…

pentaho-lmartins:
… by.

The previous PR introduced a bug in both Group by and Memory Group by steps when the aggregation doesn't have a subject field, For example, a count(*) instead of a max(date).

@duarteteixeira @pamval @mbatchelor


kolinus:
@pentaho-lmartins  could you please add unit tests demonstrating the issue and demonstrating your fix works?

pentaho-lmartins:
Will provide unit tests shortly.

pentaho-lmartins:
Included unit tests to the changed code.
Squashed previous commit with new commit that includes unit tests.

@duarteteixeira @pamval @mbatchelor @kolinus

kolinus:
@pentaho-lmartins , look good to me
thank you.

PR #4717, opened 2017-12-21
[PDI-16596]-When I write a transformation from an injection, paramete…

ArtsiomYurhevich:
…rs from template is gone


ArtsiomYurhevich:
@mbatchelor , @pamval , @bmorrise Could you please take a look?

PR #4718, opened 2017-12-21
[BACKLOG-20210] - supporting injected lists for saving xml in stream …

kurtwalker:
…steps


kurtwalker:
I think I screwed up wingman with too many for pushes.  closing and re-opening

PR #4720, opened 2017-12-22
[PDI-16697] - Error reading file in Text File Input step causes trans formation to fail when "Skip error files?" and error handling is enabled

AndreyBurikhin:
@bmorrise, @mchen-len-son please review


kolinus:
@bmorrise @mbatchelor could you please take a look

PR #4721, opened 2017-12-22
[PDI-16809|PDI-16812|PDI-16824] Add a checksum - generates wrong checksum

Advent51:
- added new logic for the binary data checksum generation
- added new compatibility flag "Old Checksum Behaviour Mode"
- added new checkbox on the "Add Checksum" step UI
- regression with the wrong field checksum calculation was fixed
- new tests were created


Advent51:
@mchen-len-son @pamval @mbatchelor @bmorrise please review

PR #4726, opened 2017-12-26
[SP-3966][PDI-16591] Able to import Job/Trans to /<root>

AliaksandrShuhayeu:


mchen-len-son:
@AliaksandrShuhayeu same here. +checkstyles

PR #4736, opened 2017-12-28
[SP-3945] - Backport of [PDI-16726] Cannot connect to MSSQL with curl…

vasilii-komarov:
…y brace in password


vasilii-komarov:
@mchen-len-son @stepanovdg please review

PR #4744, opened 2017-12-29
[BACKLOG-20601] - moving RecordsFromStream to main kettle because it …

kurtwalker:
…is now needed by the base streaming dialog


kurtwalker:
@pentaho/ackbar please review

PR #4749, opened 2018-01-03
French characters and typo issues fix

bcolas:


mbatchelor:
@bcolas  - Thank-you for the contribution, but it's impossible for us to accept this as it's a merge-commit (there are 58 commits in here). So, the way forward is as follows:
1- Re-base your fork
2- Re-apply your fixes
3- Submit a PR that has *only one commit* with your contribution!

Thanks again,

Marc

PR #4762, opened 2018-01-04
[SP-4001][PDI-14296] Load File Content in Memory corrupts data read from binary files

stepanovdg:
- LoadFileInput step was modified in order to work with file content as a binary array, not as a String
- filecontent field in the LoadFileInputData became byte[]
- tests were added


stepanovdg:
@mchen-len-son @kolinus Could you please take a look?

PR #4766, opened 2018-01-04
[BACKLOG-20601] UX edits for creating sub trans from streaming step

DFieldFL:
Also reused the spoon file save so their is less dup code.


DFieldFL:
@pentaho/ackbar please review

PR #4769, opened 2018-01-08
[BACKLOG-20617] - remove call to StopAll.  returning false will stop …

kurtwalker:
…the trans


kurtwalker:
@pentaho/ackbar this one should fix the mqtt stop error

kurtwalker:
@pentaho/ackbar added a second commit for @DFieldFL save issue

PR #4774, opened 2018-01-09
[PDI-16747]-Spoon crashes with StackOverflow if job has loop of steps

ArtsiomYurhevich:


ArtsiomYurhevich:
@mbatchelor , @pamval , @bmorrise Could you please take a look?

pamval:
@mdamour1976 can you take a look at this one since you wrote the original code?

PR #4775, opened 2018-01-09
[BACKLOG-20210] - do not put more rows when trans is stopped

kurtwalker:


kurtwalker:
@pentaho/ackbar now we start the long wait for wingman

PR #4779, opened 2018-01-10
[PDI-16527] - XML Input Stream step re-writes data in fields from previous  steps

stepanovdg:
What was done:
1) changed unit test
2) added shift for initial fields


stepanovdg:
@bmorrise Could you please take a look?

PR #4784, opened 2018-01-11
Create FileSpool.java

LordNycon:
helper to save ktr/kjb as PNG

### DIFF
--- a/ui/src/main/java/org/pentaho/di/ui/core/FileSpool.java
+++ b/ui/src/main/java/org/pentaho/di/ui/core/FileSpool.java
@@ -1,0 +1,40 @@
+package org.pentaho.di.ui.core;
+
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.widgets.FileDialog;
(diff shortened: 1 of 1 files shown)


LordNycon:
the requests #4785 , #4784 and #4783 have to be considered as one request

mbatchelor:
Hey LordNycon,

Thank-you for the contribution, but I'm afraid that we cannot accept the contribution as it currently exists. To get your changes included, please do the following:
1- Please open an improvement JIRA case at https://jira.pentaho.com . Please describe the use-case for this feature/improvement.
2- Please create a unit test that demonstrates that the new improvement works as desired. This will also keep the feature working as we run all the unit tests when we build PDI.
3- Please create *a single* pull request with all four file changes (including your new unit test)/adds that references the JIRA case in the commit message. It should look something like: [PDI-xxxxx] - Contributing ability to export the PDI canvas as a PNG.
4- Then, please update the improvement JIRA case you created with the PR.

Thanks!

PR #4788, opened 2018-01-11
[BACKLOG-20822] Set default filename for repository save dialog

DFieldFL:


DFieldFL:
@pentaho/ackbar please review

plugins/file-open-save/core/src/main/java/org/pentaho/repo/dialog/RepositoryOpenSaveDialog.java
```diff
@@ -58,7 +58,8 @@ public RepositoryOpenSaveDialog( Shell shell, int width, int height ) {
     super( shell, width, height );
   }
 
-  public void open( Repository repository, String directory, String state, String title, String filter, String origin ) {
+  public void open( Repository repository, String directory, String state, String title, String filter, String origin,
```
kurtwalker: Is it possible adding a parameter here will break other projects that might be using this?  Should we keep the original method and overload with this new one?

DFieldFL: I only saw it used in one place (currently in this commit), but you're right, lets be safe and overload it.

PR #4790, opened 2018-01-12
[SP-4001] Backport of PDI-14296 - Load File Content in Memory corrupts data read from binary files (7.1 Suite)

Advent51:
- fix for the found regression


Advent51:

PR #4791, opened 2018-01-12
[SP-4002] Backport of PDI-14296 - Load File Content in Memory corrupts data read from binary files (8.0 Suite)

Advent51:
- fix for the found regression


Advent51:

PR #4794, opened 2018-01-15
[PDI-16803]-Error in Microsoft Excel Writer when using xlsx file as t…

ArtsiomYurhevich:
…emplate


ArtsiomYurhevich:
@mbatchelor, @pamval Could you please take a look?

mchen-len-son:

PR #4799, opened 2018-01-16
[PDI-16805] Salesforce Input doesn't get the transformation status 'changed'

AliaksandrShuhayeu:
[PDI-16805] Salesforce Input doesn't get the transformation status 'changed'
- added status 'changed' when button 'Get fields' is pressed


AliaksandrShuhayeu:

PR #4801, opened 2018-01-16
[BACKLOG-20816] - SubtransExecutor should not process any more rows w…

kurtwalker:
…hen stopped


kurtwalker:
@pentaho/ackbar please review.  note I pushed the copyright change, no need to wait for wingman again

PR #4804, opened 2018-01-17
[PDI-16747]-Spoon crashes with StackOverflow if job has loop of steps

ArtsiomYurhevich:


ArtsiomYurhevich:
@mbatchelor, @pamval Could you please take a look? The pr applies the same logic as we have in jobs for transformations

mchen-len-son:

PR #4806, opened 2018-01-17
[PDI-16798]-Get Files Rows Count - changing checkboxes does not promp…

ArtsiomYurhevich:
…t for save


mchen-len-son:

PR #4807, opened 2018-01-17
[BACKLOG-20816] - empty result should not throw an error

kurtwalker:


kurtwalker:
@pentaho/ackbar please review

PR #4815, opened 2018-01-18
[BACKLOG-20859] Run MQTT Step in AEL/Spark

DFieldFL:
Move retrieving the Resource Dependencies to Base Stream Meta.
Put libraries in the appropriate spots for AEL/Spark.


DFieldFL:
@pentaho/ackbar please review

kurtwalker:
why are the mqtt libs are needed in kettle?

mkambol:
@DFieldFL Can these dependencies be loaded in OSGi rather than putting them in lib?  Looks like mqttv3 is a bundle.

DFieldFL:
Ran into the same problem as Kafka where spark can't see the libs. I'd be willing to discuss this further.

PR #4816, opened 2018-01-19
[BACKLOG-18455] Removing deleted files from Recents...

pentaho-amessier:
…and gracefully handling the opening of a Recent file that no longer exists

@pentaho/r2d2

### DIFF
--- a/ui/src/main/java/org/pentaho/di/ui/core/PropsUI.java
+++ b/ui/src/main/java/org/pentaho/di/ui/core/PropsUI.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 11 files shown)


pentaho-mracine:
Did IntelliJ automatically refactor a lot of the imports? It looks like some are moved or condensed - not a super big deal just adds more to the PR.

ui/src/main/java/org/pentaho/di/ui/core/PropsUI.java
```diff
@@ -1285,4 +1291,104 @@ public void setRecentSearches( String recentSearches ) {
...
+  private static void removeRecentImpl(
+    final String repoName, final String repoDirectory, final String repoFileName, final List<LastUsedFile> files ) {
+
+    if ( files != null ) {
+      for ( LastUsedFile lastUsedRepoFile : files ) {
+        if ( lastUsedRepoFile.getDirectory().equals( repoDirectory )
+          && lastUsedRepoFile.getFilename().equals( repoFileName )
+          && lastUsedRepoFile.getRepositoryName().equals( repoName ) ) {
+          files.remove( lastUsedRepoFile );
+          Spoon.getInstance().addMenuLast();
```
bmorrise: This is the only line that I'm not sure about. Can we move the updating of the UI to right after we call removeRecent in the various spots. I know it would be some code duplication, it just feels strange to update the UI inside a file that should just be updating the ui properties file.

pentaho-amessier: I had the same reservations - I can certainly move this line out

pentaho-amessier:
@pentaho-mracine yes, this was automatic - I agree it's a bit distracting, but to work around it seems counter-productive

pentaho-amessier:
Closing for now, as we try to determine the scope of this work

PR #4820, opened 2018-01-20
[PDI-16813]: Request Error: Invalid class name: org.jets3t.service.utils.RestUtils$ConnManagerFactory when connecting to Amazon S3 bucket in PDI 8.0

TatsianaKasiankova:
There are fix, unit tests.

### DIFF
--- a/plugins/s3csvinput/core/pom.xml
+++ b/plugins/s3csvinput/core/pom.xml
@@ -60,5 +60,11 @@
       <artifactId>powermock-api-mockito</artifactId>
       <scope>test</scope>
     </dependency>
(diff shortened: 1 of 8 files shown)


TatsianaKasiankova:
The reason of the issue:
Httpclient **v4.5.3**+loads Connection manager class using `Thread.currentThread().getContextClassLoader()` (which is ApplicationClassloader). Previously Httpclient just used `Class.forName(ClassName)` which used Class loader of the calling class.

TatsianaKasiankova:
@mbatchelor, @bmorrise, @pamval  please review the PR.

PR #4829, opened 2018-01-23
[PDI-16969] - Fix for step fields cache not being cleared.

mbatchelor:


mbatchelor:
Re-pushed fixes - Artsiom's fix for RowMeta was not in the branch so his test case broke. Changing the constructor that receives a node to the more recent version from Artsiom fixes the RowMeta test case.

mbatchelor:
That test does not fail locally. Re-submitting PR based on a rebase.

PR #4830, opened 2018-01-23
[BACKLOG-21028] Metadata Injection: Table Input step needs to support the Connection field

e-cuellar:


e-cuellar:
@bmorrise Please Sir!

mdamour1976:
@bmorrise I think this looks good, double-check that the "old style injection" field names match the annotation field names.  Also, when I looked at the resource bundle values, I imagine that PJ will want to redo those.

PR #4835, opened 2018-01-24
Revert "[PDI-16747]-Spoon crashes with StackOverflow if job has loop of steps"

emartin-pentaho:
Reverts pentaho/pentaho-kettle#4804


lucboudreau:
Not sure we need this anymore.

PR #4841, opened 2018-01-25
BACKLOG-21021 - reduced AEL WebSocket message sizes by gziping Message objects.  Reduced by 80%

pentaho-jrice:
@ccaspanello @flbrino @baudekin

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/util/EncodeUtil.java
+++ b/core/src/main/java/org/pentaho/di/core/util/EncodeUtil.java
@@ -1,0 +1,91 @@
+/*
+ * ! ******************************************************************************
+ *
+ *  Pentaho Data Integration
+ *
+ *  Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
+ *
+ * ******************************************************************************
+ *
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/pentaho/di/core/util/EncodeUtil.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    // base 64 decode
+    byte[] bytes64 = org.apache.commons.codec.binary.Base64.decodeBase64( string.getBytes(  ) );
+    ByteArrayInputStream zip = new ByteArrayInputStream( bytes64 );
+
+    GZIPInputStream unzip = null;
+    BufferedInputStream in = null;
+    try {
```
baudekin: Both GZipInputStream and BufferedInputStream both have close methods so they can be written in the new Java 8 try style. See https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html . Should be be using the new style?

baudekin:
Are we concerned with code coverage in this section of code.

PR #4849, opened 2018-01-25
[BACKLOG-20931] - adding index and indent for sub-metrics in table view

kurtwalker:


kurtwalker:
@pentaho/ackbar ready for review

PR #4852, opened 2018-01-29
[SP-4022] Backport of PDI-16824 - The sample "Add a checksum - basic CRC32" does not work in 8.0 if you delete a field (8.0 Suite)

Advent51:
- backport of #4721


Advent51:

Advent51:
The broken test passed locally successfully.
Also this is strange, because these changes does not affect TextFileInputMeta at all.

PR #4855, opened 2018-01-29
[PDI-16833] Invalid thread access error when opening a transformation, containing steps of missing plugins, from a repository

AliaksandrShuhayeu:
[PDI-16833] Invalid thread access error when opening a transformation, containing steps of missing plugins, from a repository

-fixed SWTException: Invalid thread access


AliaksandrShuhayeu:

PR #4863, opened 2018-01-29
[SP-4066][PDI-16805] Salesforce Input doesn't get the transformation status 'changed'

AliaksandrShuhayeu:
[SP-4066][PDI-16805] Salesforce Input doesn't get the transformation status 'changed'

- added status 'changed' when button 'Get fields' is pressed
- added status 'changed' for checkboxes of tab Content if they are changed


AliaksandrShuhayeu:

PR #4867, opened 2018-01-30
[BACKLOG-20931] - need to keep the order of sub statuses

kurtwalker:


kurtwalker:
@pentaho/ackbar ready for review

PR #4871, opened 2018-01-30
[CLEANUP] Split assembly into partial components to allow downstream …

hbfernandes:
@pentaho/2-1b @lgrill-pentaho @pedrofvteixeira 
Final assembly is coming out the same, no downstream impact is expected from this change.


pedrofvteixeira:
So, reading through these changes, this new/refactored PDI assembly model can be mapped like so:


 - ### assembly: pdi-ce
   - **pdi-libs**
   - **pdi-plugins**
   - **pdi-static**
   - karaf-assembly-client 
   - dataservice-driver
   - app-launcher
   - kettle-ui-swt
   - swt libs

 - ### assembly: pdi-libs
   - all the content one can see in `kettle/libs` ( reporting, platform, jersey, metaverse, ... )  

 - ### assembly: pdi-plugins
   - all the content one can see in `kettle/plugins`

 - ### assembly: pdi-static
   - .bat / .sh / .command
   - samples ( ktr / kjb )
   - .XUL
   - .xml 


Am I reading this correctly ?

hbfernandes:
@pedrofvteixeira That is correct. Eventually, for your work it might make sense to break these down further so that your minimal assembly can pick only the pieces that it needs.

Note: that kettle-ui-swt on the pdi-ce is there just to get a properties file from within the jar, the actual jar is already included in the pdi-lib. For the sake of simplicity, don't think about it as a required dependency in that module.

pedrofvteixeira:
Cool. This looks pretty 👍  to me. 

One minor suggestion, if I may ( this is merely food for thought, **not** a necessity ): as we're doing this work of breaking down, would it make sense to have `samples` as an assembly of its own, and not a part of `PDI static` ? They are isolated / self-contained enough to be considered an artifact of its own... dunno, I'm seeing that as opening up doors in the likes of "you can download our pdi-samples.zip artifact" 

Something like:

 - ### assembly: pdi-ce
   - **pdi-libs**
   - **pdi-plugins**
   - **pdi-static**
   - **pdi-samples**
   - karaf-assembly-client 
   - (...)

 - ### assembly: pdi-libs
   - (...)

 - ### assembly: pdi-plugins
   - (...)

 - ### assembly: pdi-static
   - .bat / .sh / .command
   - .XUL
   - .xml 

 - ### assembly: pdi-samples
   - samples ( ktr / kjb )

hbfernandes:
I can do that i think, i'll have a look.

hbfernandes:
@pedrofvteixeira Updated with the samples split into it's own module.

pedrofvteixeira:
👍  

@lgrill-pentaho , @pentaho/2-1b 

CC'ing @pentaho/rogueone

PR #4887, opened 2018-02-02
[PDI-16382]-Select Values Step Throws Errors when multiple fields with

SergeyTravin:
same field name occur


SergeyTravin:
@mchen-len-son, please review

mchen-len-son:
@mbatchelor @pamval please take a look
thanks

PR #4890, opened 2018-02-02
[BACKLOG-21097] Added common step dialog components in preparation for BACKLOG-21097

pentaho-amessier:
@pentaho/r2d2


mdamour1976:
@pentaho-amessier So you think a common step includes:
username
password
port
connection?

Something for your (infinite) consideration is:
https://github.com/pentaho/maven-project-archetypes

PR #4893, opened 2018-02-05
[PDI-16387] Running a Job for Every Input Row, sub job does not reset variables or parameters with every execution

AliaksandrShuhayeu:
[PDI-16387] Running a Job for Every Input Row, sub job does not reset variables or parameters with every execution
added recovering changed values of variables if we use VARIABLE_TYPE_CURRENT_JOB


AliaksandrShuhayeu:

PR #4894, opened 2018-02-06
[PDI-16715] Hadoop File Output/Input: Parameter support: If using par…

pentaho-lmartins:
…ameters to define hdfs-path additional */* is required in the path

@duarteteixeira @pamval


pentaho-lmartins:
Updated copyright year.

@pamval @duarteteixeira

PR #4898, opened 2018-02-06
[BACKLOG-20717] Google Drive VFS pressing Close button throws an exception

e-cuellar:


e-cuellar:
@bmorrise Please Sir!

PR #4899, opened 2018-02-07
[PDI-16747] Spoon crashes with StackOverflow if job has loop of steps

mchen-len-son:


mbatchelor:
Oddly, wingman shows this was fine, but the code doesn't compile for me locally or in CI. Preparing a PR to fix it.

kurtwalker:
Thanks @mbatchelor

PR #4901, opened 2018-02-07
[PDI-16773]-Error message is thrown when using Variables for Run

SergeyTravin:
Configuration in Spoon


SergeyTravin:
@mchen-len-son, @mbatchelor please review

PR #4906, opened 2018-02-08
[BACKLOG-20363] When a user accesses the Help>Documentation... option…

ddiroma:
…s from PDI it should load in a new window not a canvas tab


ddiroma:
@bmorrise

PR #4268, opened 2017-08-10
[BACKLOG-17366] - I want to support multiple topic ingestion from a s…

Aliaksandr-Kastenka:
…treaming ingestion step(bug with freezing addressed)

Fix for the bug of freezing spoon in case user chose invalid Hadoop Cluster in Kafka Consumer step and click on topics table row.
Goes with https://github.com/pentaho/pentaho-ee/pull/502

PR #4908, opened 2018-02-09
[PDI-16895] - In the String operations step, there is no select optio…

Aliaksandr-Kastenka:
…ns in Padding and InitCap item.

Small addition to https://github.com/pentaho/pentaho-kettle/pull/4268


Aliaksandr-Kastenka:
@kurtwalker 
Please review.

PR #4909, opened 2018-02-09
[BACKLOG-20951] - refactoring so that rx stream can stop without inte…

kurtwalker:
…rrupt


kurtwalker:
@pentaho/ackbar fixed test

PR #4911, opened 2018-02-12
[PDI-16527] - XML Input Stream step re-writes data in fields from previous steps

stepanovdg:
What was done:
1) changed unit test
2) added shift for initial fields


stepanovdg:
Test failures do`s not relate to fix.

kurtwalker:
even though the failed tests on here are not the fault of this PR, we should really try and avoid committing anything while there are failed tests.  I like to try and track down when they started failing and notify the devs involved so that they can correct the problem

PR #4924, opened 2018-02-13
[PDI-16844] The step Job Executor should overwrite the child parameter by the parent parameter.

AliaksandrShuhayeu:
[PDI-16844] The step Job Executor should overwrite the child parameter by the parent parameter.

Overwriting parameters and the next logic:
1. When the child parameter does not exist in the parent parameters, keep it.
2. When the child parameter does exist in the parent parameters, overwrite the child parameter by the parent parameter.
3. All other parent parameters need to get copied into the child parameters (when the 'Inherit all variables from the transformation?' option is checked)


AliaksandrShuhayeu:

AliaksandrShuhayeu:

PR #4931, opened 2018-02-15
[PDI-16910] Excel Input does not work when reading from an XLSX created by the MSSQL report

Advent51:
- support of inline strings was added
- test was added


Advent51:

PR #4932, opened 2018-02-15
[PDI-16915] Master doesn't identify slaves as inactive in dynamic car…

vasilii-komarov:
…te cluster when master is Pentaho Server

- The poll for status of slave servers have been added into ClusterSchema and GetSlavesServlet classes


mchen-len-son:
Closing in favor of #4934

PR #4933, opened 2018-02-15
[BACKLOG-20951] - changing text of Stop menu item

kurtwalker:


kurtwalker:
@pentaho/ackbar message changes.  probably don't need wingman

PR #4936, opened 2018-02-16
[BACKLOG-21243] MQTT Consumer - SSL

mkambol:
BaseStreamMeta xml serialization will now use a group tag
if the attribute is part of a injection group.

This will be used to bucket SSL properties under a <SSL> tag.

https://jira.pentaho.com/browse/BACKLOG-21243

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/streaming/common/BaseStreamStepMeta.java
+++ b/engine/src/main/java/org/pentaho/di/trans/streaming/common/BaseStreamStepMeta.java
@@ -49,9 +49,15 @@ import org.pentaho.metastore.api.IMetaStore;
 import org.w3c.dom.Node;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
+
+import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.base.Strings.isNullOrEmpty;
+
 
 public abstract class BaseStreamStepMeta extends StepWithMappingMeta implements StepMetaInterface {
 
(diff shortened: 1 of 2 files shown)


mkambol:
@pentaho/ackbar

engine/src/main/java/org/pentaho/di/trans/streaming/common/BaseStreamStepMeta.java
```diff
@@ -75,20 +80,44 @@
     BeanInjectionInfo info = new BeanInjectionInfo( this.getClass() );
     BeanInjector injector = new BeanInjector( info );
     Map<String, BeanInjectionInfo.Property> properties = info.getProperties();
+
+    StringBuilder buider = new StringBuilder();
+
+    info.getGroups()
+      .forEach( group -> {
+        boolean groupHasName = !isNullOrEmpty( group.getName() );
```
kurtwalker: Can we move this lambda to a method and use a method reference?  It's pretty big for inline lambda.  Giving it a name will help understand what it's doing too.

engine/src/main/java/org/pentaho/di/trans/streaming/common/BaseStreamStepMeta.java
```diff
@@ -75,20 +80,44 @@
...
+
+          properties.entrySet().stream()
+            .filter( entry -> groupProps.contains( entry.getValue() ) )
+            .collect( Collectors.toMap( Map.Entry::getKey, Map.Entry::getValue ) )
+        ) );
+        if ( groupHasName ) {
+          buider.append( "</" ).append( group.getName() ).append( ">\n" );
+        }
+      } );
+    return buider.toString();
```
kurtwalker: I'm also wondering if a map to string and a Collectors.joining would be simpler than the StringBuilder.

mkambol: I ended up switching to a StringBuilder after needing to conditionally insert a <group> tag.  Including the opening/closing tag was getting messy, and really this whole mess should probably be done with a DocumentBuilder or real serialization library so we're not assembling string xml.  Didn't want to go that far now, though.

engine/src/main/java/org/pentaho/di/trans/streaming/common/BaseStreamStepMeta.java
mkambol: Good idea.

PR #4940, opened 2018-02-16
[BACKLOG-20951] - row generator should not send row on next timer whe…

kurtwalker:
…n stopped


kurtwalker:
@pentaho/ackbar fyi, it'll probably be 530 before wingman is done with this one.  I'll log in this evening to check it out.  In the mean time if someone wants to approve, I can merge when wingman finishes

PR #4945, opened 2018-02-20
[PDI-16927]: PostgreSQL Bulk Loader : "DB Name Override" option is not working as expected.

TatsianaKasiankova:
Fixed, added unit tests.


TatsianaKasiankova:
@mbatchelor please review the PR.
`dbNameOverride` using was deleted in https://github.com/pentaho/pentaho-kettle/pull/3788.
The fix is just to add back using this option on processing rows as it was before.

PR #4948, opened 2018-02-21
[JIRA PDI-17053] also create image snapshor from job files

LordNycon:
extend exisiting copyTransformationImage() to also create image from jobs


mbatchelor:
@LordNycon - thank-you for your contribution!! Unfortunately, the product manager for PDI (Jens) has determined that there's a better/automated mechanism for doing this, and has closed the JIRA case as Won't Fix. So I have to close this PR. But thank-you again for contributing to the project!

PR #4978, opened 2018-02-27
[SP-4090][PDI-16857] External Id doesn't work in the step Salesforce upsert.

AliaksandrShuhayeu:
[SP-4090][PDI-16857] External Id doesn't work in the step Salesforce upsert.
- fixed sending Foreign key for Salesforce upsert step


AliaksandrShuhayeu:

PR #4981, opened 2018-02-27
[BACKLOG-21239] Adding support for case where log object deregistered

mkambol:
...while logging is in-flight.

When viewing a referenced subtrans object, if you close the tab
it causes the object to be deregistered, which was formerly causing
an NPE in Slf4LoggingEventListener when a log event is received
for the no-longer registered object.


mkambol:
@pentaho/ackbar

PR #4982, opened 2018-02-27
[BACKLOG-21583] - Prevent creating multiple files when append timestamp is checked

rmansoor:


rmansoor:
Angelo and I have reviewed it together.

PR #4963, opened 2018-02-25
[TEST] Expanding PanIT with a few more test cases

pedrofvteixeira:
Expanding PanIT with a few more test cases:
- a KTR that requires params to be passed along as part of the execution request
- a few KTRs that are designed to purposely fail
- a few KTRs that are meant to have a successful execution
- a few KTRs that specify a logfile and validates the KTR execution output is written on it
  - `new File( logFilePath ).deleteOnExit();` is called in a `try - finally` block

@pentaho/rogueone @mbatchelor @lgrill-pentaho @smaring @pentaho/2-1b 

A few (handy) nice-to-know: 

Under `pentaho-kettle/engine`:
- to trigger the execution of Pan-specific UTs / ITs and no others:
  - `mvn clean verify -Dtest=PanTest -DrunITs -Dit.test=PanIT`
- to do the above **and** allow for source code attachment on executing test cases ( port 5005 ):
  - `mvn clean verify -Dtest=PanTest -DrunITs -Dit.test=PanIT -Dmaven.failsafe.debug`

PR #4983, opened 2018-02-27
[BACKLOG-21649] Decoupling the parameter parsing from the execution l…

pedrofvteixeira:
…ogic in Pan

- From this point onward, `Pan` keeps the parameter parsing logic and the `System.exitJVM()`, but delegates the execution to `PanCommandExecutor`
- `PanCommandExecutor` handles all the execution logic, returning an enum `PanStatusCode`, as per https://help.pentaho.com/Documentation/8.0/Products/Data_Integration/Command_Line_Tools

The PR ( https://github.com/pentaho/pentaho-kettle/pull/4963 ) we issued a few days ago to expand `PanIT`'s integration tests served as the basis, where this refactoring work was proven out

@pentaho/rogueone please review

### DIFF
--- a/engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
+++ b/engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
@@ -1,14 +1,33 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
+ *
+ *******************************************************************************
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
```diff
@@ -0,0 +1,581 @@
...
+
+        } else {
+            // the trans does not have a return code.
+            return PanReturnCode.ERRORS_DURING_PROCESSING.getCode();
+        }
+      }
+
+    } catch ( KettleException ke ) {
+
+      System.out.println( BaseMessages.getString( getPkgClazz(), "Pan.Log.ErrorOccurred", "" + ke.getMessage() ) );
```
pdesai16: Is there a reason we are using System.out and logger

pedrofvteixeira: Yes. Maintaining the way the logger took place, when it was located in `Pan.java` ( see the way it used to be, on the 8.0.0.0. branch, at https://github.com/pentaho/pentaho-kettle/blob/8.0.0.0-R/engine/src/main/java/org/pentaho/di/pan/Pan.java#L606-L607 )

Apparently, Pan's execution uses `System.out`, unless you pass along a "logfile" parameter, stating a path to a log file it would write into ( see "logfile" param at https://help.pentaho.com/Documentation/8.0/Products/Data_Integration/Command_Line_Tools )

PR #4985, opened 2018-02-28
[SP-4129]-[PDI-16895] - In the String operations step, there is no se…

PashkovichAndrei:
…lect

options in Padding and InitCap item. (#4908) (8.0 Suite)


PashkovichAndrei:
@mchen-len-son could you please do review?

PR #4987, opened 2018-02-28
[BACKLOG-21649] Addding Scale.java; passing hardcoded strings to Cons…

pedrofvteixeira:
…t.java; adding a few more tests on PanIT

@pentaho/rogueone please review


pentaho-apandit:
Ignoring coverage warning from wingman as the changes are in existing class and the new class just has a main method.

PR #4992, opened 2018-03-01
[SP-4000][PDI-16818] - Fix for run window taking ages to appear and t…

mchen-len-son:
…est cases


mchen-len-son:
@pentaho-lmartins @stepanovdg please merge

mchen-len-son:
adding @mbatchelor

PR #4993, opened 2018-03-01
[SP-4138][PDI-16747]-Spoon crashes with StackOverflow if job has loop…

mchen-len-son:
… of steps


mchen-len-son:
the newly broken test gets fixed by PR https://github.com/pentaho/pentaho-kettle/pull/4995.

@pentaho-lmartins @stepanovdg @mbatchelor please review

PR #4994, opened 2018-03-01
[SP-4139][PDI-16747]-Spoon crashes with StackOverflow if job has loop…

mchen-len-son:
… of steps


mchen-len-son:
@pentaho-lmartins @stepanovdg please merge

PR #4997, opened 2018-03-01
[BACKLOG-16710] - Adds a RestorePDIEnvironment Class Rule to be used …

tmcsantos:
…with some tests.

It allows tests to be started on a clean PDI environment, regardless of any previous tests run.
Enables reuseForks property.


tmcsantos:
@pentaho/x-wing take your time to review
follow the ticket for a better discription

PR #5000, opened 2018-03-02
[PDI-16562]-Comparison of Binary fields of same length and value fails

denisprotopopov:
- fixed. changed Arrays.hashCode->Arrays.deepHashCode


denisprotopopov:
@bmorrise @mbatchelor @pamval please review the changes

PR #5003, opened 2018-03-02
[BACKLOG-18455] Improve the experience for when a user tries to access a recently deleted file in PDI

pentaho-amessier:


plugins/file-open-save/core/src/main/javascript/app/components/error/error.component.js
```diff
@@ -285,6 +286,12 @@ define([
             "",
             i18n.get("file-open-save-plugin.error.unable-to-rename-folder-opened.close.button"));
           break;
+        case 16:// Unable to Open Recent file
+          _setMessage(i18n.get("file-open-save-plugin.missing-recent.title"),
+              i18n.get("file-open-save-plugin.missing-recent.message", {"filename": vm.errorFile.name}),
```
ddiroma: @pentaho-amessier We need to check for filename truncation here...like some of the other errors do like 1, 2, 5, 6, and 7

PR #5007, opened 2018-03-05
[PDI-16818] Improvements to RowMeta have slowed down again in 7.x+

mchen-len-son:
- added cache invalidation on changing trans hop`s state


mchen-len-son:
@mbatchelor @pentaho-lmartins @stepanovdg please review

PR #5009, opened 2018-03-05
[SP-3999][PDI-16818] Improvements to RowMeta have slowed down again i…

mchen-len-son:
…n 7.x+

 - added cache invalidation on changing trans hop`s state


mchen-len-son:
@mbatchelor @pentaho-lmartins @stepanovdg please review

PR #5012, opened 2018-03-06
[PDI-16932] Specifying the Select Fieldname in the Metadata Injectionstep does not give correct results.

AliaksandrShuhayeu:
[PDI-16932] Specifying the Select Fieldname in the Metadata Injectionstep does not give correct results.
added default value for default values for length and precision


AliaksandrShuhayeu:

PR #4944, opened 2018-02-20
[PDI-16893] TextFileOutput has a dangerous "feature" allowing data to be piped to OS commands.

AliaksandrShuhayeu:
[PDI-16893] TextFileOutput has a dangerous "feature" allowing data to be piped to OS commands.

- created TextFileOutputLegacy with capability run as command
- removed capability run as command from TextFileOutput

PR #5015, opened 2018-03-07
Revert "[PDI-16893] TextFileOutput has a dangerous "feature" allowing data to be piped to OS commands."

mbatchelor:
Reverts pentaho/pentaho-kettle#4944


mbatchelor:
The Copyright stuff is actually correct because the revert is putting the files back to what they were prior to Aliaksandr's commit

PR #5019, opened 2018-03-09
[BACKLOG-21866] Removing nonNull requirement for trans and step objid

mkambol:
They can be null during repo save operations.


mkambol:
@pentaho/ackbar

PR #5027, opened 2018-03-12
[BACKLOG-22072] DET / DataServices not working with Stream Lookup step

mchen-len-son:
- getPrevSteps should also be using step meta cache keys


mchen-len-son:
@mbatchelor @mdamour1976 @pamval please have a look

mchen-len-son:
@pamval @pentaho-lmartins can you please take a look?
this is a cleanup after #5023. we need to include it to the current patch.

PR #5030, opened 2018-03-12
[SP-4000][BACKLOG-21569] - AEL - some of t_Call_Vol_analysis_Spark tr…

mchen-len-son:
…ansformation steps works wrong, transformation is writing empty output only over AEL daemon


mchen-len-son:
@stepanovdg @pentaho-lmartins please merge

PR #5032, opened 2018-03-12
[Backlog-20158] implements annotated job plugin dialogs

tmcsantos:
- refactors previous StepDialog annotation to PluginDialog
- updated core plugins to use new annotation
- refactored ms-access to be inline with standards


tmcsantos:
@pentaho/x-wing pls review

PR #5036, opened 2018-03-13
[SP-3999][BACKLOG-21569] AEL - some of t_Call_Vol_analysis_Spark tran…

mchen-len-son:
…sformation steps works wrong, transformation is writing empty output only over AEL daemon


mchen-len-son:
@stepanovdg @pentaho-lmartins please merge

PR #5056, opened 2018-03-15
[PDI-16845] When accessing files using vfs, sftp connections remain o…

mchen-len-son:
…pen even after transformation/job completes

 - re-worked custom sftp provider, file system and file object
 - added cusomt providers.xml to make our custom sftp provider managed by vfs
 - sftp file system is now releasable even when there are open file objects (but not open streams!). this allows us to release sftp connections when a job or transformation is finished.

please, have a look at SftpFileSystemWindowsProvider's javadoc for details


mchen-len-son:
@mbatchelor can you please take a look?

mbatchelor:
Great job on this. Really nice solution!

PR #5058, opened 2018-03-15
[BACKLOG-21800] Adding support for @InjectionDeep to StepMetaProps

mkambol:
serialization.

This will allow embedding objects in JmsConsumer meta objects for
better reuse.


mkambol:
@pentaho/ackbar

PR #5060, opened 2018-03-15
[BACKLOG-21831] - Do not see streaming dataservices when creating a p…

plagoa:
…entaho datasource

- filtered out streaming data services from the getTableMap method

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/database/Database.java
+++ b/core/src/main/java/org/pentaho/di/core/database/Database.java
@@ -3,7 +3,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/pentaho/di/core/database/Database.java
```diff
@@ -3677,7 +3677,11 @@ public RowMetaInterface getReturnRowMeta() {
   }
 
   public String[] getTablenames( String schemanamein, boolean includeSchema ) throws KettleDatabaseException {
-    Map<String, Collection<String>> tableMap = getTableMap( schemanamein );
+    return getTablenames( null, includeSchema, null );
```
marcoslarsen: shouldn't this be getTablenames( schemanamein, includeSchema, null ); instead of null in the first parameter?

core/src/main/java/org/pentaho/di/core/database/Database.java
```diff
@@ -3697,6 +3701,10 @@ public RowMetaInterface getReturnRowMeta() {
   }
 
   public Map<String, Collection<String>> getTableMap( String schemanamein ) throws KettleDatabaseException {
+    return getTableMap( null, null );
```
joaormf: Maybe return getTableMap( schemanamein , null ); ??

PR #5064, opened 2018-03-16
[SP-4104] Backport of PDI-16804 - DET charts can show no data, when it should show data (7.1 Suite)

Advent51:
- partial backport of BACKLOG-18754


Advent51:
@mchen-len-son please review

PR #5069, opened 2018-03-18
[BACKLOG-21800] JMS Consumer - Changes for blueprint load

mkambol:
WrappedWebpshereMQProvider is used to prevent blueprint
from failing with ClassNotFound when loading the
WebsphereMQProvider, since that class directly references
IBMMQ classes which may not be available.
Also makes it possible to give a message for users to
check whether they've installed the IBMMQ jars, comparable
to the older version of the jms plugin.


mkambol:
@pentaho/ackbar

PR #5070, opened 2018-03-19
[PDI-16986] - Metadata Injection: Get "index out of bound" error when

SergeyTravin:
option "Run resulting transformation" is unchecked


SergeyTravin:
@pentaho-lmartins, please review

mchen-len-son:
@mbatchelor @pamval please review

PR #5081, opened 2018-03-20
[PDI-17061] "Get rows from result" step does not pass rows when a transformation is executed on a remote carte server

AliaksandrShuhayeu:
[PDI-17061] "Get rows from result" step does not pass rows when a transformation is executed on a remote carte server

added possibility to add result to XML when it needs


AliaksandrShuhayeu:
@mbatchelor Please could you please review it? This is a regression https://jira.pentaho.com/browse/PDI-15781. I affected your changes.

PR #5089, opened 2018-03-20
[BACKLOG-21800] JMS Consumer - Receiver value validation and setting

mkambol:


mkambol:
@pentaho/ackbar

plugins/streaming/impls/jms/src/main/java/org/pentaho/di/trans/step/jms/ui/JmsConsumerDialog.java
```diff
@@ -106,8 +117,17 @@ public JmsConsumerDialog( Shell parent, Object in, TransMeta tr, String sname )
...
-    jmsDelegate.destinationField = fieldsTab.getFieldNames()[1];
+    jmsDelegate.messageField = fieldsTab.getFieldNames()[ 0 ];
+    jmsDelegate.destinationField = fieldsTab.getFieldNames()[ 1 ];
+
+
+    try {
+      jmsDelegate.receiveTimeout = Integer.parseInt( wReceiverTimeout.getText() );
+    } catch ( NumberFormatException nfe ) {
+      // this shouldn't happen due to validation, but just in case.
+      throw new RuntimeException( getString( PKG, "JmsDialog.ReceiveTimeoutInvalid" ), nfe );
```
kurtwalker: I think we usually use ErrorDialog here rather than throwing an exception.  Are we handling environment substitution?

mkambol: I made it so you can't enter non integer values into the text box, so this should only happen if something funny is going on.

PR #5091, opened 2018-03-21
[PDI-15875] - Steps which use JDBC connection with enabled pooling do…

SergeyTravin:
… not

run on DI server


SergeyTravin:
@pentaho-lmartins, @mchen-len-son, @mbatchelor please review.
Notes:
The Unit test was reactoried to speed up the test execution, was added the test case for the issue.
was merged the two files with simular logic.

PR #5093, opened 2018-03-21
[BACKLOG-22185] Decoupled "Column Exists" step from from PDI Engine a…

wseyler:
…nd placed it in plugins/core.


plugins/core/ui/src/main/java/org/pentaho/di/ui/trans/steps/columnexists/ColumnExistsDialog.java
```diff
@@ -60,6 +61,8 @@
 import org.pentaho.di.ui.core.widget.TextVar;
 import org.pentaho.di.ui.trans.step.BaseStepDialog;
 
+@PluginDialog( id = "ColumnExists", image = "CEX.svg", pluginType = PluginDialog.PluginType.STEP,
+    documentationUrl = "https://wiki.pentaho.com/display/EAI/Check+if+a+column+exists" )
 public class ColumnExistsDialog extends BaseStepDialog implements StepDialogInterface {
   private static Class<?> PKG = ColumnExistsMeta.class; // for i18n purposes, needed by Translator2!!
```
tmcsantos: change this to PKG = ColumnExistsDialog.class when the Dialogs messages are splited from the impl

PR #5095, opened 2018-03-22
[PDI-17077]-Erroneous Synchronization in ChangedFlag.java

denisprotopopov:
-changed according Marc Batchelor comment


mchen-len-son:
@mbatchelor @pamval please review

PR #5100, opened 2018-03-22
[SP-4180]: Backport of PDI-16927 - PostgreSQL Bulk Loader : "DB Name Override" option is not working as expected (8.0 Suite).

TatsianaKasiankova:
This is the cherry-pick of the fix for [PDI-16927]: PostgreSQL Bulk Loader : "DB Name Override" option is not working as expected.


TatsianaKasiankova:
@pentaho-lmartins , @mchen-len-son please review.

PR #5108, opened 2018-03-22
[BACKLOG-22404] Sorting dialects as per UX/QA recommendation

e-cuellar:


e-cuellar:
@bmorrise Please Sir!
@mdamour1976 Please Sir!

dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
```diff
@@ -1326,7 +1327,7 @@ private void setConnectionSpecificInfo( DatabaseMeta meta ) {
     getControls();
 
     if ( databaseDialectList != null ) {
-      databaseDialectList.setElements( databaseDialects  );
+      databaseDialectList.setElements( Arrays.asList( databaseDialects.toArray() )  );
```
bmorrise: Would it work to use Collections.sort on the ArrayList instead of using a TreeSet?

e-cuellar: Let my try that

PR #5113, opened 2018-03-23
[SP-4182] Backport of PDI-16910 - Excel Input does not work when reading from an XLSX created by the MSSQL report (8.0 Suite)

Advent51:
- backport of #4931


Advent51:

PR #5118, opened 2018-03-23
[PDI-16921] Concurrent access to outputRowSets in Mapping Init

mchen-len-son:
- syncronized methods were inroduced
- all unsafe modifications were replaced by new safe methods

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
+++ b/engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
@@ -34,7 +34,6 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Hashtable;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
@@ -165,7 +164,7 @@ public class BaseStep implements VariableSpace, StepInterface, LoggingObjectInte
 
   private Trans trans;
 
-  private Object statusCountersLock = new Object();
+  private final Object statusCountersLock = new Object();
 
   /**
    * nr of lines read from previous step(s)
@@ -242,10 +241,14 @@ public class BaseStep implements VariableSpace, StepInterface, LoggingObjectInte
    */
   private List<RowSet> inputRowSets;
 
+  private final Object inputRowSetsLock = new Object();
+
   /**
    * the rowsets on the output, size() == nr of target steps
    */
   private List<RowSet> outputRowSets;
+
+  private final Object outputRowSetsLock = new Object();
 
   /**
    * The remote input steps.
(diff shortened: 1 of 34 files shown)


mchen-len-son:
@mbatchelor @pamval please review this one

guarding input/output row sets with a plain mutex instead of rwlock or stamped lock, as they seem to be unconteded most of the time (+ lots of write operations). so the main goal is thread safety. and we don't want to affect execution cost much.

mchen-len-son:
when merging this, please merge as well
- https://github.com/pentaho/big-data-plugin/pull/1356
- https://github.com/pentaho/pentaho-cassandra-plugin/pull/108

engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
```diff
@@ -242,11 +241,15 @@
    */
   private List<RowSet> inputRowSets;
 
+  private final Object inputRowSetsLock = new Object();
```
mbatchelor: This would be much better using a ConcurrentReadWriteLock. Many of the accesses are read-only.

engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
```diff
@@ -1310,44 +1313,42 @@ private void handlePutRow( RowMetaInterface rowMeta, Object[] row ) throws Kettl
...
   /**
    * Copy always to all target steps/copies
-   *
-   * @param rowMeta
-   * @param row
    */
   private void mirrorPartitioning( RowMetaInterface rowMeta, Object[] row ) {
-    for ( int r = 0; r < outputRowSets.size(); r++ ) {
-      RowSet rowSet = outputRowSets.get( r );
+    for ( RowSet rowSet : outputRowSets ) {
```
mbatchelor: Seems to me, the outputRowSets can mutate during iteration. So I'd think a this would need a readLock at this point.

engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
```diff
@@ -1750,33 +1751,33 @@ private void verifyRejectionRates() {
...
   /**
    * Find the next not-finished input-stream... in_handling says which one...
    */
   private void nextInputStream() {
-    synchronized ( inputRowSets ) {
-      blockPointer = 0;
+    blockPointer = 0;
 
-      int streams = inputRowSets.size();
+    int streams = inputRowSets.size();
```
mbatchelor: Same comment as above - inputRowSets can be mutated above, so likely need a readWriteLock.lock here.

engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
```diff
@@ -1839,106 +1840,109 @@ protected void waitUntilTransformationIsStarted() {
...
+        } else {
+          // Try once more...
+          // If row is still empty and the row set is done, we remove the row
+          // set from
+          // the input stream and move on to the next one...
+          //
+          if ( inputRowSet.isDone() ) {
+            row = inputRowSet.getRowWait( 1, TimeUnit.MILLISECONDS );
+            if ( row == null ) {
+              inputRowSets.remove( currentInputRowSetNr );
```
mbatchelor: This is problematic - you are mutating the inputRowSets without using the synchronization block. As said above, think this would be a write lock here...

mchen-len-son: This is actually inside a synchronized block. See https://github.com/pentaho/pentaho-kettle/pull/5118/files/1199c579aef9506c52ed8c1b5b07cafae7117b30#diff-9a94b8b1359d88cee88bab433700873dR1846

engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
mchen-len-son: This is private extracted method only used in synchronized context. I see the only reason to add synchronization here - to make the code less fragile. Should I?

engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
mchen-len-son: Same here - private method only used in synchronized context.

engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
mchen-len-son: @mbatchelor thanks for review.

I tried to avoid doing premature optimization. As far as I see, input/output row sets are read and modified mostly uncontended. When most of the time a single thread is accessing the collections, we would only increase execution cost with no benefits.

let me quote Brian Goetz
>In general, you should probably stick with synchronization unless you have a specific need for one of the advanced features of Lock, or if you have demonstrated evidence (not just a mere suspicion) that synchronization in this particular situation is a scalability bottleneck.
>- https://www.ibm.com/developerworks/java/library/j-jtp10264/

please, correct me if I'm wrong at my assumptions about contention (otherwise, we would have faced ConcurrentModificationException a lot earlier, wouldn't we?) and confirm we do need to convert synchronized to locks.

Thanks.

PR #5121, opened 2018-03-23
[BACKLOG-21593] - logging more informative error when publish fails. …

kurtwalker:
… stopping trans when publish fails


kurtwalker:
@pentaho/ackbar please review

PR #5125, opened 2018-03-23
[BACKLOG-22173] - Decoupled "Blocking Step" from Engine and placed it…

wseyler:
… in plugins/core


plugins/core/ui/src/main/resources/BLK.svg
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
```
cardosov: You haven't remove 'BLK.png' and 'BLK.svg' from 'ui/src/main/resources/ui/images'. Is there a reason for that?

PR #5131, opened 2018-03-26
[SP-4163] - Backport of PDI-16653 - Safe Mode does not error out like it is suppo…

YuryBY:
…sed to in 7.1 (8.0 Suite)

 - JUnit test is provided

@mchen-len-son, @pentaho-lmartins, could you please take a look?


pentaho-lmartins:
@YuryBY, can you check the broken tests?

PR #5132, opened 2018-03-26
[SP-4192][PDI-16915] Master doesn't identify slaves as inactive in dy…

mchen-len-son:
…namic carte cluster when master is Pentaho Server

- The poll for status of slave servers have been added into ClusterSchema and GetSlavesServlet classes


mchen-len-son:
@pentaho-lmartins can you please have a look?

PR #5133, opened 2018-03-26
[SP-4183][PDI-16904]-Calculator step : The result of Jaro similitude and JaroWinkler similitude formula's are showing incorrect values.

denisprotopopov:
- fixed


mchen-len-son:
test passes locally.
looks like it's non-deterministic.
merging.

PR #5137, opened 2018-03-26
[SP-4170] - Backport of PDI-16884 - Cassandra Output step does not wo…

YuryBY:
…rk with apache-cassandra-2.1.0 (8.0 Suite)


YuryBY:
@mchen-len-son, @pentaho-lmartins, could you please take a look?

PR #5139, opened 2018-03-26
[BACKLOG-21471] Adds a custom set of fields for GBQ connections.

lucboudreau:


lucboudreau:
@bennychow For a quick second set of eyes. This was validated by UX.

dbdialog/src/main/resources/org/pentaho/ui/database/databasedialog.properties
```diff
@@ -23,6 +23,8 @@ DatabaseDialog.label.UseConnectionPool=Enable Connection Pooling
 DatabaseDialog.label.InitialPoolSize=Initial\:  
 DatabaseDialog.label.MaximumPoolSize=Maximum\:
 DatabaseDialog.OracleTab.title=Oracle
+DatabaseDialog.label.GBQServiceURL=Google BigQuery API URL\:
```
bennychow: According to Simba documentation, this should still be called "Host" and not "URL".  I'd recommend leaving this as the default "Host Name" label.  Port is technically part of a URL but not part of the Host.

lucboudreau: I've been scratching my head about this one too. I'll put it back to host. I was really on the fence about this. I just didn't want it to be confused with a full JDBC URL.

PR #5141, opened 2018-03-26
[BACKLOG-22283] Revert to functionality where table comboVars show variable icon in cell when selected and not outside of table

ricksjames:
[BACKLOG-22283] Revert to functionality where table comboVars show variable icon in cell when selected and not outside of table


ricksjames:
@pentaho/ackbar Please review

ricksjames:
Need to remove insertImage boolean from TableView since the variable icon should never show up on the outside of the table.

ricksjames:
@pentaho/ackbar ready again for review.

PR #5144, opened 2018-03-27
[PDI-16844] Steps Job Executor and Transformation should overwrite the child parameter by the parent parameter.

AliaksandrShuhayeu:
[PDI-16844] Steps Job Executor and Transformation should overwrite the child parameter by the parent parameter.

- correction for case when mapping Variables is null. JobEntryTrans has parameters == null, when it was created not from UI.


AliaksandrShuhayeu:
@bmorrise @mchen-len-son Could you please review and merge it?

PR #5150, opened 2018-03-28
[SP-4193][PDI-16891]-Incorrect remainder in division of numbers in Calculator step 

denisprotopopov:
- fixed. now source values convert to the target data type before calculation.


denisprotopopov:
@mbatchelor could you please to merge this PR

PR #5152, opened 2018-03-28
[BACKLOG-22533] Getting error dialog when attempting to save a job wi…

ddiroma:
…th same name as trans and vice versa


ddiroma:
@bmorrise

PR #5170, opened 2018-03-30
[PDI-16986]-Metadata Injection: Get "index out of bound" error when o…

SergeyTravin:
…ption

"Run resulting transformation" is unchecked - the clone should return new
same object


SergeyTravin:
@pentaho-lmartins, @mbatchelor please review

CravoBranco:
Replacing @SergeyTravin as assigned dev to this PR.

PR #5181, opened 2018-04-03
[PDI-16658] - Fixes for the effects of PDI-16658.

mbatchelor:
On PDI-16658, I advocate for actual re-writes of the code creating the issues in this case. The code is really broken. That said, these fixes are tactical and alleviate the massive performance slow-down, so they're valuable until such time as a sprint team can take up the task of re-implementation of the logging registry using a proper LRU cache and the observer (or possibly the visitor) pattern to make this area of PDI sane.


mbatchelor:
Added test case

PR #5182, opened 2018-04-03
[PDI-16387]-Running a Job for Every Input Row, sub job does not reset

SergeyTravin:
variables or parameters with every execution


SergeyTravin:
Closed as it was declined

PR #5194, opened 2018-04-04
[PDI-16921] Concurrent access to outputRowSets in Mapping Init

mchen-len-son:
- replaced synchronized with read write locks

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
+++ b/engine/src/main/java/org/pentaho/di/trans/step/BaseStep.java
@@ -39,6 +39,7 @@ import java.util.Map;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


mchen-len-son:
@mbatchelor @mkambol please review
this is a fix for regression found, when the thread trying to get inputrowsets size gets blocked, since the row processing thread acquired the lock and is waiting for timeout.

@mbatchelor eventually we've come to readwritelocks as you suggested earlier.

mkambol:
I think this looks okay.  Out of curiosity, did you consider using CopyOnWriteArrayList for the input and output rowset lists?  Could eliminate the risky business of lock juggling.

mchen-len-son:
@mkambol I did. 
inputrowsets are modified widely (not only on initialization, but also on reading -  when a rowset becomes empty it gets removed from the collection). also, mapping steps often write to them.
considering that copyonwritelist uses the same reentrantlock for writing operations, plus making a copy of the entire collection, I think, we would only increase memory consumption.
outputrowsets are modified even more widely.

PR #5198, opened 2018-04-05
[BACKLOG-22642] - trimming jms url before making connections

kurtwalker:


kurtwalker:
@pentaho/ackbar please review

PR #5201, opened 2018-04-05
[BACKLOG-22575]JMS- No data in Preview when param in "Ret fields from"

mkambol:
sub step needed env substitution in the getFields() method.

https://jira.pentaho.com/browse/BACKLOG-22575


mkambol:
@pentaho/ackbar

PR #5206, opened 2018-04-05
[BACKLOG-22633] Set all step rows without errors backgrounds to white so that fixed step errors don't stay red on the step metrics table.

ricksjames:
[BACKLOG-22633] Set all step rows without errors backgrounds to white so that fixed step errors don't stay red on the step metrics table.


ricksjames:
@pentaho/ackbar please review

PR #5209, opened 2018-04-05
[BACKLOG-22362] - populating substebs every time it gains focus

kurtwalker:


kurtwalker:
@pentaho/ackbar please review

PR #5215, opened 2018-04-06
[BACKLOG-22651] As an ETL developer, I want to select files using the VFS browser in the TextFileInput/Output, so that I can easily get VFS URLs

e-cuellar:


e-cuellar:
@bmorrise Please Sir!

PR #5219, opened 2018-04-09
[BACKLOG-22673] Transformation/Job Executor lose state after editing

mdamour1976:


ui/src/main/java/org/pentaho/di/ui/trans/steps/jobexecutor/JobExecutorDialog.java
```diff
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2018 by Hitachi Vantara : http://www.pentaho.com
```
pentaho-amessier: nit-picky, but shouldn't this be changed to 2002-2018?

pminutillo:
@pentaho/r2d2

PR #5222, opened 2018-04-09
[BACKLOG-20815] Extra columns included in append data step.

mkambol:
Caused by a change to TransMeta.getStepFields() (adc9506bd)

That change switched from checking the number of prev steps via
findNrPrevSteps() to doing a count on the results of
findPreviousSteps().  The new logic, however, also includes the “info”
steps, changing behavior and resulting in extra columns in output
for some types of steps.

To address, switched to the overload of findPreviousSteps which
excludes info steps.

https://jira.pentaho.com/browse/BACKLOG-20815


mkambol:
@pentaho/ackbar @VasilinaTerehova

PR #5226, opened 2018-04-09
[BACKLOG-22651] As an ETL developer, I want to select files using the…

e-cuellar:
… VFS browser in the TextFileInput/Output, so that I can easily get VFS URLs


e-cuellar:
@ddiroma Please Sir!

PR #5228, opened 2018-04-09
[BACKLOG-22165] Set the jms property name column to use variables, and get the environment substitute for the property variables

ricksjames:
[BACKLOG-22165] Set the jms property name column to use variables, and get the environment substitute for the property variables


ricksjames:
@pentaho/ackbar please review

PR #5229, opened 2018-04-10
[BACKLOG-22310] Ensure when MQTT is stopped, it is only stopped once.

DFieldFL:


DFieldFL:
@pentaho/ackbar please review

PR #5236, opened 2018-04-10
[BACKLOG-22766] Applying encryption to nested attributes in meta

mkambol:
Formerly the @Sensitive annotation only applied to top level
attributes, not those within an @InjectionDeep property.

https://jira.pentaho.com/browse/BACKLOG-22766


mkambol:
@pentaho/ackbar

PR #5242, opened 2018-04-11
Fixed JavaScript TruncateDate function for Month

bbrieber:
The Truncate Date function returned the wrong value for Month (Arg 5). 

Previous Behavior Truncate sets first of February.
New Behavior Truncate sets first of January.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/script/ScriptAddedFunctions.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/script/ScriptAddedFunctions.java
@@ -2434,7 +2434,7 @@ public class ScriptAddedFunctions {
         switch ( level.intValue() ) {
         // MONTHS
           case 5:
-            cal.set( Calendar.MONTH, 1 );
+            cal.set( Calendar.MONTH, 0 );
             // DAYS
           case 4:
             cal.set( Calendar.DAY_OF_MONTH, 1 );


mbatchelor:
Hey Benjamin,

Thanks for the contribution. I had to close this PR out for a couple of reasons:
1- The PR is for a branch that is not master. We can only accept contributions against the master branch.
2- There was no associated JIRA case. We cannot commit code without a JIRA case. In the future, if you see issues like this in the code, please file a JIRA defect at https://jira.pentaho.com. I created a case to capture your findings here: https://jira.pentaho.com/browse/PDI-17179
3- There are actually two script steps. The one that is being currently maintained is the Modified Javascript step. It's the one we recommend everyone use. Yes - it has the bug as well... :-)
4- The contribution doesn't comply with our formatting requirements. Specifically, the copyright date still says 2017, and whenever we modify files, we have to make sure the copyright date reflects the year the file was updated.
5- There is no test case accompanying your change. This one could likely be acceptable anyway because there's not already an existing unit test for that class - so I probably would have accepted the contribution without a test (or maybe have written my own test case for it). But ordinarily, we don't process pull-requests without accompanying unit tests.

Thanks again for the contribution, and welcome to the Pentaho (now Hitachi-Vantara) community. It would be wonderful if you could submit another PR that at least fixes 1-4 above. It would even be more awesome if you could create a new unit test. Right now, there are two unit tests for ScriptValues_mod that test some things in scriptvalues_mod/ScriptValuesAddedFunctions. You can see them here: https://github.com/pentaho/pentaho-kettle/tree/8.0.0.0-R/engine/src/test/java/org/pentaho/di/trans/steps/scriptvalues_mod.

PR #5244, opened 2018-04-11
[BACKLOG-22805] Displaying MQTT Error messages appropriately

DFieldFL:
When a bad username/password was used the error message wasn't being
caught correctly and displayed. Samething was happening for QOS. This
fix unifies the behavior between MQTT Producer and Consumer when an
error message occurs and disconnects the from the client as needed.


DFieldFL:
@pentaho/ackbar please review

PR #5247, opened 2018-04-11
[BACKLOG-22806] Variable substitution doesn't work with MQTT SSL params

mkambol:
2 commits.  First commit adds a method to BaseSerializingMeta that supports swapping environment variables in all strings within a meta.  This can be used as an all purpose way of setting env variables without needing to remember to handle them individually.  Second commit updates MQTT consumer / producer to use this new method.

### DIFF
--- a/plugins/streaming/impls/mqtt/src/main/java/org/pentaho/di/trans/step/mqtt/MQTTClientBuilder.java
+++ b/plugins/streaming/impls/mqtt/src/main/java/org/pentaho/di/trans/step/mqtt/MQTTClientBuilder.java
@@ -95,7 +95,6 @@ public final class MQTTClientBuilder {
   private String automaticReconnect;
   private MqttCallback callback;
   private String clientId = MqttAsyncClient.generateClientId();  // default
-  private VariableSpace variableSpace;
   private LogChannelInterface logChannel;
   private String stepName;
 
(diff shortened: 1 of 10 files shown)


mkambol:
@pentaho/ackbar

plugins/streaming/impls/mqtt/src/main/java/org/pentaho/di/trans/step/mqtt/MQTTClientBuilder.java
```diff
@@ -242,41 +224,36 @@ public MqttClient buildAndConnect() throws MqttException {
...
       getString( PKG, "MQTTInput.Error.ConnectionURL" ) );
     try {
-      int qosVal = Integer.parseInt( variableSpace.environmentSubstitute( this.qos ) );
+      int qosVal = Integer.parseInt( this.qos );
       checkArgument( qosVal >= 0 && qosVal <= 2 );
     } catch ( Exception e ) {
-      String errorMsg = getString( PKG, "MQTTClientBuilder.Error.QOS",
-        stepName, variableSpace.environmentSubstitute( qos ) );
+      String errorMsg = getString( PKG, "MQTTClientBuilder.Error.QOS", stepName, qos );
+      logChannel.logError( errorMsg );
```
DFieldFL: logError was removed in a previous commit https://github.com/pentaho/pentaho-kettle/pull/5244/files#diff-08cf9294a17b3dad77fb3375fd388bc1L280. The logging is handled in the MQTT Producer and MQTT Consumer.

mkambol:
@DFieldFL - removed log message.

PR #5252, opened 2018-04-12
[PDI-17179] - truncDate typo for Month

mbatchelor:


engine/src/main/java/org/pentaho/di/trans/steps/script/ScriptAddedFunctions.java
```diff
@@ -2412,58 +2413,63 @@ public static Object trunc( ScriptEngine actualContext, Bindings actualObject, O
...
-    } catch ( Exception e ) {
-      throw new RuntimeException( e.toString() );
+      return ScriptAddedFunctions.truncDate( dArg1, level );
+    } else {
+      throw new RuntimeException( "The function call truncDate requires 2 arguments: a date and a level (int)" );
     }
   }
 
+  @VisibleForTesting
+  static Date truncDate( Date dArg1, Integer level ) {
```
pentaho-amessier: This method appears to be exactly the same as the method in the other class. Any reason it can't be moved to some common utility class so the code doesn't have to be duplicated?

mbatchelor: No reason at all. Was just continuing the bifurcation of the two steps (deprecated, non-deprecated). Didn't figure it was valuable to refactor things out of their silos since one is deprecated. If preferred, I can consolidate the methods to one of the utility classes (goodness knows there are a few of them in PDI) and call the common one from both.

pentaho-amessier: Fair enough, and clearly not a huge deal.

pentaho-amessier: Fair enough, and clearly not a huge deal.

PR #5258, opened 2018-04-13
[PDI-17103]-Transformation Executor: Parameters: "Static Input value"field are not populated with Variable value defined in child transformation if Getting Parameters

denisprotopopov:
-fixed

### DIFF
--- a/ui/src/main/java/org/pentaho/di/ui/trans/steps/transexecutor/TransExecutorDialog.java
+++ b/ui/src/main/java/org/pentaho/di/ui/trans/steps/transexecutor/TransExecutorDialog.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
  *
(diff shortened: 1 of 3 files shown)


ui/src/main/java/org/pentaho/di/ui/trans/steps/transexecutor/TransExecutorDialog.java
```diff
@@ -684,7 +684,7 @@ protected void getParametersFromTrans( TransMeta inputTransMeta ) {
 
         TableItem item = new TableItem( wTransExecutorParameters.table, SWT.NONE );
         item.setText( 1, Const.NVL( name, "" ) );
-        item.setText( 3, Const.NVL( desc, "" ) );
+        item.setText( 3, Const.NVL( inputTransMeta.getParameterDefault( name ), "" ) );
```
mbatchelor: So should anything be done with the desc variable? I see that you're getting the parameter default, but it appears to me that if anything used to show up in desc, it won't now. Would it make sense to do something like:

String str = inputTransMeta.getParameterDefault( name );
str = ( str != null ? str : ( desc != null ? desc : "" ) );
item.setText( 3, Const.NVL( str ) );

denisprotopopov:

PR #5263, opened 2018-04-15
[BACKLOG-21778] Removed the finishProcess call after receveing the ca…

flbrino:
…ncel confirmation from the driver. The finishProcess method will be executed by the thread that is waiting for the cancel confirmation, so no need for this call and removing this call avoids block behavior noticed while testing


engine/src/main/java/org/pentaho/di/trans/ael/websocket/TransWebSocketEngineAdapter.java
```diff
@@ -94,6 +94,7 @@
   private final String host;
   private final String port;
   private final boolean ssl;
+  private boolean cancelling = false;
```
baudekin: Shouldn't this be an AtomicBoolean to make it thread safe?

PR #5269, opened 2018-04-17
[BACKLOG-22892] - base stream meta uses a different property for the …

kurtwalker:
…transformation path, probably for no good reason.  export procedure can possibly change filename from StepWithMappingMeta which only knows about filename


kurtwalker:
@pentaho/ackbar please review

PR #5271, opened 2018-04-17
[BACKLOG-20613] Transformation tab name changes after dismissing Unable to save dialog

pentaho-amessier:


pentaho-amessier:
@pentaho/r2d2

PR #5274, opened 2018-04-17
[BACKLOG-21599] Kettle: Trans Executor transform does not handle the …

e-cuellar:
…parameters tab correctly.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/transexecutor/TransExecutor.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/transexecutor/TransExecutor.java
@@ -22,7 +22,11 @@
 
 package org.pentaho.di.trans.steps.transexecutor;
 
+import java.util.List;
 import java.util.ArrayList;
+import java.util.Map;
+import java.util.HashMap;
+import java.util.Arrays;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


e-cuellar:
@baudekin Please Sir!

e-cuellar:
@baudekin Please Sir!

engine/src/main/java/org/pentaho/di/trans/steps/transexecutor/TransExecutor.java
```diff
@@ -272,31 +285,98 @@ Trans createInternalTrans() throws KettleException {
...
+        }
+      } catch ( Exception e ) {
+        //Set the value to the first parameter in the resolvingValuesMap.
+        resolvingValuesMap.put( (String) resolvingValuesMap.keySet().toArray()[i], null );
+      }
+    }
+    /////////////////////////////////////////////
+
+    /////////////////////////////////////////////
+    //If "Fields to use" ARE NOT provided and "Inherit all variables from transformation" IS checked.
```
baudekin: It looks like this is all or nothing. If you have one "field to use", then you look to resolve all parameters for the input row.   The way we did it on AEL we would first look to match the parameter from the "field to use" and if we could not find the parameter there we would then check for the parameter being defined on the transform. If the inherent variables where set.

PR #5276, opened 2018-04-19
[PDI-17117]-"Text file input" step picks up the last file in order when supplies with 2 files of same names

denisprotopopov:
-fixed


denisprotopopov:

PR #5277, opened 2018-04-19
[BACKLOG-22892] - saving in a repository uses filename differently.  …

kurtwalker:
…creating new method for replacing filename so as not to disturb other code paths


kurtwalker:
@pentaho/ackbar please review

PR #5278, opened 2018-04-19
[PDI-17110]: Excel Input cannot process Cognos generated Excel report

TatsianaKasiankova:
There are fix, unit test.


TatsianaKasiankova:
@mbatchelor please review.
I've fixed the case with ref attribute presented as one start cell - `<dimension ref="A1"/>`.
Also I've fixed NPE occurred when we have row(s) empty before table.
Changes covered with tests.

PR #5281, opened 2018-04-19
[BACKLOG-22892] - when sending export, the resources should be found …

kurtwalker:
…in zip rather than repository.  do not include repository config when sending to slave server


kurtwalker:
@pentaho/ackbar @pentaho/r2d2 @mdamour1976 
please review.  I tried a few different scenarios running on slave servers and pentaho servers both connected and not connected to a repository.  all those worked

PR #5293, opened 2018-04-24
BACKLOG-22955 : Create daemon threads for CarteStatusCache

pentaho-apandit:
Use daemon threads to avoid blocking shutdown. Switching the thread launched to be daemon otherwise it blocks the pentaho server shutdown. Also naming the thread created with class name to make tracking of threads easy
Testing : tested with QAT-352. Replaced kettle-engine jar and verified that the server stops and process goes away.

@pentaho/rogueone @pamval @kcruzada @dkincade


pentaho-apandit:
Corrected Copyright year, checkStyle errors and did force push.

pentaho-apandit:
The coverage change warnings are for other classes not changed in this PR.

PR #5311, opened 2018-04-26
[BACKLOG-22915] - handling forced stop and safe stop correctly for AEL

kurtwalker:


kurtwalker:
@pentaho/ackbar please review

PR #5316, opened 2018-04-28
[PDI-8929] JobEntryTalendJobExec has static hashmap classLoaderCache that is mutated without protection

AliaksandrShuhayeu:
[PDI-8929] JobEntryTalendJobExec has static hashmap classLoaderCache that is mutated without protection

- changed HashMap to ConcurrentHashMap


AliaksandrShuhayeu:

PR #5320, opened 2018-04-30
Help Links need to be created for the JMS Consumer step, JMS Producer step, Amazon EMR Job Executor entry, and Amazon Hive Job entry

ddiroma:
…ep, JMS Producer step, Amazon EMR Job Executor entry, and Amazon Hive Job entry


ddiroma:
@dkincade @kcruzada @pentaho-svigneri

PR #5329, opened 2018-05-01
 [BACKLOG-23129] - Use access key and secret key from existing transformation to execute s3csvfileinput

rmansoor:


rmansoor:
@kcruzada this is for the 8.1 patch release

PR #5334, opened 2018-05-07
[PDI-17144] A sub-job will add up the log entries in the sub-job log …

ppatricio:
…table for every iteration  

@bmorrise @mbatchelor @pentaho-lmartins


ppatricio:
PR updated. Added some comments and three more tests, validating the new function call. Basically, when we start a job we have to unsure, before we write that record to database, that the job's logging buffer do not contain previous events of previous executions. Removing those events(by logChannelId) before execute, keeps the log specific for each execution.

PR #5340, opened 2018-05-08
[BACKLOG-23049] Adds sample standard deviation to Group By step

pentaho-mracine:


engine/src/main/resources/org/pentaho/di/trans/steps/groupby/messages/messages_en_US.properties
```diff
@@ -43,7 +43,8 @@ GroupByDialog.SortDir.Button=&Variable...
 GroupBy.Log.StartingToRun=Starting to run...
 GroupByMeta.TypeGroupLongDesc.CONCAT_COMMA=Concatenate strings separated by ,
 GroupByDialog.Log.GettingKeyInfo=getting key info...
-GroupByMeta.TypeGroupLongDesc.STANDARD_DEVIATION=Standard deviation
+GroupByMeta.TypeGroupLongDesc.STANDARD_DEVIATION=Standard deviation (population)
```
pentaho-amessier: I know this is already merged, but aren't you essentially changing the original standard deviation, which was based on sample size, to a calculation based on the entire population? I thought we wanted to preserve the current (8.1) sample based behavior and add the population based calculation as a new option. It seems that you're doing the opposite here, changing the current behavior to be population based and adding the sample based calculation as a new option.

pentaho-mracine: I know I already talked to you about this but I just wanted to leave this here in case anyone else was looking on. The way we were calculating it was the population based calculation. It calculates it differently from the way standard deviation is normally calculated (with the changing average) but still is a correct calculation in regards to the population. The option I added is for the sample standard deviation.

pentaho-amessier: My understanding of "population based" and "sample based" was wrong, I retract my previous comment :)

PR #5347, opened 2018-05-10
[BACKLOG-23112] DET: many errors running test case "with SDR"

mchen-len-son:
- StreamLookupMeta should clone value metas before changing

Having introduced the caching of rowMetas, we observe that StreamLookupMeta fails to find a field it needs when a cache hit occurs. Because previously it already found and changed that field in rowMeta (due to its configuration). 
Before introducing step field cache, rowMeta was instantiated every time it was needed, and it was fine to change the field.
Now StreamLookupMeta should clone valueMetas before changing, as not to break cached objects.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/TransMeta.java
+++ b/engine/src/main/java/org/pentaho/di/trans/TransMeta.java
@@ -2044,9 +2044,10 @@ public class TransMeta extends AbstractMeta
     // Go get the fields...
     //
     RowMetaInterface before = row.clone();
-    compatibleGetStepFields( stepint, row, name, inform, nextStep, this );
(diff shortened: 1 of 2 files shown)


mchen-len-son:
@mbatchelor @kurtwalker please have a look

mchen-len-son:
@mbatchelor @kurtwalker @bmorrise can you please have a look?

kurtwalker:
Change is fine, but I'm worried that there are many steps besides this one that would need a similar change as a result of the performance improvements for getFields.  When individual steps need to worry about the caching, I think we're getting something wrong.

mchen-len-son:
@kurtwalker updated the changes - cloning is done in TransMeta covering all staps that might change informational row meta.

pentaho-lmartins:
Backported to 8.1 in #5543.

PR #5349, opened 2018-05-10
[BACKLOG-22705][CLEANUP] GlobalMessageUtil optimizations and performance improvements

pentaho-amessier:
**Stats prior to this fix:**
`GlobalMessageUtil.getBundle() ....................... 473,847 ms`
`GlobalMessageUtil.calculateString() ................... 6,243 ms`
`GlobalMessageUtil.getActiveLocales() .................. 6,165 ms`
`GlobalMessageUtil.getLocaleString() ..................... 573 ms`
`GlobalMessageUtil.buildHashKey() ........................ 480 ms`
`GlobalMessageUtil.<clinit> .............................. 203 ms`
_Total build time: 10:38 min_

**Stats after the fix:**
`GlobalMessageUtil$GlobalMessageControl.newBundle() ...... 235 ms`
`GlobalMessageUtil.calculateString() ..................... 200 ms`
`GlobalMessageUtil.getActiveLocales() .................... 192 ms`
`GlobalMessageUtil.getBundle() ............................ 99 ms `
`GlobalMessageUtil.isMissingKey()	......................... 98 ms`
`GlobalMessageUtil.<clinit> ................................  0 ms`
_Total build time: 02:31 min_

@pentaho/r2d2 @mbatchelor @tmcsantos @emartin-pentaho

### DIFF
--- a/core/src/main/java/org/pentaho/di/i18n/GlobalMessageUtil.java
+++ b/core/src/main/java/org/pentaho/di/i18n/GlobalMessageUtil.java
@@ -32,18 +32,18 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.net.URL;
+import java.net.URLConnection;
+import java.nio.charset.StandardCharsets;
 import java.text.MessageFormat;
 import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
-import java.util.Map;
 import java.util.MissingResourceException;
 import java.util.PropertyResourceBundle;
 import java.util.ResourceBundle;
-import java.util.stream.Collectors;
+import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


lgrill-pentaho:
yowza!

pentaho-amessier:
@lgrill-pentaho it looks like more than it actually is, a good chunk of it is some old cleanup, and updated/re-formatted Javadoc. The most significant change was to move the bundle creation to the controller class, where we have some built in fallback for free, hence I was able to trim the active locale set. That’s the change that seemed to have the most significant performance impact. Second, I got rid of the “potential match” concept, which basically looked at all the locales and returned the best match for a given key, based on the order of active locales, rather than return the first match, which is quicker. Lastly, I replaced String.format, which is inefficient, with StringBuilder, this may not have had any visible impact, but definitely doesn’t hurt.

pentaho-amessier:
@pentaho/r2d2

101000:
There is also a major performance issue when using MessageFormat.format API in this class .. can you consider switching to StringBuilder ?

PR #5351, opened 2018-05-11
[BACKLOG-21139] Add new "static base" assembly with minimal files for worker nodes

pentaho-whartman:
@pentaho/rogueone @pentaho/x-wing

Jira: https://jira.pentaho.com/browse/BACKLOG-21136

This PR adds a smaller version of the `static` assembly called `static-base` (which only has the files required to run pan/kitchen on linux). This is needed for the `pdi-ee-client-light` assembly that will be used for worker nodes.

The `static` assembly was changed to build off of the `static-base` assembly to hopefully simplify the maintenance of the `pdi-ee-client-light` assembly.

This PR is related to https://github.com/pentaho/pentaho-ee/pull/1221


assemblies/static-base/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+
+  <artifactId>pdi-static-base</artifactId>
+  <packaging>pom</packaging>
+
+  <name>PDI Client Static Assets</name>
+
+  <parent>
+    <artifactId>pdi-assemblies</artifactId>
+    <groupId>org.pentaho.di</groupId>
+    <version>8.2.0.0-SNAPSHOT</version>
```
lgrill-pentaho: Will need to update versions from 8.2.0.0-SNAPSHOT to 9.0.0.0-SNAPSHOT

PR #5360, opened 2018-05-15
[PDI-17119] Mapping (sub transformation) step : Using variables/parameters in the parent transformation to resolve the sub-transformation name

AliaksandrShuhayeu:
[PDI-17119] Mapping (sub transformation) step : Using variables/parameters in the parent transformation to resolve the sub-transformation name

- added resolving a child file name via a parent space


AliaksandrShuhayeu:

PR #5363, opened 2018-05-16
[PDI-16614]-Data service in CSV File Input does not work with Lazy conversion active

denisprotopopov:
- fixed

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
+++ b/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
@@ -988,7 +988,8 @@ public class ValueMetaBase implements ValueMetaInterface {
     }
 
     try {
-      return getDecimalFormat( false ).format( number );
+      DecimalFormat format = getDecimalFormat( false );
+      return format.format( number );
     } catch ( Exception e ) {
       throw new KettleValueException( toString() + " : couldn't convert Number to String ", e );
     }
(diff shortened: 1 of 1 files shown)


pentaho-lmartins:
@mbatchelor @bmorrise @pamval

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -988,7 +989,12 @@ protected synchronized String convertNumberToString( Double number ) throws Kett
     }
 
     try {
-      return getDecimalFormat( false ).format( number );
+      DecimalFormat format = getDecimalFormat( false );
+      /*String num = new BigDecimal( number.toString() ).stripTrailingZeros().toPlainString();
```
mbatchelor: Seems this commented code should be removed.

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -1012,6 +1018,9 @@ protected synchronized Double convertStringToNumber( String string ) throws Kett
 
     try {
       DecimalFormat format = getDecimalFormat( false );
+      if ( string.indexOf( format.getDecimalFormatSymbols().getDecimalSeparator() ) > 0  ) {
```
mbatchelor: As I understand the "maximum faction digits" setting, it's unnecessary to be so precise and set it to the length of the fractional part. The better (and less costly performance-wise) is to set it to 100 or so. Then no precision is lost, and you don't have to strip off any zeros. See my comment on this case: https://jira.pentaho.com/browse/ANALYZER-3728?focusedCommentId=338607&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-338607

cravobranco:
The unit tests were not added by PR owner, so I will close this PR in order to add unit tests.
Therefore, a new PR will be open soon.

PR #5368, opened 2018-05-17
[BACKLOG-21387] Fixing 'enter number field' dialog width for previewing data

pentaho-amessier:


pentaho-amessier:
@pentaho/r2d2

PR #5377, opened 2018-05-21
[SP-4301][PDI-17086] - Transformation Executor Step does not receive parameters from fields

denisprotopopov:
-fixed


pentaho-lmartins:
Running CheckStyle on IntelliJ shows no problems.

PR #5386, opened 2018-05-21
[BACKLOG-23396] Adds metadata injection to Set Field Value step

axelguiloff:
@pentaho/r2d2

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/setvaluefield/SetValueField.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/setvaluefield/SetValueField.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/pentaho/di/trans/steps/setvaluefield/SetValueFieldMeta.java
```diff
@@ -48,10 +50,14 @@
...
 import org.w3c.dom.Node;
 
+@InjectionSupported( localizationPrefix = "SetValueField.Injection.", groups = { "FIELDS" } )
 public class SetValueFieldMeta extends BaseStepMeta implements StepMetaInterface {
   private static Class<?> PKG = SetValueFieldMeta.class; // for i18n purposes, needed by Translator2!!
 
+  @Injection( name = "FIELD_NAME", group = "FIELDS" )
   private String[] fieldName;
+
+  @Injection( name = "REPLACE_BY_FIELD_VALUE", group = "FIELDS" )
```
pentaho-amessier: Out of curiosity - have you tested the injections without specifying this field in the injector? The reason I ask is because with the step I'm working with (Modified Java Script Value) I initially was getting ArrayOutOfBoundsExceptions, unless I provided values for ALL field columns, because these field related String arrays were not being sized properly (allocate method was not called) when going through MDI. I had to add provisions specifically to prevent these ArrayOutOfBoundsExceptions. This may not apply in your case, but just wanted to point this out.

axelguiloff: It appears it does happen with mine, nice catch. I'll look into adding handling for it, thanks.

engine/src/main/java/org/pentaho/di/trans/steps/setvaluefield/SetValueField.java
```diff
@@ -89,7 +89,10 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
           throw new KettleStepException( BaseMessages.getString(
             PKG, "SetValueField.Log.CouldNotFindFieldInRow", meta.getFieldName()[i] ) );
         }
-        String sourceField = environmentSubstitute( meta.getReplaceByFieldValue()[i] );
+        String sourceField = environmentSubstitute(
+          meta.getReplaceByFieldValue() != null && meta.getReplaceByFieldValue().length > 0
```
pentaho-amessier: For the future: ArrayUtils.isEmpty( meta.getReplaceByFieldValue() )

PR #5389, opened 2018-05-22
[SP-4281][PDI-16921] Concurrent access to outputRowSets in Mapping Init

mchen-len-son:
- syncronized methods were inroduced
- all unsafe modifications were replaced by new safe methods


mchen-len-son:
@pentaho-lmartins please review
this is a backport of #5118 and #5194

PR #5400, opened 2018-05-23
[BACKLOG-23399][BACKLOG-23394] MDI support for 'Modified Java Script Value' step...

pentaho-amessier:
and an update to 'Add a checksum' MDI description for the 'old checksum behavior' field


pentaho-amessier:
@pentaho/r2d2

pentaho-amessier:
Closing for now, as there might be a better way to deal with the type conversion from string to int

PR #5403, opened 2018-05-23
[BACKLOG-23395] Adds metadata injection support for the Get System In…

pentaho-mracine:
…fo step


pentaho-mracine:
Making more changes

PR #5407, opened 2018-05-23
[BACKLOG-23395] Adds metadata injection support for the Get System In…

pentaho-mracine:
…fo step


engine/src/main/java/org/pentaho/di/trans/steps/systemdata/SystemDataMetaInjectionTypeConverter.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Override
+  public Enum<?> string2enum( Class<?> enumClass, String v ) throws KettleValueException {
+    if ( v == null ) {
+      return null;
+    }
+
+    // For SystemDataMeta, enum should be a SystemDataTypes
+    SystemDataTypes type = SystemDataTypes.getTypeFromString( v );
+    if ( !v.equals( SystemDataTypes.TYPE_SYSTEM_INFO_NONE.toString() ) && type == SystemDataTypes.TYPE_SYSTEM_INFO_NONE ) {
```
pentaho-amessier: fyi, switching the order of the equals check saves you a 'v' null check: !SystemDataTypes.TYPE_SYSTEM_INFO_NONE.toString().equals( v )

Also, why not just simply call SystemDataTypes.getTypeFromString( v )? Why return null if v is null, throw an exception when type is 'none' (but only when value isn't explicitly set to 'none') and return the enum in other cases?

pentaho-mracine: Specifically here https://github.com/pentaho/pentaho-kettle/blob/master/engine/src/test/java/org/pentaho/di/core/injection/BaseMetadataInjectionTest.java#L145-L146 it checks for a bad enum to see it fail. Perhaps I should just have the getTypeFromString( v ) throw the exception, I was just trying to get around having all calls to that function to have to be wrapped in a try/catch.

pentaho-mracine:
I'll be making some changes to this, please hold off merging for now.

pentaho-mracine:
@pentaho/r2d2 please review

PR #5410, opened 2018-05-24
[SP-4277][BACKLOG-22632]-MapReduce job with "Hadoop File Output" step…

mchen-len-son:
… in sub-transform fail with "Could not create output file" error (HA)


YuryBY:
https://github.com/pentaho/pentaho-kettle/pull/5415

PR #5415, opened 2018-05-24
[SP-4277] - [BACKLOG-22632] - MapReduce job with "Hadoop File Output"…

YuryBY:
… step in sub-transform fail with "Could not create output file" error (HA)


YuryBY:
Have updated the PR to repair two remained tests.

YuryBY:
@pentaho-lmartins, tests pass

PR #5417, opened 2018-05-24
[CLEANUP]  Extracting calls to pop a message dialog in TransGraph

mkambol:
To reduce code footprint and allow for consistent changes if needed.


mkambol:
@pentaho/ackbar

PR #5428, opened 2018-05-30
[BACKLOG-23276] Clean up deprecated steps presentation in PDI

ddiroma:


ddiroma:
@pentaho/r2d2

engine/src/main/java/org/pentaho/di/trans/step/StepMeta.java
```diff
@@ -1144,4 +1145,18 @@ public String getAttribute( String groupName, String key ) {
   public boolean isMissing() {
     return this.stepMetaInterface instanceof MissingTrans;
   }
+
+  public boolean isDeprecated() {
```
bmorrise: Have you tried this fix on a transformation that has a lot of steps. I'm just curious as to how this method affects the performance of a transformation with a 100+ steps.

ddiroma: @bmorrise I moved the check for deprecation to the constructor so it is only called once and sets the class variable to be picked up in transpainter....so each time the trans graph is changed/hovered over we don't have to keep setting the same thing.

ddiroma:
@pentaho/r2d2 
@bmorrise 
see comments in response to Ben's concern

PR #5429, opened 2018-05-31
[BACKLOG-23831] Fixing a mis-configured localized error message in the UserRoleListDelegate

pentaho-amessier:


pentaho-amessier:
@pentaho/r2d2

PR #5431, opened 2018-06-01
[PDI-16614] - Data service in CSV File Input does not work with Lazy …

cravobranco:
…conversion active

* Fraction digits should be preserved when converting a string to a number.


pentaho-lmartins:
@cravobranco, can you take a look at this? We have a broken test with this PR.
Which do we need to fix, the test of the fix?

cravobranco:
This PR is no more relevant and will be closed.
The fix for PDI-16614 is now being handled and tracked here: https://github.com/pentaho/pentaho-kettle/pull/5505

PR #5432, opened 2018-06-01
[BACKLOG-23395] Revises MDI descriptions

pentaho-mracine:


pentaho-mracine:
@pentaho/r2d2 Shouldn't need to wait for wingman, just descriptions revision

PR #5433, opened 2018-06-04
[PDI-17300] 8.0 Execute Job fails when job name has .kjb Extension - …

pentaho-lmartins:
…6.1 Does Not

@bmorrise @mbatchelor @pamval


pamval:
@mbatchelor please review

PR #5435, opened 2018-06-04
[BACKLOG-23402] Adds MDI support to Execute Row SQL script step

pentaho-mracine:


pentaho-mracine:
@pentaho/r2d2 please review

PR #5445, opened 2018-06-06
[PDI-17325] Makes Kettle use JDBC metadata first and only in case of …

lucboudreau:
…failure use a prepared statement to lookup table and column metadata.


lucboudreau:
@pentaho/moseisley For a quick review. Would appreciate a second set of eyes. I tried this over oracle, mysql, pgsql, google bigquery.

PR #5448, opened 2018-06-06
[PDI-17248] MDI Set Variable is not replacing value after applying Se…

ppatricio:
…rvice Pack

As discussed with Jens, the child always get parameters from parent, even if the child doesn't have that same parameter in his context.  In that way, we need to create that same parameter in childNamedParams every time child doesn't have it, otherwise it gets lost, as in PDI-17248.

@pamval @mbatchelor @pentaho-lmartins

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/StepWithMappingMeta.java
+++ b/engine/src/main/java/org/pentaho/di/trans/StepWithMappingMeta.java
@@ -32,6 +32,7 @@ import org.pentaho.di.core.Const;
 import org.pentaho.di.core.ObjectLocationSpecificationMethod;
 import org.pentaho.di.core.exception.KettleException;
 import org.pentaho.di.core.logging.LogChannel;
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/pentaho/di/trans/StepWithMappingMeta.java
```diff
@@ -248,6 +249,15 @@ public static void activateParams( VariableSpace childVariableSpace, NamedParams
           // this is explicitly checked for up front
         }
       } else {
+        try {
+          childNamedParams.addParameterDefinition( key, "", "" );
+          childNamedParams.setParameterValue( key, value );
+        } catch ( DuplicateParamException e ) {
+          e.printStackTrace();
```
mbatchelor: No to e.printStackTrace() - these should use the standard logging mechanism, not print a stack trace

ppatricio:
Hi Mark, PR updated.

PR #5453, opened 2018-06-07
[PDI-17325] Makes Kettle use JDBC metadata first and only in case of …

lucboudreau:
…failure use a prepared statement to lookup table and column metadata.

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
+++ b/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
@@ -4539,6 +4539,16 @@ public class ValueMetaBase implements ValueMetaInterface {
     return originalScale;
   }
 
+  @Override
+  public int getOriginalNullable() {
+    return originalNullable;
+  }
+
+  @Override
+  public boolean getOriginalSigned() {
+    return originalSigned;
+  }
+
   /*
    * Original JDBC RecordSetMetaData
    *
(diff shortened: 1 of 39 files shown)


lucboudreau:
Please do not merge just yet. This is close to the final version, but there will be another PR for data-access coming.

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -4953,6 +4963,242 @@ protected void getOriginalColumnMetadata( ValueMetaInterface v, ResultSetMetaDat
...
+            // If we're dealing with PostgreSQL and double precision types
+            if ( databaseMeta.getDatabaseInterface() instanceof PostgreSQLDatabaseMeta && originalColumnType == java.sql.Types.DOUBLE
+                && precision >= 16 && length >= 16 ) {
+              precision = -1;
+              length = -1;
+            }
+
+            // MySQL: max resolution is double precision floating point (double)
+            // The (12,31) that is given back is not correct
+            if ( databaseMeta.getDatabaseInterface().isMySQLVariant() ) {
```
mbatchelor: Is there any way we can put this kind of stuff into the specific dialect ? I know this class is littered with these things already, but I'd rather not add more if we can avoid it.

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -4953,6 +4963,242 @@ protected void getOriginalColumnMetadata( ValueMetaInterface v, ResultSetMetaDat
...
+                valtype = ValueMetaInterface.TYPE_BIGNUMBER;
+              }
+            } else { // we have a precision: keep NUMBER or change to BIGNUMBER?
+              if ( length > 15 || precision > 15 ) {
+                valtype = ValueMetaInterface.TYPE_BIGNUMBER;
+              }
+            }
+          }
+
+          if ( databaseMeta.getDatabaseInterface() instanceof PostgreSQLDatabaseMeta
```
mbatchelor: Is there any way we can put this kind of stuff into the specific dialect ? I know this class is littered with these things already, but I'd rather not add more if we can avoid it.

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -4953,6 +4963,242 @@ protected void getOriginalColumnMetadata( ValueMetaInterface v, ResultSetMetaDat
...
+              || databaseMeta.getDatabaseInterface() instanceof GreenplumDatabaseMeta ) {
+            // undefined size => arbitrary precision
+            if ( originalColumnType == java.sql.Types.NUMERIC && length == 0 && precision == 0 ) {
+              valtype = ValueMetaInterface.TYPE_BIGNUMBER;
+              length = -1;
+              precision = -1;
+            }
+          }
+
+          if ( databaseMeta.getDatabaseInterface() instanceof OracleDatabaseMeta ) {
```
mbatchelor: As above

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -4953,6 +4963,242 @@ protected void getOriginalColumnMetadata( ValueMetaInterface v, ResultSetMetaDat
...
+            if ( precision == 0 && length == 38 ) {
+              valtype = ( (OracleDatabaseMeta) databaseMeta.getDatabaseInterface() )
```
mbatchelor: Ditto

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -4953,6 +4963,242 @@ protected void getOriginalColumnMetadata( ValueMetaInterface v, ResultSetMetaDat
...
+          break;
+
+        case java.sql.Types.DATE:
+          if ( databaseMeta.getDatabaseInterface() instanceof TeradataDatabaseMeta ) {
+            precision = 1;
+          }
+        case java.sql.Types.TIME:
+          valtype = ValueMetaInterface.TYPE_DATE;
+          //
+          if ( databaseMeta.getDatabaseInterface().isMySQLVariant() ) {
```
mbatchelor: zzzzzz... same

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -4953,6 +4963,242 @@ protected void getOriginalColumnMetadata( ValueMetaInterface v, ResultSetMetaDat
...
+        case java.sql.Types.BLOB:
+        case java.sql.Types.VARBINARY:
+        case java.sql.Types.LONGVARBINARY:
+          valtype = ValueMetaInterface.TYPE_BINARY;
+
+          if ( databaseMeta.isDisplaySizeTwiceThePrecision()
+              && ( 2 * originalPrecision ) == originalColumnDisplaySize ) {
+            // set the length for "CHAR(X) FOR BIT DATA"
+            length = originalPrecision;
+          } else if ( ( databaseMeta.getDatabaseInterface() instanceof OracleDatabaseMeta )
```
mbatchelor: Yet again

engine/src/main/java/org/pentaho/di/trans/steps/dimensionlookup/DimensionLookupMeta.java
```diff
@@ -1360,7 +1359,7 @@ private void checkLookup( List<CheckResultInterface> remarks, StepMeta stepinfo,
 
         if ( !Utils.isEmpty( tableName ) ) {
           String schemaTable = databaseMeta.getQuotedSchemaTableCombination( schemaName, tableName );
```
mbatchelor: Don't think we need this line anymore except for error condition. Maybe only run that code in case of error.

engine/src/main/java/org/pentaho/di/trans/steps/gettablenames/GetTableNames.java
```diff
@@ -244,7 +244,13 @@ public boolean processRow( StepMetaInterface smi, StepDataInterface sdi ) throws
         // handle simple primary key (one field)
         String sql =
           data.db
-            .getCreateTableStatement( tableName, data.db.getTableFields( tableName ), null, false, pk, true );
+            .getCreateTableStatement(
```
mbatchelor: don't think we need these all on their own line

engine/src/main/java/org/pentaho/di/trans/steps/sqlfileoutput/SQLFileOutputMeta.java
```diff
@@ -602,13 +602,13 @@ public void check( List<CheckResultInterface> remarks, TransMeta transMeta, Step
         if ( !Utils.isEmpty( tablename ) ) {
           String schemaTable = databaseMeta.getQuotedSchemaTableCombination( schemaName, tablename );
```
mbatchelor: Think we can remove this line or only define in case of error

engine/src/main/java/org/pentaho/di/trans/steps/tableoutput/TableOutputMeta.java
```diff
@@ -606,17 +606,18 @@ public void check( List<CheckResultInterface> remarks, TransMeta transMeta, Step
         remarks.add( cr );
 
         if ( !Utils.isEmpty( tableName ) ) {
+          String realSchemaName = db.environmentSubstitute( schemaName );
+          String realTableName = db.environmentSubstitute( tableName );
           String schemaTable =
```
mbatchelor: I think this can go or only define in case of error

ui/src/main/java/org/pentaho/di/ui/trans/steps/databaselookup/DatabaseLookupDialog.java
```diff
@@ -846,7 +845,10 @@ private void getlookup() {
           String schemaTable =
```
mbatchelor: This can go except in case of error

engine/src/main/java/org/pentaho/di/trans/steps/sqlfileoutput/SQLFileOutputMeta.java
lucboudreau: This one is used at least 2 times, if not 3. I'll leave this one here.

engine/src/main/java/org/pentaho/di/trans/steps/tableoutput/TableOutputMeta.java
lucboudreau: Same here. Will be touched 2-3 times at least.

lucboudreau:
@mbatchelor I looked into the issue of the huge switch statement. To do things right, we'll need to push each of those rules into the dialect, plus we need a way to have them run in the right order. Take 'length' for example. It is defined and overridden in a number of places. This refactoring is too risky compared to fixing this issue by using the adapted code which used to read ResultSets. I will document this in Jira correctly and we'll get back to it on its own.

PR #5454, opened 2018-06-07
[PDI-16638] - Fix multi-threading issue

mbatchelor:
* Uses read/write lock on ioMeta
* Changes variable name in BaseStep
* Stops everyone from being able to use ioMeta directly
* Fixes several inefficient loops that keep calling into the getter

Note - If you process this PR, please immediately process the PR for pentaho-r-plugin. PDI will fail to start up without the fix.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/step/BaseStepMeta.java
+++ b/engine/src/main/java/org/pentaho/di/trans/step/BaseStepMeta.java
@@ -22,14 +22,7 @@
 
 package org.pentaho.di.trans.step;
 
-import java.io.InputStream;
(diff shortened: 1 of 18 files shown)


sudhanshu-Tango:
@mbatchelor 

Some heads up about this would have been appreciated, changing a protected field name in a point release caught us off guard and broke many 3rd party plugins that were using it for multi stream output (stream schema merge and others)

mattcasters:
This is completely stupid folks.  Why break plugins at all customers and partners when you could have just make things compatible with a little bit of extra effort?

PR #5457, opened 2018-06-08
[BACKLOG-23276] Changes to add tooltip for deprecated steps

ddiroma:


ddiroma:
@pentaho/r2d2 

Please disregard the checkstyle violations...they have nothing to do with my changes.

ddiroma:
@mdamour1976

core/src/main/java/org/pentaho/di/core/plugins/BasePluginType.java
```diff
@@ -672,6 +675,8 @@ protected URLClassLoader createUrlClassLoader( URL jarFileUrl, ClassLoader class
 
   protected abstract String extractDocumentationUrl( java.lang.annotation.Annotation annotation );
 
+  protected abstract String extractSuggestedStep( java.lang.annotation.Annotation annotation );
```
bmorrise: Any reason why this can't be a default implementation so we don't have to override in every plugin that doesn't need it?

core/src/main/java/org/pentaho/di/core/plugins/Plugin.java
```diff
@@ -62,6 +62,7 @@
   private String documentationUrl;
   private String casesUrl;
   private String forumUrl;
+  private String suggestedStep;
```
bmorrise: It feels strange having this meta specific to steps generically added to Plugin. I wonder if there is a way to move this to be more specific to steps.

PR #5478, opened 2018-06-12
[BACKLOG-24014] adding guards against null and incorrect data from the step metrics table

ricksjames:
[BACKLOG-24014] adding guards against null and incorrect data from the step metrics table


ricksjames:
@pentaho/ackbar please review

PR #5480, opened 2018-06-12
[BACKLOG-23276] Fixes for deprecated icon size and implementation for…

ddiroma:
… when ktr is opened to pick up deprecated UI look and feel


ddiroma:
@pentaho/r2d2 
Not quite sure why so many broken tests, but they are not related to my changes.

PR #5485, opened 2018-06-13
[PPP-4108] Use of vulnerable component jackson-databind-2.8.8.jar, ja…

pentaho-lmartins:
…ckson-databind-2.9.2.jar, jackson-mapper-asl-1.9.2.jar CVE-2017-7525, CVE-2017-15095, CVE-2017-15095

This is one of a series of PRs. See further details in https://github.com/pentaho/maven-parent-poms/pull/66.

**Do not merge without checking the status of the JIRA case.**

### DIFF
--- a/assemblies/core/lib/pom.xml
+++ b/assemblies/core/lib/pom.xml
@@ -52,9 +52,6 @@
         <blueprints-core.version>2.6.0</blueprints-core.version>
         <commons-configuration.version>1.9</commons-configuration.version>
         <flexjson.version>2.1</flexjson.version>
-        <!-- jackson-databind depends on jackson-core and jackson-annotations
-        so that this artifacts have the same version to prevent compile errors -->
-        <fasterxml.jackson.core.version>2.9.5</fasterxml.jackson.core.version>
         <jersey-multipart.version>1.19.1</jersey-multipart.version>
         <mimepull.version>1.9.3</mimepull.version>
         <jsr311-api.version>1.1.1</jsr311-api.version>
@@ -621,39 +618,6 @@
             <groupId>net.sf.flexjson</groupId>
             <artifactId>flexjson</artifactId>
             <version>${flexjson.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>*</groupId>
-                    <artifactId>*</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-core</artifactId>
-            <version>${fasterxml.jackson.core.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>*</groupId>
-                    <artifactId>*</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-annotations</artifactId>
-            <version>${fasterxml.jackson.core.version}</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>*</groupId>
-                    <artifactId>*</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-            <version>${fasterxml.jackson.core.version}</version>
             <exclusions>
                 <exclusion>
                     <groupId>*</groupId>
--- a/assemblies/lib/pom.xml
+++ b/assemblies/lib/pom.xml
@@ -51,9 +51,6 @@
     <blueprints-core.version>2.6.0</blueprints-core.version>
     <commons-configuration.version>1.9</commons-configuration.version>
     <flexjson.version>2.1</flexjson.version>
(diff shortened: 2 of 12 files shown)


plugins/file-open-save/core/pom.xml
```diff
@@ -49,25 +49,10 @@
...
     <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-core</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-annotations</artifactId>
-    </dependency>
-    <dependency>
+      <!-- this is required by OSGi bean declared in src/main/resources-filtered/OSGI-INF/blueprint/blueprint.xml -->
```
pentaho-lmartins: @graimundo, is this OK?

assemblies/core/lib/pom.xml
```diff
@@ -628,39 +625,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <dependency>
```
graimundo: By removing jackson from here aren't we removing it from data-integration/lib? Doesn't this break non OSGi plugins that assume that this lib is available?
@hbfernandes ^

assemblies/lib/pom.xml
```diff
@@ -627,39 +624,6 @@
         </exclusion>
       </exclusions>
     </dependency>
-    <dependency>
```
graimundo: By removing jackson from here aren't we removing it from data-integration/lib? Doesn't this break non OSGi plugins that assume that this lib is available?
@hbfernandes ^

plugins/repositories/core/pom.xml
```diff
@@ -32,36 +31,6 @@
...
-      <version>${fasterxml.jackson.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-databind</artifactId>
-      <version>${fasterxml.jackson.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.jaxrs</groupId>
-      <artifactId>jackson-jaxrs-json-provider</artifactId>
```
graimundo: Repositories plugin has a json provider declared in its blueprint as such it seems this dependency shouldn't be removed. Please double check.
https://github.com/pentaho/pentaho-kettle/blob/master/plugins/repositories/core/src/main/resources-filtered/OSGI-INF/blueprint/beans.xml#L82

plugins/repositories/core/pom.xml
```diff
@@ -33,34 +32,9 @@
...
-      <artifactId>jackson-annotations</artifactId>
-      <version>${fasterxml.jackson.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-databind</artifactId>
-      <version>${fasterxml.jackson.version}</version>
-    </dependency>
-    <dependency>
+      <!-- this is required by OSGi bean declared in src/main/resources-filtered/OSGI-INF/blueprint/beans.xml -->
```
pentaho-lmartins: @graimundo, is this OK?

pentaho-lmartins: `jackson-jaxrs-json-provider` will also bring `jackson-jaxrs-base` and `jackson-module-jaxb-annotations` as transitive dependencies. As far as I can see, modules from `com.fasterxml.jackson.core` are not required.

assemblies/core/lib/pom.xml
pentaho-lmartins: In the last test stack build we have `jackson-annotations-2.9.5.jar`, `jackson-databind-2.9.5.jar` and `jackson-core-2.9.5.jar` in _data-integration/lib_. We have these three declared as dependencies of **kettle-engine**. Do we also want to make these as explicit dependencies of **assemblies/core/lib**? @graimundo @hbfernandes

assemblies/lib/pom.xml
pentaho-lmartins: In the last test stack build we have `jackson-annotations-2.9.5.jar`, `jackson-databind-2.9.5.jar` and `jackson-core-2.9.5.jar` in _data-integration/lib_. We have these three declared as dependencies of **kettle-engine**. Do we also want to make these as explicit dependencies of **assemblies/lib**?
Isn't it redundant to have these as dependencies of **assemblies/lib** and **assemblies/core/lib**?
@graimundo @hbfernandes

assemblies/core/lib/pom.xml
graimundo: Same comment as: https://github.com/pentaho/pentaho-platform/pull/4161#discussion_r198817326

@pentaho-lmartins I think that for the scope of this story it is ok to remove the explicit dependencies and let them come in transitively. That means we are assuming that the lib contains the necessary libs for the Pentaho code base and that we are not putting anything there just for the sake of third parties. So lets go with this solution to unblock the work and create that followup thread that we talked about. If necessary, as a result of that discussion we can re add the dependencies explicitly later on.

assemblies/lib/pom.xml
graimundo: https://github.com/pentaho/pentaho-kettle/pull/5485#discussion_r198820920

PR #5486, opened 2018-06-14
[CLEANUP] Fixing errors preventing ErrorProne compilation, + checkestyle

mkambol:
https://github.com/google/error-prone

Specifically, these 5 categories of errors:

[NullTernary] This conditional expression may evaluate to null, which will result in an NPE when the result is unboxed.
    (see http://errorprone.info/bugpattern/NullTernary)

[ArrayToString] Calling toString on an array does not provide useful information
    (see http://errorprone.info/bugpattern/ArrayToString)

[SizeGreaterThanOrEqualsZero] Comparison of a size >= 0 is always true, did you intend to check for non-emptiness?
    (see http://errorprone.info/bugpattern/SizeGreaterThanOrEqualsZero)

[InfiniteRecursion] This method always recurses, and will cause a StackOverflowError
    (see http://errorprone.info/bugpattern/InfiniteRecursion)

[DeadException] Exception created but not thrown
    (see http://errorprone.info/bugpattern/DeadException)

### DIFF
--- a/engine/src/main/java/org/pentaho/di/job/entry/validator/JobEntryValidatorUtils.java
+++ b/engine/src/main/java/org/pentaho/di/job/entry/validator/JobEntryValidatorUtils.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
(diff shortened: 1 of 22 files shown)


mkambol:
@pentaho/ackbar

PR #5492, opened 2018-06-18
[PDI-16387] Running a Job for Every Input Row, sub job does not reset…

ppatricio:
… variables or parameters with every execution

In this case i had to create a new map to guarantee the logic of reading variables from properties files. The parameters are also considered and they have a specific behavior.
Some properties files are also included for testing.

@pamval @mbatchelor @pentaho-lmartins @bmorrise

### DIFF
--- a/engine/src/main/java/org/pentaho/di/job/entry/JobEntryBase.java
+++ b/engine/src/main/java/org/pentaho/di/job/entry/JobEntryBase.java
@@ -27,6 +27,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.pentaho.di.cluster.SlaveServer;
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/pentaho/di/job/entry/JobEntryBase.java
```diff
@@ -1520,4 +1524,28 @@ public JobMeta getParentJobMeta() {
   public void setParentJobMeta( JobMeta parentJobMeta ) {
     this.parentJobMeta = parentJobMeta;
   }
+
+  /**
+   * Gets a Map of variables set in EntryStepSetVariables
+   *
+   * @return a map of variable names and values
+   */
+  public Map<String, String> getEntryStepSetVariablesMap() {
```
mbatchelor: We should not expose the map outside the class. Please make this protected instead of public.

ppatricio:
PR updated

PR #5494, opened 2018-06-19
[PDI-17278] Repository Job Fails Using Variable to Run Local Job

pentaho-lmartins:
Maybe I went too far with the code changes, but there was some logic to handle use cases we should not be supporting. The new code makes much more sense to me while still running all possible configurations of `JobEntryJob` created prior to 7.0, namely, when it was possible to specify a child job by repository name, repository object id and file system filename.

Please review and let me know if you think we should be supporting any other use cases.
See use cases in the comment blocks [here](https://github.com/pentaho-lmartins/pentaho-kettle/blob/d24e413b66c7dab188d83ee8aab054ad6ac6949c/engine/src/main/java/org/pentaho/di/job/entries/job/JobEntryJob.java#L369-L377) and [here](https://github.com/pentaho-lmartins/pentaho-kettle/blob/d24e413b66c7dab188d83ee8aab054ad6ac6949c/engine/src/main/java/org/pentaho/di/job/entries/job/JobEntryJob.java#L466-L473).

There are 20 test cases in `JobEntryJobTest` to cover all the options I could think of.

@mbatchelor @pamval @bmorrise @pmalves

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/util/StringUtil.java
+++ b/core/src/main/java/org/pentaho/di/core/util/StringUtil.java
@@ -613,4 +613,47 @@ public class StringUtil {
       return obj.toString().toLowerCase();
     }
   }
+
+  /**
+   * Removes all instances of the specified character from the start of the given {@link String}.
+   *
+   * @param source the {@link String} to trim
+   * @param c the character to remove from the {@link String}
+   * @return a new string with all instances of the specified character removed from the start
+   */
+  public static String trimStart( final String source, char c ) {
+    if ( source == null ) {
+      return null;
+    }
+
+    int length = source.length();
+    int index = 0;
+
+    while ( index < length && source.charAt( index ) == c ) {
+      index++;
+    }
+
+    return source.substring( index );
+  }
+
+  /**
+   * Removes all instances of the specified character from the end of the given {@link String}.
+   *
+   * @param source the {@link String} to trim
+   * @param c the character to remove from the {@link String}
+   * @return a new string with all instances of the specified character removed from the end
+   */
+  public static String trimEnd( final String source, char c ) {
+    if ( source == null ) {
+      return null;
+    }
+
+    int index = source.length();
+
+    while ( index > 0 && source.charAt( index - 1 ) == c ) {
+      index--;
+    }
+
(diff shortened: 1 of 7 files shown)


pentaho-lmartins:
Updated PR to fix check style issues.

core/src/main/java/org/pentaho/di/core/util/StringUtil.java
```diff
@@ -613,4 +613,46 @@ public static String safeToLowerCase( final Object obj ) {
...
   }
+
+  /**
+   * Removes all instances of the specified character from the start of the given {@link String}.
+   *
+   * @param s the {@link String} to trim
+   * @param c the character to remove from the {@link String}
+   * @return a new string with all instances of the specified character removed from the start
+   */
+  public static String trimStart( final String s, char c ) {
```
mbatchelor: Suggest you pattern this after Const.ltrim - this is a lot of substring operations which create temporary strings and doesn't perform well at all. Something more like this:

public static String ltrim( String source, char c ) {
    if ( source == null ) {
      return null;
    }
    int from = 0;
    int len = source.length();
    while ( from < len && ( source.charAt( from ) == c ) ) {
      from++;
    }
    return source.substring( from );
  }

core/src/main/java/org/pentaho/di/core/util/StringUtil.java
```diff
@@ -613,4 +613,46 @@ public static String safeToLowerCase( final Object obj ) {
...
+  }
+
+  /**
+   * Removes all instances of the specified character from the end of the given {@link String}.
+   *
+   * @param s the {@link String} to trim
+   * @param c the character to remove from the {@link String}
+   * @return a new string with all instances of the specified character removed from the end
+   */
+  public static String trimEnd( final String s, char c ) {
```
mbatchelor: Same criticism as above. Something more like this performs much better and doesn't create tons of temporary strings.

  public static String rtrim( String source, char c ) {
    if ( source == null ) {
      return null;
    }
    int max = source.length();
    while ( max > 0 && ( source.charAt( max - 1 ) == c ) ) {
      max--;
    }
    return source.substring( 0, max );
  }

PR #5503, opened 2018-06-21
[BACKLOG-23995] Split every..row in Text File Output is not evenly sp…

angelorod:
…litting rows at file creation.


engine/src/main/java/org/pentaho/di/trans/steps/textfileoutput/TextFileOutputData.java
```diff
@@ -86,7 +86,10 @@ public void setIndex( int index ) {
 
     @Override
     public int compareTo( FileStreamsKey o ) {
-      return fileName.compareTo( o.getFileName() );
+      if ( o.fileName.equals( fileName ) ) {
+        return 0;
+      }
+      return index < o.index ? -1 : index > o.index ? +1 : 0;
```
jgminder: Could just do this after the if statement.
Integer idx = index;
return idx.compareTo( o.index );

PR #5504, opened 2018-06-21
[PDI-17024] Step name called "User Defined Java Expression 4" causes …

smmribeiro:
…in error

@pentaho-lmartins

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/step/StepMetaInterface.java
+++ b/engine/src/main/java/org/pentaho/di/trans/step/StepMetaInterface.java
@@ -715,7 +715,7 @@ public interface StepMetaInterface {
   Object loadReferencedObject( int index, Repository rep, IMetaStore metaStore, VariableSpace space ) throws KettleException;
 
   /**
-   * Action remove hop from this step
+   * Action remove hop exiting this step
    * @return step was changed
    */
   default boolean cleanAfterHopFromRemove() {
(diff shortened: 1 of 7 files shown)


smmribeiro:
@mbatchelor @pamval @bmorrise please review.

mbatchelor:
@smmribeiro - Please re-base

smmribeiro:
@mbatchelor, during the development for this issue, I saw that the **org.pentaho.di.trans.step.StepMetaInterface#cleanAfterHopFromRemove** method, as it had no parameters, was in need of a refactoring (while debugging, I saw it clearing things that it shouldn't in org.pentaho.di.trans.steps.transexecutor.TransExecutorMeta).
As a matter of fact, I was going to create a JIRA for this. However, someone saw this behaviour and created [BACKLOG-22673](https://jira.pentaho.com/browse/BACKLOG-22673) and as its related commit was merged before mine, we've a conflict.
The approach was similar to mine, however I pass a StepMeta object:
```
default boolean cleanAfterHopToRemove( StepMeta fromStep ) {
  return false;
}
```
The other implementation uses a String (with the step name):
```
default boolean cleanAfterHopFromRemove( String toStepName ) {
  return false;
}
```

I feel that we should be homogeneous and coherent, and so, we should use the same type in both.
My feeling is that, while passing the step name is enough, passing a StepMeta object provides more flexibility for the future.
So, when I rebase my commit, should I:
1. let the two methods with their parameters
2. change my method and use “String fromStepName”
3. change the other method and use “StepMeta toStep”

mbatchelor:
1- Leave the one with the string parameter, have it resolve the step meta, and call a new method with the step meta as a parameter.
2- Deprecate the one receiving a string

smmribeiro:
@mbatchelor, I've talked with @ddiroma: as the method was so recently created and had no other references in code, instead of deprecating the String parameter version, I've eliminated it and made the needed changes where it was invoked.

PR #5518, opened 2018-06-25
[BACKLOG-19467] should not use classes from javafx

kurtwalker:


kurtwalker:
@pentaho/ackbar please review

PR #5519, opened 2018-06-25
[BACKLOG-22673] When editing some hops of Job Executor &  one of Tran…

ddiroma:
…sformation Executor appropriate xml tags related to hops are cleared

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/step/StepMetaInterface.java
+++ b/engine/src/main/java/org/pentaho/di/trans/step/StepMetaInterface.java
@@ -723,6 +723,15 @@ public interface StepMetaInterface {
   }
 
   /**
+   * Action remove hop from this step
+   * @param toStepName - String of step name of the "to" step of the hop
+   * @return step was changed
+   */
+  default boolean cleanAfterHopFromRemove( String toStepName ) {
+    return false;
+  }
+
+  /**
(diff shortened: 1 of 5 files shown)


ddiroma:
@pentaho/r2d2

engine/src/main/java/org/pentaho/di/trans/step/StepMetaInterface.java
```diff
@@ -722,6 +722,15 @@ default boolean cleanAfterHopFromRemove() {
     return false;
   }
 
+  /**
+   * Action remove hop from this step
+   * @param toStepName - String of step name of the "to" step of the hop
+   * @return step was changed
+   */
+  default boolean cleanAfterHopFromRemove( String toStepName ) {
```
pentaho-amessier: Do we still need the no-arg version of this method? If not, and it's too dangerous to remove it at this point, would it make sense to mark it deprecated?

ddiroma: not quite sure if anything else uses it...non-kettle related. doesn't hurt to leave it if it's not used

pentaho-amessier: I searched all pentaho projects on github and didn't find any references. I understand removing it might be risky, just wanted to point it out. If you're not sure, it's fine to leave it.

PR #5523, opened 2018-06-26
[BACKLOG-23995] Split every..row in Text File Output is not evenly sp…

angelorod:
…litting rows at file creation.


pentaho-lmartins:
This PR was reverted from 8.1 and 8.1.0.2 by: #5595 and #5596.

PR #5525, opened 2018-06-27
[SP-4417] - Backport of PDI-16614 - Data service in CSV File Input do…

cravobranco:
…es not work with Lazy conversion active (7.1 Suite)

* Backport of PDI-16614 - Data service in CSV File Input does not work with Lazy conversion active (7.1 Suite)


cravobranco:
@pentaho-lmartins , @ricardosilva88

PR #5533, opened 2018-06-27
[BACKLOG-24042] also set prot p on ftps upload

tgf:


tgf:

PR #5536, opened 2018-06-28
[SP-4401] Backport of PDI-17203 - XML Output step: Null pointer excep…

ssamora:
…tion when processing fields with null values (7.1 Suite)
@pentaho-lmartins


pentaho-lmartins:
Commit message and PR title is not OK.

PR #5538, opened 2018-06-28
[BACKLOG-23973] AEL - Inside Unique Row step,Ktr hangs if the Redirect duplicate row is unchecked but there is error handling step

baudekin:
Fixed https://jira.pentaho.com/browse/BACKLOG-23973 for both Kettle and AEL by removing the error hops when the "Reject Duplicate Rows" is unselected.


baudekin:
Fixed copy right year.

PR #5563, opened 2018-07-05
[PDI-17293] - Pentaho Reporting Output: Paged HTML option throws erro…

cravobranco:
…r when selected as output processor

* Added Java7 NIO's API to deal with paths and filenames.


cravobranco:
@pamval , @pentaho-lmartins , @mbatchelor , @lucboudreau please review and merge if it's all ok.

mbatchelor:
Bah ... this class is a disaster. I don't see a reasonable way to create a test case for this without completely refactoring the thing. Accepting the change in spite of the lack of a test case because in it's current condition, it's unsuitable for a test case.

PR #5567, opened 2018-07-06
[PDI-17270] Sort rows : Cannot sort Internet Address Types

smmribeiro:
@pentaho-lmartins

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/row/ValueMetaInterface.java
+++ b/core/src/main/java/org/pentaho/di/core/row/ValueMetaInterface.java
@@ -1279,8 +1279,6 @@ public interface ValueMetaInterface extends Cloneable {
    *
    * @param databaseMeta
    *          the database metadata to reference capabilities and so on.
-   * @param lazyConversion
-   *          use lazy conversion
    * @param rs
    *          A ResultSet from getColumns, positioned correctly on a column to read.
    */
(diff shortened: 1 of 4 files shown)


smmribeiro:
@mbatchelor @pamval @bmorrise please review.

smmribeiro:
@mbatchelor:

Then, instead of working with exceptions, it should be better to call directly getBigNumber and return the ".doubleValue()" of the returned value, don't you think?

One more thing: won't the "address.getAddress()" invocation return a 4-position array for an IPV4 address and an 8-position array for an IPV6 address?
Then in getInteger() the code "if ( addr.length > 8 )" will never occur!
Shall I change it to ">4"?

mbatchelor:
1- I'd usually say that we only want to resort to the BigDecimal classes when necessary I'd say. But I have a correction in my thinking down below
2- IPV6 is 128bits and returns 16 bytes. So > 8 tells us it's too long for a Long ( 64 bits ). So the  exception (>8) will be thrown on every IPV6. 
3- So a double is 64bits as well - so trying to represent an IPV6 address as a double or a long is silly. Furthermore, using a double for an IPV6 address is awful because of the loss of precision inherent in IEEE floating point numbers. 

However - I'm wrong insofar as we shouldn't try to catch the exception if we're calling getNumber with an IPV6 address. We need that to fail because an IPV6 address doesn't translate to a double. So my observation in this instance was wrong.

PR #5569, opened 2018-07-06
[BACKLOG-22946] Override the getFileName method to return the path to…

peterrinehart:
… the subtransformation if filename is null.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/streaming/common/BaseStreamStepMeta.java
+++ b/engine/src/main/java/org/pentaho/di/trans/streaming/common/BaseStreamStepMeta.java
@@ -22,6 +22,7 @@
 
 package org.pentaho.di.trans.streaming.common;
 
+import com.google.common.base.Strings;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/pentaho/di/trans/streaming/common/BaseStreamStepMeta.java
```diff
@@ -148,6 +148,15 @@ public void check( List<CheckResultInterface> remarks, TransMeta transMeta,
     }
   }
 
+  @Override
+  public String getFileName() {
+    String returnString = this.fileName;
+    if ( ( null == returnString ) || returnString.isEmpty() ) {
```
kurtwalker: We've got Guava available to us, so you can use Strings.isNullOrEmpty().   Also, I'd prefer a ternary operator, but that's just personal style.  
return Strings.isNullOrEmpty(fileName) ? getTransformationPath() : fileName

PR #5581, opened 2018-07-10
[PDI-17421] The parent parameter still overwrites the child parameter…

ppatricio:
… even if the option Pass Parameter values to sub transformation is not checked

@pentaho-lmartins @mbatchelor @bmorrise

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/StepWithMappingMeta.java
+++ b/engine/src/main/java/org/pentaho/di/trans/StepWithMappingMeta.java
@@ -219,6 +219,11 @@ public abstract class StepWithMappingMeta extends BaseSerializingMeta implements
 
   public static void activateParams( VariableSpace childVariableSpace, NamedParams childNamedParams, VariableSpace parent, String[] listParameters,
                                      String[] mappingVariables, String[] inputFields ) {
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/pentaho/di/trans/StepWithMappingMeta.java
```diff
@@ -269,6 +269,62 @@ public static void activateParams( VariableSpace childVariableSpace, NamedParams
     childNamedParams.activateParameters();
   }
 
+// created to pass also the isPassingAllParameters checkbox selection
```
mbatchelor: Please NEVER copy a whole method just to add a single variable to an else statement. In fact, please don't ever copy a whole method to just make a trivially different new method. 

For this case:
1- Move old logic into method with the new method signature
2- In the old method, call new method signature with hard-coded boolean that represents what the old behavior did
3- Make your change to the method with the new method signature.
4- Make sure that all references to this method in PDI proper (and possibly even in our whole code line) *only* call the method with the new signature

pentaho-lmartins: 5- Deprecate the old method? @mbatchelor

ppatricio:
Hi Marc, thanks for the comments.
PR updated.

PR #5583, opened 2018-07-10
[PDI-17398] Scheduling a Job More than Once Causes the Logs to Displa…

pentaho-lmartins:
…y Incorrectly

Adds a constructor to support the creation of a new `Job` with a given `containerObjectId` which can be the Carte object Id. This needs to be set before creating the `LogChannel` so different schedules of the same job can have a different log channel Id and logs from both executions don't get concatenated.

This needs to be merged before https://github.com/pentaho/pdi-platform-plugin/pull/61.

@bmorrise @mbatchelor @graimundo @pedrofvteixeira


pentaho-lmartins:
@ppatricio @ricardosilva88 @cravobranco

PR #5594, opened 2018-07-16
[PDI-16909] Remove a schedule using the REST API using HTTP DELETE me…

pentaho-lmartins:
…thod

- Changes REST Client so the DELETE verb doesn't send data in the request body
- Changes the verbs drop-down behavior so the DELETE option disables the body field drop-down
- The application type is also disabled for all non entity enclosing verbs
- Included test cases to cover changes

Please see JIRA case for details on this change.

This fix is now complete with https://github.com/pentaho/pentaho-platform/pull/4188

@mbatchelor @bmorrise


pentaho-lmartins:
@mbatchelor, @bmorrise, can you please take a look at this?

PR #5622, opened 2018-07-23
[SP-4478] Backport of BACKLOG-22673 - When editing some hops of Job E…

smmribeiro:
…xecutor &  one of Transformation Executor appropriate xml tags related to hops are cleared (7.1 Suite)

@pentaho-lmartins @cravobranco @ricardosilva88

### DIFF
--- a/engine/src/org/pentaho/di/trans/step/StepMetaInterface.java
+++ b/engine/src/org/pentaho/di/trans/step/StepMetaInterface.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Pentaho : http://www.pentaho.com
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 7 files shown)


ricardosilva88:
From my analysis, the base case had 2 PRs:
https://github.com/pentaho/pentaho-kettle/pull/5219/files
https://github.com/pentaho/pentaho-kettle/pull/5519/files

Please verify, in this PR your are back porting the 2nd case only

smmribeiro:
@ricardosilva88: I've added the missing code.

engine/src/org/pentaho/di/trans/step/StepMetaInterface.java
```diff
@@ -166,7 +166,7 @@
   /**
    * Set default values
    */
-  public void setDefault();
+  void setDefault();
```
ricardosilva88: Changing public access modifier to default it's making all the methods access visibility more restrictive. If some other code part or lib is depending from these methods and if it is outside of the package it'll cause problems. I would recommend to keep the modifier as it is.

ricardosilva88: It applies to all methods that you changed the modifier in this class, ty

PR #5628, opened 2018-07-23
[SP-4442] - Backport of PDI-17378 - Different column renaming in mapp…

cravobranco:
…ing step, PDI7.0 vs PDI8.1 (8.1 Suite)

* Backport of PDI-17378 - Different column renaming in mapping step, PDI7.0 vs PDI8.1 (8.1 Suite)

@pentaho-lmartins , @ricardosilva88


ricardosilva88:
Consistent with the base case

PR #5633, opened 2018-07-23
[PPP-4108] Use of vulnerable component jackson-databind-2.8.8.jar, ja…

pentaho-lmartins:
…ckson-databind-2.9.2.jar, jackson-mapper-asl-1.9.2.jar CVE-2017-7525, CVE-2017-15095, CVE-2017-15095

Due to a regression found in DET - Publish Model, we need to add **jackson-jaxrs-json-provider** and **jackson-jaxrs-base** back to `data-integration/lib`. These libraries were there before PPP-4108.

Since these are not transitive dependencies of `kettle-engine`, I figure this is the best place to add them. In #5485 these were removed from: https://github.com/pentaho/pentaho-kettle/pull/5485/files#diff-2648864198c6f1a9221ce8ac97fe85bcL686

@graimundo @pamval

### DIFF
--- a/assemblies/lib/pom.xml
+++ b/assemblies/lib/pom.xml
@@ -544,6 +544,16 @@
 
     <!-- Other third-party dependencies -->
     <dependency>
+      <groupId>com.fasterxml.jackson.jaxrs</groupId>
+      <artifactId>jackson-jaxrs-json-provider</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>com.fasterxml.jackson.module</groupId>
+          <artifactId>jackson-module-jaxb-annotations</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
       <groupId>log4j</groupId>
       <artifactId>apache-log4j-extras</artifactId>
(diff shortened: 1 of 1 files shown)


pentaho-lmartins:
These are transitive dependencies of `jersey-client` needed by https://github.com/pentaho/pentaho-data-refinery, which is an OSGi bundle. And rightly so, these dependencies are deployed to Karaf. However, there is a hack in the code and they're being loaded from `data-integration/lib`.

Hack: https://github.com/pentaho/pentaho-data-refinery/blob/006a6e48829875ae81dd55b5aec02f7df451da0f/src/main/java/org/pentaho/di/core/refinery/publish/agilebi/ModelServerAction.java#L57-L88

pentaho-lmartins:
@graimundo, hold on... let me check something....

pentaho-lmartins:
I can't find another way to fix this. I'm sure there is one though. @graimundo

graimundo:
We've already talked about it in person but leaving a comment here for the record.
It seems that the solution is to refactor the code to drop the dependency on the jersey implementation. We can probably use instead a client from the jax-rs spec.

PR #5639, opened 2018-07-24
[PDI-17295] Switch / Case step is not working as expected when input …

ppatricio:
…data is of Binary type.
There was a problem when validating binary values to match in the SwitchCase Step. When testing if a binary was on our "SwitchCase values to match" map, a byte[] was inserted as key in the SwitchCase Map and using byte[] as key, results on bad comparison results. So we need to convert it to it's specific/unique hashCode and using it instead, as comparator.

@ricardosilva88  @pentaho-lmartins @mbatchelor


ppatricio:
PR Updated

PR #5657, opened 2018-07-27
[BACKLOG-24315] Add SSL capability to ActiveMQ JMS producer and consu…

peterrinehart:
…mer.

### DIFF
--- a/plugins/streaming/impls/jms/src/main/java/org/pentaho/di/trans/step/jms/JmsDelegate.java
+++ b/plugins/streaming/impls/jms/src/main/java/org/pentaho/di/trans/step/jms/JmsDelegate.java
@@ -90,9 +90,15 @@ public class JmsDelegate {
 
   @Injection ( name = "SSL_CONTEXT_ALGORITHM", group = "SSL_GROUP" ) public String sslContextAlgorithm = "";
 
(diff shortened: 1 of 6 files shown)


plugins/streaming/impls/jms/src/main/java/org/pentaho/di/trans/step/jms/JmsDelegate.java
```diff
@@ -90,10 +90,16 @@
 
   @Injection ( name = "SSL_CONTEXT_ALGORITHM", group = "SSL_GROUP" ) public String sslContextAlgorithm = "";
 
-  @Injection ( name = "IBM_SSL_CIPHERSUITE", group = "SSL_GROUP" ) public String ibmSslCipherSuite = "";
+  @Injection ( name = "SSL_CIPHERSUITE", group = "SSL_GROUP" ) public String sslCipherSuite = "";
```
kurtwalker: We can't change this because it will break existing ktr's

kurtwalker: oh, wait, this didn't exist until last sprint.  so should be ok.  nevermind

peterrinehart: Correct; I added all those parameters last sprint, so the only things that break are some of our test files.

PR #5673, opened 2018-08-08
[PDI-17366] CSV input : BigNumber datatype does not work without prec…

LeonardoCoelho71950:
…ision

Trying wrongly to cast a Number into a BigDecimal.

@pamval @duarteteixeira 
@pentaho-lmartins @ricardosilva88


pamval:
@mbatchelor @bmorrise can you take a look at this one please?

PR #5685, opened 2018-08-10
[PDI-13055] CSV File Input step: Get Fields incorrectly reverse field…

ppatricio:
…s length
This is a special characters related issue.
Since my suggested workaround could be massive in some cases, when we have to deal with multiple tables with multiple rows, drop them all and recreate again could be a not easy and danger solution. So I provide a possible solution to ensure we deal differently when we have to deal with special characters in both ways, bytes and chars.
In that way this PR creates a new compatibility flag where users can now choose from reading a string length by char or byte before inserting it to a database, and if that Const is not defined, then it uses a new configuration at "DatabaseInterface" in which Oracle (in "OracleDatabaseMeta") is defined by default with bytes. The Const value **overwrites** the default value in "OracleDatabaseMeta".
This means that if configured right, Kettle i'll not stop executing, neither giving the ORA error, but instead use the actual logic in "ValueMetaBase", providing now the already existing message:
 ValueMetaBase - Truncating 2 symbols of original message in 'CONTACTFIRSTNAME' field

@pentaho-lmartins @mbatchelor @bmorrise

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
+++ b/core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
@@ -1200,4 +1200,11 @@ public interface DatabaseInterface extends Cloneable {
     return "";
   }
 
+  /**
+   * @return true if database supports the default char length validation
+   */
+  default boolean supportsGetLengthByChar() {
+    return true;
+  }
+
 }
--- a/core/src/main/java/org/pentaho/di/core/database/DatabaseMeta.java
+++ b/core/src/main/java/org/pentaho/di/core/database/DatabaseMeta.java
@@ -2930,6 +2930,13 @@ public class DatabaseMeta extends SharedObjectBase implements Cloneable, XMLInte
   }
 
   /**
+   * @return true if database supports the default char length validation
+   */
+  public boolean supportsGetLengthByChar() {
+    return databaseInterface.supportsGetLengthByChar();
+  }
+
+  /**
    * Get the SQL to insert a new empty unknown record in a dimension.
    *
    * @param schemaTable
--- a/core/src/main/java/org/pentaho/di/core/database/OracleDatabaseMeta.java
+++ b/core/src/main/java/org/pentaho/di/core/database/OracleDatabaseMeta.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
@@ -677,5 +677,9 @@ public class OracleDatabaseMeta extends BaseDatabaseMeta implements DatabaseInte
    */
   public void setStrictBigNumberInterpretation( boolean strictBigNumberInterpretation ) {
     getAttributes().setProperty( STRICT_BIGNUMBER_INTERPRETATION, strictBigNumberInterpretation ? "Y" : "N" );
+  }
+
+  @Override public boolean supportsGetLengthByChar() {
+    return false;
   }
 }
--- a/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
+++ b/core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
@@ -526,6 +526,51 @@ public class ValueMetaBase implements ValueMetaInterface {
   }
 
   /**
+   *
+   * @return get length in bytes (dealing with special characters)
+   */
+  int getLengthInBytes( String value ) {
+    int total = 0;
+    try {
+      for ( int i = 0; i < value.length(); i++ ) {
+        String c = new String( new char[] {value.charAt( i )} );
+        int length = c.getBytes( "UTF-8" ).length;
+        total += length;
+      }
+      return total;
+
+    } catch ( UnsupportedEncodingException e ) {
+      return value.length();
+    }
+  }
+
+  /**
+   *
+   * @return get length in bytes (if const is defined or database only supports char validation ) or chars (default
+   * except for Oracle Databases)
+   */
+
+  int getLengthInCharOrBytes( String string, DatabaseMeta databaseMeta )  {
+    String val = EnvUtil.getSystemProperty( Const.KETTLE_COMPATIBILITY_LENGTH_VALIDATION_USES_BYTES );
+    int total = string.length();
+    if ( val != null ) {
+      boolean usesBytes = convertStringToBoolean( Const.NVL( val, "N" ) );
+      // if const is defined and it's true, then use bytes
+      if ( usesBytes ) {
+        total = getLengthInBytes( string );
+      }
+    } else {
+      // if const is not defined then check the default databaseConfiguration
+      boolean databaseSupportsChar = databaseMeta.supportsGetLengthByChar();
+      if ( !databaseSupportsChar ) {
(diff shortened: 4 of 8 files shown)


core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
```diff
@@ -1200,4 +1200,11 @@ default String getUnsupportedTableOutputMessage() {
     return "";
   }
 
+  /**
+   * @return true if database supports the default char length validation
+   */
+  default boolean supportsGetLengthByChar() {
```
pentaho-lmartins: Please change this method to: `lengthSemanticsUsesChars()`

core/src/main/java/org/pentaho/di/core/database/DatabaseMeta.java
```diff
@@ -2929,6 +2929,13 @@ public boolean supportsErrorHandlingOnBatchUpdates() {
     return databaseInterface.supportsErrorHandlingOnBatchUpdates();
   }
 
+  /**
+   * @return true if database supports the default char length validation
+   */
+  public boolean supportsGetLengthByChar() {
```
pentaho-lmartins: Same as above.

core/src/main/java/org/pentaho/di/core/database/OracleDatabaseMeta.java
```diff
@@ -678,4 +678,8 @@ public boolean strictBigNumberInterpretation() {
   public void setStrictBigNumberInterpretation( boolean strictBigNumberInterpretation ) {
     getAttributes().setProperty( STRICT_BIGNUMBER_INTERPRETATION, strictBigNumberInterpretation ? "Y" : "N" );
   }
+
+  @Override public boolean supportsGetLengthByChar() {
+    return false;
```
pentaho-lmartins: Same as above + this is where we need to read `ORACLEDB_LENGTH_SEMANTICS_USES_CHARS`. By default this method will return `false` but if the kettle property is defined it will return it's value.

This will then match the behavior of the Oracle DB property which is set to **bytes** by default. If for some reason the value has been changed to **chars**, then the user can set the kettle property to **Y**.

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -525,6 +525,51 @@ boolean isLengthInvalidOrZero() {
...
+  }
+
+  /**
+   *
+   * @return get length in bytes (if const is defined or database only supports char validation ) or chars (default
+   * except for Oracle Databases)
+   */
+
+  int getLengthInCharOrBytes( String string, DatabaseMeta databaseMeta )  {
+    String val = EnvUtil.getSystemProperty( Const.KETTLE_COMPATIBILITY_LENGTH_VALIDATION_USES_BYTES );
```
pentaho-lmartins: In this method we just call `lengthSemanticsUsesChars() ` on the given `databaseMeta` and decide if we just return the `string.length()` or the length in bytes. A single ternary operator will do.

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -525,6 +525,51 @@ boolean isLengthInvalidOrZero() {
     return this.length < 1;
   }
 
+  /**
+   *
+   * @return get length in bytes (dealing with special characters)
+   */
+  int getLengthInBytes( String value ) {
+    int total = 0;
```
pentaho-lmartins: Why looping through each character in the string? The length in bytes can be found with just `value.getBytes( "UTF-8" ).length`. I can't see why this needs to be like this.

core/src/main/java/org/pentaho/di/core/row/value/ValueMetaBase.java
```diff
@@ -5357,7 +5402,8 @@ public void setPreparedStatementValue( DatabaseMeta databaseMeta, PreparedStatem
               setLength( databaseMeta.getMaxTextFieldLength() );
             }
             String string = getString( data );
-            int len = string.length();
+            //int len = string.length(); - changed to support length in char or bytes
+            int len = getLengthInCharOrBytes( string, databaseMeta );
```
pentaho-lmartins: You can place the ternary operator here. There is no need for the two methods above.

core/src/main/java/org/pentaho/di/core/util/StringEvaluator.java
```diff
@@ -255,6 +255,7 @@ private void evaluateLength( String value ) {
     }
   }
 
+
```
pentaho-lmartins: Why are you updating this file just to add a blank line?
Please revert this file.

core/src/test/java/org/pentaho/di/core/util/StringEvaluatorTest.java
```diff
@@ -27,6 +27,7 @@
 import java.util.*;
 
 import org.junit.AfterClass;
+import org.junit.Assert;
```
pentaho-lmartins: Why are you updating this file just to add a new import?
Please revert this file.

core/src/test/java/org/pentaho/di/core/row/value/ValueMetaBaseTest.java
```diff
@@ -1085,4 +1092,62 @@ public void testHashCodes() throws Exception {
     valueMetaString.type = ValueMetaInterface.TYPE_NONE;
     assertEquals( valueMetaString.hashCode( "any" ),  0 );
   }
+
+  @Test
```
pentaho-lmartins: Please update these tests so that you call `setPreparedStatementValue()` with `ValueMetaInterface.TYPE_STRING` and verify that you get the expected values for the different use cases.

core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
cravobranco: Fully disagree with this review comment. Please refer to:
"Although a method name can be any legal identifier, code conventions restrict method names. By convention, method names should be a verb in lowercase or **a multi-word name that begins with a verb in lowercase, followed by adjectives, nouns, etc. In multi-word names**, the first letter of each of the second and following words should be capitalized."
in https://docs.oracle.com/javase/tutorial/java/javaOO/methods.html

mbatchelor: I agree with Miguel on this one.

core/src/main/java/org/pentaho/di/core/database/DatabaseMeta.java
mbatchelor: I agree with Miguel on this one.

core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
pentaho-lmartins: @mbatchelor and @cravobranco, the only problem is that it's not a question of if the DB supports the lenght by bytes or chars but instead that the DB is configured to use it or not.

So If the issue here is that  the method doesn't start with a verb, may I suggest: `getLengthSemanticsUsesChars()`. I trust you'll agree with this one.

core/src/main/java/org/pentaho/di/core/database/DatabaseMeta.java
pentaho-lmartins: @mbatchelor, see comment above.

ppatricio:
Closed as not a Bug.

PR #5688, opened 2018-08-14
[PDI-17336] HTTP Client step doesn't respect the encoding

pentaho-lmartins:
HTTP Client step would not handle the content encoding if not specified in the response entity.
The value specified in the step dialog was never used in the code path.

@pamval @cravobranco @bmorrise @mbatchelor

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/http/HTTP.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/http/HTTP.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
@@ -22,6 +22,7 @@
 
 package org.pentaho.di.trans.steps.http;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang.StringUtils;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
@@ -56,7 +57,6 @@ import org.pentaho.di.trans.step.StepInterface;
 import org.pentaho.di.trans.step.StepMeta;
 import org.pentaho.di.trans.step.StepMetaInterface;
 
-import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.UnknownHostException;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/pentaho/di/trans/steps/http/HTTP.java
```diff
@@ -56,15 +57,14 @@
...
-import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.List;
 
 /**
- * Retrieves values from a database by calling database stored procedures or functions
+ * Retrieves data from an HTTP endpoint
```
pentaho-lmartins: Class description was not OK.

engine/src/main/java/org/pentaho/di/trans/steps/http/HTTP.java
```diff
@@ -187,7 +188,7 @@ public HTTP( StepMeta stepMeta, StepDataInterface stepDataInterface, int copyNr,
           default:
             HttpEntity entity = httpResponse.getEntity();
             if ( entity != null ) {
-              body = EntityUtils.toString( entity );
+              body = StringUtils.isEmpty( meta.getEncoding() ) ? EntityUtils.toString( entity ) : EntityUtils.toString( entity, meta.getEncoding() );
```
pentaho-lmartins: This is the actual fix.

engine/src/main/java/org/pentaho/di/trans/steps/http/HTTP.java
```diff
@@ -282,14 +283,6 @@ protected int requestStatusCode( HttpResponse httpResponse ) {
     return httpResponse.getStatusLine().getStatusCode();
   }
```
pentaho-lmartins: Unused code.

engine/src/test/java/org/pentaho/di/trans/steps/http/HTTPTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+    Object[] row = http.callHttpService( rmi, new Object[] { 0 } );
+    assertEquals( DATA, row[0] );
+  }
+
+  @Test
+  public void callHttpServiceWithoutEncoding() throws Exception {
+    doReturn( null ).when( meta ).getEncoding();
+
+    HTTP http = mock( HTTP.class );
```
cravobranco: **(minor)** The following 9 lines are the exact duplication of lines 105-113.
Evaluate to extract this to a single method.

pentaho-lmartins:
@cravobranco, PR updated with requested changes.

PR #5697, opened 2018-08-15
PDI-17530 - Synchronize on MDI Repo Save

jonjarvis:


lucboudreau:
Sync on a class is to be avoided. Please use a higher level concurrency pattern, like a ReentrantLock, or other.

jonjarvis:
@lucboudreau So you would rather have me use a static RenentrantLock over the built in monitor on the class? All instances of MetaInject need to lock to get around this concurrency issues.

ccaspanello:
This is an interesting read:  https://stackoverflow.com/questions/442564/avoid-synchronizedthis-in-java

lucboudreau:
Ideally you want to use a proper 'latch' or 'lock', one designed for reentrant checks and interruptions. There's usually no need to synchronize directly. Synchronization requires more than just slapping it to pieces of code. You need to handle monitor states and interruptions, etc. Better use the JDK classes from the concurrent packages that encapsulates this logic for you.

In this case you probably want a static property in this class which implements a lock or something similar. I mentioned the ReentrantLock, but there are other things usable, like a Semaphore ( https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Semaphore.html ) Up to you to read and pick.

jonjarvis:
@lucboudreau Updated accordingly - RenentrantLock was fine for me, because that is closest to monitor lock.

PR #5706, opened 2018-08-16
[BACKLOG-24983] Change JSON plugin back to non-OSGI and make new OSGI…

ddiroma:
… plugin for get fields


ddiroma:
@pentaho/r2d2

PR #5714, opened 2018-08-17
[BACKLOG-23109] Change default QoS value to match client default.

peterrinehart:


plugins/streaming/impls/mqtt/src/main/java/org/pentaho/di/trans/step/mqtt/MQTTClientBuilder.java
```diff
@@ -80,7 +80,7 @@
 
   private String broker;
   private List<String> topics;
-  private String qos = "0";
```
mkambol: I don't remember if we defaulted to 0 for a particular reason.  @kurtwalker may know, or Luke.  Probably best not to change our default till we know for sure.

kurtwalker: Pretty sure that came from PM and PO to make it 0 for the default.  I see no reason to change it.

PR #5718, opened 2018-08-20
[BACKLOG-24402] Adds more styling to Kettle status page and fixes issue with styles displaying on Linux

pentaho-mracine:


pentaho-mracine:
@pentaho/r2d2 Please review

engine/src/main/java/org/pentaho/di/www/GetStatusServlet.java
```diff
@@ -253,74 +251,78 @@ public void doGet( HttpServletRequest request, HttpServletResponse response ) th
...
-          line = mantleBr.readLine();
-          while ( line != null ) {
-            sb.append( line );
-            line = mantleBr.readLine();
-          }
-
-          styleBr.close();
-          out.println( sb.toString() );
-          out.println( "</STYLE>" );
+          out.println( "<link rel=\"stylesheet\" type=\"text/css\" href=\"/pentaho/content/common-ui/resources/themes/" + themeName + "/" + themeCss + "\"/>" );
```
bmorrise: How will the styles be handle in the context of standalone carte when pentaho server is not available?

pentaho-mracine: I guess this code doesn't account for that, but should a standalone carte even display themes? If so I would think we'd need to copy all of the Pentaho themes located in the pentaho-solutions/system/common-ui folder over to PDI and then add more logic into this to look into both locations.

pentaho-mracine:
Closing until we can have the themes more accessible from PDI.

pentaho-mracine:
Reopening as we will tackle the standalone carte server styling later.

PR #5728, opened 2018-08-22
[PDI-17505] GetXMLData step is 2-3 times slower due to change made fo…

pentaho-lmartins:
…r PDI-11018

Reverts work done in PDI-11018 (https://github.com/pentaho/pentaho-kettle/commit/7e05d74c391441a7a9749fb8ab826505a19a2330) which solves the performance regression.

Provides a different solution to PDI-11018.
Updated test cases to cover new solution for PDI-11018.

@ricardosilva88 @bmorrise @pamval @mbatchelor


pentaho-lmartins:
@bmorrise please review.

PR #5729, opened 2018-08-22
[SP-4563] Preventing steps with mis-configured or not properly …

pminutillo:
…localized categories from going missing


pminutillo:
@pentaho/r2d2

PR #5738, opened 2018-08-29
[PDI-17345] Adding/Removing a checkpoint causes duplication of <attri…

smmribeiro:
…butes> tag

@pentaho-lmartins


plugins/pur/core/src/test/java/org/pentaho/di/repository/pur/JobDelegateTest.java
```diff
@@ -99,63 +98,15 @@ public static DataNode setNodes( DataNode node, String nrProperty, Iterable<Data
     return node;
   }
 
-  @Test
```
pentaho-lmartins: Why are we deleting this test?

smmribeiro: With this commit, there's an effective change in behaviour: while, until now, the attribute map was persisted once and used on both JobEntry and JobEntryCopy, now the attribute maps of each instance is persisted individually (and separately).

This test was ensuring that the persisted map was loaded to both instances; that is now invalid.

The intended behaviour (separate maps for both instances) is covered in other tests.

pentaho-lmartins:
@bmorrise, can you review this one? It looks OK to me, but this one is out of my depth.

PR #5744, opened 2018-08-29
[BACKLOG-25290] Add libraries for AMQP Consumer  on AEL support.

peterrinehart:


assemblies/lib/pom.xml
```diff
@@ -197,6 +202,61 @@
...
+      <artifactId>amqp-client</artifactId>
+      <version>${amqp366.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
```
mkambol: Why does the 5.2 version need to be here, isn't it only needed by the osgi plugin?  There will be conflicts if both are in the app classpath.

PR #5747, opened 2018-09-03
[PDI-17328] Excel Output:Not able to create Dynamically new worksheets

smmribeiro:
@pentaho-lmartins


pentaho-lmartins:
We need some test cases to confirm all use cases are OK.

smmribeiro:
I've added tests for the following scenarios:

- No template + append
- No template + don't append
- Template (without the intended sheet) + append
- Template (without the intended sheet) + don't append
- Template (with the intended sheet) + append
- Template (with the intended sheet) + don't append

PR #5749, opened 2018-09-05
[BACKLOG - 25262] - Provided a connection to the repository for the "…

RPAraujo:
…data-services" in the CDA services

@pentaho/millenniumfalcon please review

### DIFF
--- a/plugins/pur/repository-proxy/src/main/java/org/pentaho/di/repository/pur/provider/PurRepositoryProxy.java
+++ b/plugins/pur/repository-proxy/src/main/java/org/pentaho/di/repository/pur/provider/PurRepositoryProxy.java
@@ -1,3 +1,19 @@
+/*!
+ * Copyright 2018 Hitachi Vantara.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
 package org.pentaho.di.repository.pur.provider;
 
 import org.pentaho.di.cluster.ClusterSchema;
--- a/plugins/pur/repository-proxy/src/main/java/org/pentaho/di/repository/pur/provider/PurRepositoryProxy.java
+++ b/plugins/pur/repository-proxy/src/main/java/org/pentaho/di/repository/pur/provider/PurRepositoryProxy.java
@@ -1,0 +1,717 @@
+package org.pentaho.di.repository.pur.provider;
+
+import org.pentaho.di.cluster.ClusterSchema;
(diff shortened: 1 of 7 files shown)


plugins/pur/repository-proxy/src/main/java/org/pentaho/di/repository/pur/provider/PurRepositoryProxy.java
```diff
@@ -0,0 +1,717 @@
...
+          + "changed id.", PUR_PLUGIN_ID );
+    }
+    return this.delegate = repository;
+  }
+
+  /**
+   * @param locationUrl
+   */
+  public void setLocationUrl( String locationUrl ) {
+//    this.getDelegate();
```
nantunes: If this is just forgotten commented-out old code, it should be removed. If not, please add a comment to explain why this is here.

plugins/pur/repository-proxy/src/main/java/org/pentaho/di/repository/pur/provider/RepositoryProvider.java
```diff
@@ -0,0 +1,88 @@
...
+  public void setRepository( Repository repository ) {
+    this.repository = repository;
+  }
+
+
+  /**
+   * Resets the repository so that next time {@link #getRepository()} is called a new repository is returned.
+   * This is called whenever the user or password is changed for this instance of the repository
+   */
+  public void reconnectToRepository() {
```
nantunes: Couldn't this be private? Isn't it just internal connection management logic?

RPAraujo: Yes it's not used anywhere else. I'll change it to private.

RPAraujo: Actually it has to be protected since it's used in unit tests

plugins/pur/repository-proxy/src/main/java/org/pentaho/di/repository/pur/provider/PurRepositoryProxy.java
RPAraujo: The comment was forgotten there. It's been removed

plugins/pur/repository-proxy/src/main/java/org/pentaho/di/repository/pur/provider/RepositoryProvider.java
nantunes: It can be package private. But also maybe the unit test could be revised, as it is directly calling private logic? Calling `getRepository()` instead would test the "real world" scenario.

graimundo: It is public because I see it as part of the API provided by the class. The class is a provider of connected repositories and as such a reconnectToRepository method makes sense to be public in the same way that the get/set Username / password are public.

graimundo:
*Note* The coverage is low because there are generated methods for the delegate pattern for which we don't gain value on testing them. https://github.com/pentaho/pentaho-kettle/pull/5749/files#diff-57c63a4fdbfcb275a67c93b14345f56eR127

graimundo:
[BACKLOG - 25262] - Created PURRepositoryProxy to use a PurRepository in an OSGi context while having PUR loaded as a non OSGi plugin.

PR #5754, opened 2018-09-07
[PDI-17644] - StaxPoiSheet.java creates additional object with null

MAakula:
Empty Cell creates an additional Object with null in the cells field.
Problem Scenario:
Consider an Excel (xlsx with POI Streaming) file with 5 columns.
Header1,Header2,Header3,Header4,Header5
Data11,Data12,Data13,Data14,Data15                      => cells arraylist holds 5 objects [Data11,Data12,Data13,Data14,Data15] =>Correct
Data21,Data22,,Data24,Data25                                 => cells arraylist holds 6 objects [Data21,Data22,null,null,Data24,Data25] =>Wrong

Solution:
undefinedColIndex must be incremented always.


MAakula:
Fix for Excel empty cell data shift issue.

mbatchelor:
PDI-17644 - StaxPoiSheet creates an additional object with null when it encounters a null field.

mbatchelor:
@MAakula - Thank-you for the contribution. Ordinarily, I'd ask that you file a JIRA case at https://jira.pentaho.com but in this case, I created it for you and linked the check-in to the JIRA case. Thank-you again for the fix!

PR #5758, opened 2018-09-10
[BACKLOG-25453] "No new fields to process" error when using "Get Fiel…

ddiroma:
…ds" with valid remote file in Text File Input step


ddiroma:
@pentaho/r2d2 
@pentaho-amessier

PR #5760, opened 2018-09-10
[BACKLOG-25556] Fix for text file output filename handing

pentaho-amessier:
Also making the related unit test more robust to catch this specific bug


pentaho-amessier:
@pentaho/r2d2

PR #5772, opened 2018-09-14
[PDI-17387] The import script should not be blocked from loading repo…

ppatricio:
…sitory exports to the root of the repository

This PR fixes the issue that always give an error if user selects Root for the destination path. ( Introduced in PDI-16591). Now it considers the full path ( destination path + path defined in file to import ) instead of validating only the destination path.
It was added a new error message when the user tries to create a new directory at root level. For instance if you choose / as destination path and your repository file to import has a directory attribute like /anyFolderName/filename this will invoke the new message, since it is not valid path. You can only save to the "home" and "public" folders, at root level.

@mbatchelor @bmorrise @kcruzada

### DIFF
--- a/engine/src/main/java/org/pentaho/di/imp/Import.java
+++ b/engine/src/main/java/org/pentaho/di/imp/Import.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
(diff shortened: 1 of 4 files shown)


ppatricio:
This test case sometimes breaks building, checked in Cirene too. I don't think it's related to this PR but asked @ricardosilva88 for some review on BISERVER-13923.

ricardosilva88:
@ppatricio, just submitted a new PR to fix those tests https://github.com/pentaho/pentaho-kettle/pull/5796

PR #5790, opened 2018-09-19
[PDI-17209] - Group By: When using Cumulative sum (all rows option on…

cravobranco:
…ly!)) option with "Include all rows" item is unchecked, the data transformation finished normal

* Fix the edge case when GroupBy dialog is opened for the first time and the new implementation is not according to .ktr stored values

@pentaho-lmartins , @pamval , @mbatchelor Please review and merge if it's all ok.


pentaho-lmartins:
@bmorrise please review.

PR #5798, opened 2018-09-24
[PDI-17255] - Add Constants : Format settings are actually not gettin…

cravobranco:
…g applied to the output.

* Added mechanism to apply the format settings through the conversion mask in the Add Constants step.

@pentaho-lmartins , @mbatchelor , @pamval , @mkambol Please review and merge if it's all ok.


pentaho-lmartins:
@bmorrise please review.

PR #5804, opened 2018-09-25
[BACKLOG-25488] Updates Kettle Trans status page to match new styles

pentaho-mracine:


engine/src/main/java/org/pentaho/di/www/StatusServletUtils.java
```diff
@@ -0,0 +1,78 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
+ * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
```
ddiroma: @pentaho-mracine 
No need to have 2002-2018 on new files...just 2018

pentaho-mracine: @ddiroma thanks for noticing that, just a leftover from copying it from another file. Updating now...

PR #5811, opened 2018-09-26
[PDI-17492] The list of variables in the Run Config UI for Transforma…

ppatricio:
…tions shows variables from other Transformations

Do not merge yet. Just presenting a possible solution. Read Jira comments first.
Needs validation.


xiaoaowanghu:
Can u please take a look on  #5810. That's a big bug I think

PR #5815, opened 2018-09-27
8.1 feature

tkafalas:


pentaho-lmartins:
@tkafalas, thank you for the PR, but we need to follow the maintenance process which means:
1. Create an SP case in JIRA cloned from the backlog case
2. Create PRs to 8.1 branches that reference the SP case created in JIRA
3. Commit messages in maintenance branches should contain the valid SP case number
I'm closing this PR and will create a new one once the SP cases are created.

PR #5820, opened 2018-09-28
[PDI-17623] PGBulkLoader match client encoding automatically

wangzw:
On the OS which has default character set other than UTF8, PGBulkloader may throw error of encoding issue.

In this patch, I fix above issue by encoding data string using PostgreSQL client encoding instead of OS default.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/pgbulkloader/PGBulkLoader.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/pgbulkloader/PGBulkLoader.java
@@ -31,7 +31,10 @@ package org.pentaho.di.trans.steps.pgbulkloader;
 //
 
 import java.math.BigDecimal;
+import java.nio.charset.Charset;
 import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.sql.Statement;
 
 import org.pentaho.di.core.Const;
@@ -67,6 +70,7 @@ import org.postgresql.PGConnection;
 public class PGBulkLoader extends BaseStep implements StepInterface {
   private static Class<?> PKG = PGBulkLoaderMeta.class; // for i18n purposes, needed by Translator2!!
 
+  private Charset clientEncoding = Charset.defaultCharset();
   private PGBulkLoaderMeta meta;
   private PGBulkLoaderData data;
   private PGCopyOutputStream pgCopyOut;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/pentaho/di/trans/steps/pgbulkloader/PGBulkLoader.java
```diff
@@ -133,12 +136,40 @@ public String getCopyCommand( ) throws KettleException {
...
+
+        if (rs.getMetaData().getColumnCount() != 1) {
+          throw new KettleException( "Cannot detect client encoding" );
+        }
+
+        String clientEncodingStr = rs.getString(1);
+        logBasic( "Detect client encoding: " + clientEncodingStr );
+        clientEncoding = Charset.forName(clientEncodingStr);
+      }
+    } catch ( Exception ex ) {
```
mbatchelor: Should not just catch exception here - should catch SQLException on this line.

engine/src/main/java/org/pentaho/di/trans/steps/pgbulkloader/PGBulkLoader.java
```diff
@@ -133,12 +136,40 @@ public String getCopyCommand( ) throws KettleException {
...
 
+  void checkClientEncoding() throws Exception {
+    Connection connection = data.db.getConnection();
+
+    Statement statement = connection.createStatement();
+
+    try {
+      try (ResultSet rs = statement.executeQuery("show client_encoding")) {
+        if (!rs.next()) {
+          throw new KettleException( "Cannot detect client encoding" );
```
mbatchelor: It should not be an exception that the client encoding cannot be auto-detected. Rather, you can log a warning, and default the client encoding to the system default encoding. If for some reason you cannot get the encoding, existing systems should continue to work as they have before.

engine/src/main/java/org/pentaho/di/trans/steps/pgbulkloader/PGBulkLoader.java
```diff
@@ -133,12 +136,40 @@ public String getCopyCommand( ) throws KettleException {
...
+        }
+
+        if (rs.getMetaData().getColumnCount() != 1) {
+          throw new KettleException( "Cannot detect client encoding" );
```
mbatchelor: Same comment here as above.

engine/src/main/java/org/pentaho/di/trans/steps/pgbulkloader/PGBulkLoader.java
```diff
@@ -133,12 +136,40 @@ public String getCopyCommand( ) throws KettleException {
...
+    }
+  }
+
   private void do_copy( PGBulkLoaderMeta meta, boolean wait ) throws KettleException {
     data.db = getDatabase( this, meta );
     String copyCmd = getCopyCommand();
     try {
       connect();
 
+      checkClientEncoding();
```
mbatchelor: After this line, you need to be sure that clientEncoding is non-null - if it is, assign to the system default encoding so that the system behaves as it always has in the event of a failure to detect the encoding of the client connection.

engine/src/main/java/org/pentaho/di/trans/steps/pgbulkloader/PGBulkLoader.java
```diff
@@ -133,12 +136,40 @@ public String getCopyCommand( ) throws KettleException {
...
+          throw new KettleException( "Cannot detect client encoding" );
+        }
+
+        if (rs.getMetaData().getColumnCount() != 1) {
+          throw new KettleException( "Cannot detect client encoding" );
+        }
+
+        String clientEncodingStr = rs.getString(1);
+        logBasic( "Detect client encoding: " + clientEncodingStr );
+        clientEncoding = Charset.forName(clientEncodingStr);
```
mbatchelor: Need a try/catch around this line to catch a few specific exceptions:
IllegalCharsetNameException - If the given charset name is illegal
IllegalArgumentException - If the given charsetName is null
UnsupportedCharsetException - If no support for the named charset is available in this instance of the Java virtual machine

In all three of these cases, charset will need to be set to null so that down below, it can be specified using the system default encoding.

wangzw: IllegalArgumentException is superclass of IllegalCharsetNameException and UnsupportedCharsetException

clientEncoding has system default charset as default value

mbatchelor:
Actually, wingman is insane in that report. I've checked the file myself, even loaded it into a hex editor. There's nothing at all wrong with line 146 or any of the other lines I've looked at. No tabs, no funky control characters, nothing. Also, I checked out the PR and did a mvn site and checked the checkstyle report. Nothing reported.

wangzw:
@mbatchelor wingman is not insane, I fixed the style issue before your comments.

PR #5823, opened 2018-10-01
[BACKLOG-25274] - refactoring BaseStepMeta to match previous public A…

kurtwalker:
…PI because community steps may possibly be referencing the ioMeta field


kurtwalker:
original commit was https://github.com/pentaho/pentaho-kettle/commit/e8efd26e658349fdfc710df13044e7d0ad11a280

kurtwalker:
per marc, we should fix any plugins that use the old behavior.  protected fields not considered public api.  closing

PR #5836, opened 2018-10-05
[PDI-17656] Use internalization key instead of hardcode

HiromuHota:
This patch originally came from @Valeran86 to HiromuHota/pentaho-kettle as HiromuHota/pentaho-kettle#121 but I think it should be merged to the upstream.


mbatchelor:
@HiromuHota - thank-you so much for the contribution. Please file a defect at https://jira.pentaho.com and update the PR description to include the fixed JIRA case. That way we can process the PR. Thanks!!

HiromuHota:
I filed a defect at the JIRA, then updated the PR description and the commit message.
Please let me know if there are any other things that need editing.

PR #5849, opened 2018-10-10
[BACKLOG-24360] Error reading (g)zip files in Text File Input

NJtwentyone:
-- assumption is external code will use apache vfs2 uri selector to get list of files inside (g)zip

see jira issue for more details on background and implementation


NJtwentyone:
per discussion with Peter, going to add prefix behind the scenes and make it transparent to the user.

PR #5765, opened 2018-09-13
fixed ConcurrentModificationException in AbstractMeta

zhaxiaodong9860:
In the case of multi-thread, when one thread traverses the collection and another thread operates on the collection, an error occurs, causing the job to stop. The following is the error message:
2018/09/13 10:10:23 - MAIN_SJCJ - ERROR (version 8.0.0.0-28, build 8.0.0.0-28 from 2017-11-05 07.27.50 by buildguy) : /KETTLE8_RES/MAIN_TRAN/MAIN_SJCJ
org.pentaho.di.core.exception.KettleException: 
Unexpected error occurred while launching entry [SSCJ_2.0]
 at java.lang.Thread.run (Thread.java:748)
 at org.pentaho.di.job.Job$1.run (Job.java:850)
 at org.pentaho.di.job.Job.access$000 (Job.java:121)
 at org.pentaho.di.job.Job.execute (Job.java:732)
 at org.pentaho.di.job.entries.job.JobEntryJob.execute (JobEntryJob.java:617)
 at org.pentaho.di.base.AbstractMeta.setInternalKettleVariables (AbstractMeta.java:409)
 at org.pentaho.di.job.JobMeta.setInternalKettleVariables (JobMeta.java:2341)
 at org.pentaho.di.job.JobMeta.updateCurrentDir (JobMeta.java:2351)
 at org.pentaho.di.base.AbstractMeta.fireCurrentDirectoryChanged (AbstractMeta.java:644)
 at java.util.ArrayList$Itr.next (ArrayList.java:851)
 at java.util.ArrayList$Itr.checkForComodification (ArrayList.java:901)

	at org.pentaho.di.job.Job$1.run(Job.java:854)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.util.ConcurrentModificationException
	at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:901)
	at java.util.ArrayList$Itr.next(ArrayList.java:851)
	at org.pentaho.di.base.AbstractMeta.fireCurrentDirectoryChanged(AbstractMeta.java:644)
	at org.pentaho.di.job.JobMeta.updateCurrentDir(JobMeta.java:2351)
	at org.pentaho.di.job.JobMeta.setInternalKettleVariables(JobMeta.java:2341)
	at org.pentaho.di.base.AbstractMeta.setInternalKettleVariables(AbstractMeta.java:409)
[...]

PR #5866, opened 2018-10-15
[PDI-17646] - ConcurrentModificationException in AbstractMeta

mbatchelor:


mbatchelor:
Inspired by community contribution on #5765

engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
```diff
@@ -139,13 +139,13 @@
 
   protected List<DatabaseMeta> databases;
 
-  protected List<NameChangedListener> nameChangedListeners;
+  protected Set<NameChangedListener> nameChangedListeners = Collections.newSetFromMap( new ConcurrentHashMap<NameChangedListener, Boolean>() );
```
lucboudreau: Sets are backed by the map, not copied, so this is OK.

https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#newSetFromMap(java.util.Map)

PR #5878, opened 2018-10-16
[BACKLOG-25950, BACKLOG-25735] Resolve UX Issues with Carte server, t…

ddiroma:
…ransformation, and job status pages AND Add Tooltips for all the icons in the Carte Status page


ddiroma:
@pentaho/r2d2

PR #5898, opened 2018-10-19
[SP-4672] - Backport of PDI-17255 - Add Constants : Format settings a…

cravobranco:
…re actually not getting applied to the output. (7.1 Suite)

* Backport of PDI-17255 - Add Constants : Format settings are actually not getting applied to the output. (7.1 Suite)

@smmribeiro , @LeonardoCoelho71950


smmribeiro:
@cravobranco, seems that something went wrong with the code :-)

cravobranco:
@smmribeiro , it seems to be better now.

PR #5936, opened 2018-10-29
[BACKLOG-26045] Creating a data-lineage analyzer for the Simple Mapping (sub-transformation) step

pentaho-amessier:


pentaho-amessier:
@pentaho/r2d2

PR #5939, opened 2018-10-29
[BACKLOG-26374] steps creating field name conflicts

tgf:
repeated field names could arise after inserting or deleting value metas
select values not checking for conflicts when renaming


tgf:

PR #5949, opened 2018-10-30
[PDI-17699]  DragAndDropContainerTest fail on Windows

Ozgay:
unit test fail on line 95 and line 109 of DragAndDropContainerTest.java
"\n" should be replace by Const.CR since this case may run on Windows


ppatricio:
@Ozgay  - Thank-you for the contribution. Your fix is correct but we need to import the Const class first.
To ensure we get this in time in our master, I'm closing this PR and  provide a new PR (#6005) with the missing import and update the remain tests with the "Const.CR" too. 
Thank-you again for the fix!

PR #5955, opened 2018-10-30
[BACKLOG-26358] Fixes issue that allowed users to create folders with…

pentaho-mracine:
… special slash characters

Dev validated on 9.0.0.0-7

PR that was merged into master: https://github.com/pentaho/pentaho-kettle/pull/5946

@kcruzada @pamval @pentaho-jlien @pentaho-svigneri

### DIFF
--- a/plugins/file-open-save/core/src/main/javascript/app/services/data.service.js
+++ b/plugins/file-open-save/core/src/main/javascript/app/services/data.service.js
@@ -106,7 +106,7 @@ define(
         }
 
         function search(path, value) {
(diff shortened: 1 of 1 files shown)


pamval:
@pentaho-mracine are we sure about this fix? We're avoiding url encoding the name when it contains either / or \. This does not look correct - we should *always* encode the name to make sure it is accepted by the server.  If \ or / are not acceptable characters for a folder name (and they should not be) then the API receiving the request should block those chars.

pentaho-mracine:
Closing as we will address this in a later SP.

PR #5957, opened 2018-10-30
[TEST] fix failing integration tests in engine

tgf:
PanIT was checking for wrong error codes
PropertyOutputIT can get loaded from it-classes and not find its ktr, so
moved to it


tgf:

PR #5959, opened 2018-10-30
[BACKLOG-26377] Deferring close of mqtt client until trans is

mkambol:
stopped or disposed.

https://jira.pentaho.com/browse/BACKLOG-26377

### DIFF
--- a/plugins/streaming/impls/mqtt/src/main/java/org/pentaho/di/trans/step/mqtt/MQTTProducer.java
+++ b/plugins/streaming/impls/mqtt/src/main/java/org/pentaho/di/trans/step/mqtt/MQTTProducer.java
@@ -43,6 +43,7 @@ import org.pentaho.di.trans.step.StepMetaInterface;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Optional.ofNullable;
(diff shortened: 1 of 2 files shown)


plugins/streaming/impls/mqtt/src/main/java/org/pentaho/di/trans/step/mqtt/MQTTProducer.java
```diff
@@ -191,15 +193,25 @@ private String getTopic( Object[] row ) {
...
+  @Override public void dispose( StepMetaInterface smi, StepDataInterface sdi ) {
+    super.dispose( smi, sdi );
+    stopMqttClient();
+  }
+
   private void stopMqttClient() {
     try {
       // Check if connected so subsequent calls does not produce an already stopped exception
-      if ( null != client.get() && client.get().isConnected() ) {
+      if ( !connectionError.get()
```
kurtwalker: Is there a case when connectionError and isConnected are both true?  not sure I understand what connectionError is buying us.

mkambol: connectionError is just preventing a re-attempt at making the connection.  `client` is a supplier, so if the supplier fails to initialize the first time we'll just try again when we're in the process of closing.
Not strictly necessary, but prevents some duplicate error logging.

plugins/streaming/impls/mqtt/src/main/java/org/pentaho/di/trans/step/mqtt/MQTTProducer.java
```diff
@@ -54,6 +55,7 @@
   private MQTTProducerMeta meta;
 
   Supplier<MqttClient> client = Suppliers.memoize( this::connectToClient );
+  private AtomicBoolean connectionError = new AtomicBoolean( false );
```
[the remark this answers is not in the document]
mkambol: Also, for thread visibility.  I think volatile boolean works, too, but Atomic is easier for me to think about.
With AEL, now that forceCoalesce will be true, this _should_ only occur on a single thread, but doesn't hurt.

PR #5973, opened 2018-11-05
[BACKLOG-26264] Add a check to verify if the step on the Result field…

peterrinehart:
…s tab exists in the subtrans.


engine/src/main/java/org/pentaho/di/trans/streaming/common/BaseStreamStepMeta.java
```diff
@@ -194,18 +210,20 @@ public String getFileName() {
...
-          .ifPresent( stepMeta ->
-          {
-            try {
-              stepMeta.getStepMetaInterface()
-                .getFields( rowMeta, origin, info, nextStep, space, repository, metaStore );
-            } catch ( KettleStepException e ) {
-              throw new RuntimeException( e );
-            }
-          } );
+        if ( transMeta.getSteps().stream().anyMatch( stepMeta -> stepMeta.getName().equals( realSubStepName ) ) ) {
```
kurtwalker: Do we need an else here?  We still want an exception I think, it just needs to be more informative.  Perhaps just adding a log message to the "throw new RuntimeException" a few lines down?

PR #5985, opened 2018-11-07
[PDI-17692] Overriding the dialect on Generic Connection is not being…

ssamora:
… called as expected

@pentaho-lmartins

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
+++ b/core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
@@ -221,6 +221,25 @@ public interface DatabaseInterface extends Cloneable {
   public void setAttributes( Properties attributes );
 
   /**
+   * Add extra attribute on this connection
+   * @param attributeId the attribute identifier
+   * @param value the value of the attribute
+   */
+  public default void addAttribute( String attributeId, String value ) {
+    // Default implementation does nothing
+  };
+
+  /**
+   * Gets an attribute from the connection
+   * @param attributeId the attribute identifier
+   * @param defaultValue the default value in case the attribute is not found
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
```diff
@@ -220,6 +220,21 @@
    */
   public void setAttributes( Properties attributes );
 
+  /**
+   * Add extra attribute on this connection
+   * @param attributeId the attribute identifier
+   * @param value the value of the attribute
+   */
+  public void addAttribute( String attributeId, String value );
```
mbatchelor: Likely need a default here. Otherwise, community members with their own implementations won't work.

core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
```diff
@@ -220,6 +220,21 @@
...
+
+  /**
+   * Gets an attribute from the connection
+   * @param attributeId the attribute identifier
+   * @param defaultValue the default value in case the attribute is not found
+   * @return the attribute value
+   */
+  public String getAttribute( String attributeId, String defaultValue );
```
mbatchelor: As above

core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
ssamora: correct. will update. Thanks

core/src/main/java/org/pentaho/di/core/database/DatabaseInterface.java
ssamora: same here.

PR #5988, opened 2018-11-07
[BACKLOG-25662] Suppress file not found error message when running a …

peterrinehart:
…streaming consumer on AEL and stopping the transformation.


kurtwalker:
I don't have an alternative, but I'm not sure about changing VFS for an ael use case.  Can we explore something that only affects the code in ael land?

PR #6012, opened 2018-11-14
[PDI-17713] - GetPOP creates bad file names on duplicates

mbatchelor:


engine/src/main/java/org/pentaho/di/job/entries/getpop/MailConnection.java
```diff
@@ -837,46 +825,55 @@ private void handlePart( String foldername, Part part, Pattern pattern ) throws
     }
   }
 
+  @VisibleForTesting
+  static String findValidTarget( String folderName, String fileName ) throws KettleException {
```
pentaho-lmartins: Why not use `org.apache.commons.io.FilenameUtils` to get the filename without extension and extension.
The code would be cleaner.

mbatchelor: Agreed - good call. I've updated the PR with your suggestion as well as improved the performance of the algorithm.

mbatchelor:
Deleted old wingman report since I updated the PR.

mbatchelor:
@pentaho-lmartins - Can you review my changes

pentaho-lmartins:
Deleted Wingman reports, running Wingman again since this has been here for too long.

PR #6014, opened 2018-11-15
[PDI-17449] S3 CSV Input does not work after upgrade from 7.1 to 8.1

pentaho-lmartins:
Partial revert of 5bc0e4b85473fa7f5fe8c7f595ee7fda60ec5d41 to bring back lost functionality, i.e. ability to specify AWS credentials within the step as requested in JIRA case.

By default, if you don't specify AWS credentials, Spoon will get credentials from the Amazon Default Credentials Provider Chain. If credentials are specified in the step, Spoon will use those (legacy behavior).

For legacy transformations with embedded credentials you now have the option to override those credentials and use the provider chain by setting the Kettle property `KETTLE_USE_AWS_DEFAULT_CREDENTIALS` to `Y`.

@rmansoor @mbatchelor @pentaho/tatooine

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/Const.java
+++ b/core/src/main/java/org/pentaho/di/core/Const.java
@@ -1279,6 +1279,12 @@ public class Const {
     String.valueOf( KETTLE_ZIP_MAX_TEXT_SIZE_DEFAULT );
 
   /**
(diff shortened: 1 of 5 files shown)


pentaho-lmartins:
@mbatchelor @rmansoor can you please review this?

PR #6018, opened 2018-11-16
[PDI-17585] Variable Replacement Is Not Working in Connection Pool Co…

LeonardoCoelho71950:
…nfiguration

@pentaho/tatooine @pamval 
@bmorrise @mkambol

### DIFF
--- a/dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
+++ b/dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
@@ -764,7 +764,7 @@ public class DataHandler extends AbstractXulEventHandler {
     if ( meta.isUsingConnectionPool() ) {
       if ( poolSizeBox != null ) {
         try {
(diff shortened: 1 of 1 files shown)


dbdialog/src/main/java/org/pentaho/ui/database/event/DataHandler.java
```diff
@@ -764,7 +764,15 @@ protected void getInfo( DatabaseMeta meta ) {
     if ( meta.isUsingConnectionPool() ) {
       if ( poolSizeBox != null ) {
         try {
-          int initialPoolSize = Integer.parseInt( poolSizeBox.getValue() );
+          int initialPoolSize;
```
mbatchelor: I think you're looking for databaseMeta.environmentSubstitute here - you shouldn't be picking apart the variable like this.

LeonardoCoelho71950: @mbatchelor Now I believe is ready to be merged.

PR #6027, opened 2018-11-21
[PDI-17618] Step copies : not showing parameter

LeonardoCoelho71950:
@pentaho/tatooine @pamval 
@mkambol


LeonardoCoelho71950:
getCopiesString() is the value written on the input box. 
getCopies() is the actual number of copies.

For instance, imagine that we have a parameter N=1:
   getCopiesString() = ${N}
   getCopies() = 1

In this case, we want to show that the number of copies is x${N}

PR #6028, opened 2018-11-21
[PDI-17736] Job Runs Locally but not on a Carte server

ppatricio:
When we execute on a Carte Server, there´s no filename, since we load data from a block of xml and not from a file. Even more if we are not in a repository context, we also don't have the repository directory info. So the job was not running remotely because the internal variable directory logic didn't take this case in consideration. This logic was added.

@mbatchelor @lucboudreau


ppatricio:
PR Updated. Thanks.

PR #6042, opened 2018-11-27
[BACKLOG-24952] Adding support for clonable interfaces to address con…

pminutillo:
…currency issues for data lineage

GetXMLDataStepAnalyzer
JsonInputAnalyzer
JsonInputAnalyzer

### DIFF
--- a/plugins/json/core/src/test/java/org/pentaho/di/trans/steps/jsoninput/analyzer/JsonInputAnalyzerTest.java
+++ b/plugins/json/core/src/test/java/org/pentaho/di/trans/steps/jsoninput/analyzer/JsonInputAnalyzerTest.java
@@ -196,4 +196,17 @@ public class JsonInputAnalyzerTest {
 
     assertEquals( JsonInputMeta.class, consumer.getMetaClass() );
   }
+  
+  @Test
+  public void testCloneAnalyzer() {
+    final JsonInputAnalyzer analyzer = new JsonInputAnalyzer();
+    // verify that cloneAnalyzer returns an instance that is different from the original
+    assertNotEquals( analyzer, analyzer.cloneAnalyzer() );
+  }
+
+  @Test
+  public void testNewInstance(){
+    JsonInputAnalyzer analyzer = new JsonInputAnalyzer();
+    assertTrue( analyzer.newInstance().getClass().equals(JsonInputAnalyzer.class));
+  }
 }
--- a/plugins/xml/core/src/test/java/org/pentaho/di/trans/steps/getxmldata/GetXMLDataStepAnalyzerTest.java
+++ b/plugins/xml/core/src/test/java/org/pentaho/di/trans/steps/getxmldata/GetXMLDataStepAnalyzerTest.java
@@ -322,5 +322,17 @@ public class GetXMLDataStepAnalyzerTest {
 
     assertEquals( GetXMLDataMeta.class, consumer.getMetaClass() );
   }
+  
+  @Test
+  public void testCloneAnalyzer() {
+    final GetXMLDataStepAnalyzer analyzer = new GetXMLDataStepAnalyzer();
+    // verify that cloneAnalyzer returns an instance that is different from the original
+    assertNotEquals( analyzer, analyzer.cloneAnalyzer() );
+   }
 
+  @Test
+  public void testNewInstance(){
+    GetXMLDataStepAnalyzer analyzer = new GetXMLDataStepAnalyzer();
+    assertTrue( analyzer.newInstance().getClass().equals(GetXMLDataStepAnalyzer.class));
+  }
 }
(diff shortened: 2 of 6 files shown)


pminutillo:
@pentaho/r2d2

NJtwentyone:
Missing a couple of things use @pentaho-amessier  select values as template.
https://github.com/pentaho/pentaho-metaverse/pull/524

NJtwentyone:
Need to update /pentahoObjects.spring.xml. I forgot to on one. It's not necessary, but it affects some testing, but none in these submitted.

plugins/json/core/src/test/java/org/pentaho/di/trans/steps/jsoninput/analyzer/JsonInputAnalyzerTest.java
```diff
@@ -196,4 +196,10 @@ public void testJsonInputExternalResourceConsumer() throws Exception {
```
NJtwentyone: The test classes can extend  ClonableStepAnalyzerTest
```
public class JsonInputAnalyzerTest.java  extends ClonableStepAnalyzerTest
```

with overridding the method `newInstance`, you'll get for free the unit test `testCloneAnalyzer` which that each new call returns a unique instance.

plugins/xml/core/src/test/java/org/pentaho/di/trans/steps/getxmldata/GetXMLDataStepAnalyzerTest.java
```diff
@@ -323,4 +323,9 @@ public void testGetXMLDataExternalResourceConsumer() throws Exception {
     assertEquals( GetXMLDataMeta.class, consumer.getMetaClass() );
   }
 
+  @Test
+  public void testNewInstance(){
```
pentaho-amessier: It would be good to add a test for cloneAnlayzer - to ensure that the clone isn't equal to the original, much like is being done here:
https://github.com/pentaho/pentaho-metaverse/pull/524/files#diff-196b64eefb5cedb6bdbcdcdf5048c586R34

NJtwentyone:
missing changing IStepAnalyzer to IClonableStepAnalyzer in blueprint.xml

plugins/xml/core/src/test/java/org/pentaho/di/trans/steps/getxmldata/GetXMLDataStepAnalyzerTest.java
NJtwentyone: I didn't look closely, it not the metaverse project. my bad.

pentaho-amessier: Something like this... since you cannot extend the ClonableStepAnalzyerTest and reuse its common functionality:
 ```
 public void testCloneAnalyzer() {
    final GetXMLDataStepAnalyzer analyzer = new GetXMLDataStepAnalyzer();
    // verify that cloneAnalyzer returns an instance that is different from the original
    Assert.assertNotEquals( analyzer, analyzer.cloneAnalyzer() );
  }
```

File: plugins/json/core/src/test/java/org/pentaho/di/trans/steps/jsoninput/analyzer/JsonInputAnalyzerTest.java (hunk 1 above)
Comment (NJtwentyone) [reply plugins/json/core/src/test/java/org/pentaho/di/trans/steps/jsoninput/analyzer/JsonInputAnalyzerTest.java]: don't think this is valid due to it's not in the metaverse project.

File: plugins/xml/core/src/test/java/org/pentaho/di/trans/steps/getxmldata/GetXMLDataStepAnalyzerTest.java (hunk 2 above)
Comment (pentaho-amessier) [reply plugins/xml/core/src/test/java/org/pentaho/di/trans/steps/getxmldata/GetXMLDataStepAnalyzerTest.java]: @NJtwentyone no worries - I actually envisioned this class being reusable everywhere, so it didn't occur to me until now that it wouldn't be possible

File: plugins/json/core/src/test/java/org/pentaho/di/trans/steps/jsoninput/analyzer/JsonInputAnalyzerTest.java [hunk 3]
```diff
@@ -196,4 +196,17 @@ public void testJsonInputExternalResourceConsumer() throws Exception {
...
+  
+  @Test
+  public void testCloneAnalyzer() {
+    final JsonInputAnalyzer analyzer = new JsonInputAnalyzer();
+    // verify that cloneAnalyzer returns an instance that is different from the original
+    assertNotEquals( analyzer, analyzer.cloneAnalyzer() );
+  }
+
+  @Test
+  public void testNewInstance(){
```
pentaho-amessier: This test isn't really needed - if it was added for coverage, I think the new added testCloneAnalyzer might do the trick. It doesn't hurt though, so we can leave it here.... Also, just a nit-picky comment, there's a missing space before the last closing parenthesis ;) - I'm guessing checkstyle isn't run on this test code.

PR #6055, opened 2018-11-28
[BACKLOG-23288] Move consumer only meta fields from Jms Delegate to JmsConsumerMeta so they do not show up in jms producer mdi

ricksjames:


kurtwalker:
wait to merge.  James is testing to make sure we're backward compatible with this change

PR #6060, opened 2018-12-03
[BACKLOG-26741] 8.2 Carte status page unable to acess Scaleable Vecto…

ddiroma:
…r Graphic (.svg) files

### DIFF
--- a/engine/src/main/java/org/pentaho/di/www/GetJobStatusServlet.java
+++ b/engine/src/main/java/org/pentaho/di/www/GetJobStatusServlet.java
@@ -195,6 +195,9 @@ public class GetJobStatusServlet extends BaseHttpServlet implements CartePluginI
 
     String jobName = request.getParameter( "name" );
     String id = request.getParameter( "id" );
+    String root = request.getRequestURI() == null ? StatusServletUtils.PENTAHO_ROOT
+      : request.getRequestURI().substring( 0, request.getRequestURI().indexOf( CONTEXT_PATH ) );
+    String prefix = isJettyMode() ? StatusServletUtils.STATIC_PATH : root + StatusServletUtils.RESOURCES_PATH;
     boolean useXML = "Y".equalsIgnoreCase( request.getParameter( "xml" ) );
     int startLineNr = Const.toInt( request.getParameter( "from" ), 0 );
 
@@ -312,7 +315,11 @@ public class GetJobStatusServlet extends BaseHttpServlet implements CartePluginI
             + "\">" );
         }
         out.println( "<META http-equiv=\"Content-Type\" content=\"text/html; charset=UTF-8\">" );
-        out.print( StatusServletUtils.getPentahoStyles() );
+        if ( isJettyMode() ) {
+          out.println( "<link rel=\"stylesheet\" type=\"text/css\" href=\"/static/css/carte.css\" />" );
+        } else {
+          out.print( StatusServletUtils.getPentahoStyles() );
+        }
         out.println( "</HEAD>" );
         out.println( "<BODY style=\"overflow: auto;\">" );
         out.println( "<div class=\"row\" id=\"pucHeader\">" );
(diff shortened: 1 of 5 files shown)


ddiroma:
@pentaho/r2d2 @pminutillo @pentaho-mracine 

Wingman is complaining about indentation, but it is correct.

NJtwentyone:
line 198 in GetJobStatusServlet.java is past the normal 80 character limit. seems like it's the same for others @ddiroma

engine/src/main/java/org/pentaho/di/www/GetJobStatusServlet.java
```diff
@@ -195,6 +195,8 @@ public void doGet( HttpServletRequest request, HttpServletResponse response ) th
 
     String jobName = request.getParameter( "name" );
     String id = request.getParameter( "id" );
+    String root = request.getRequestURI() == null ? StatusServletUtils.PENTAHO_ROOT : request.getRequestURI().substring( 0, request.getRequestURI().indexOf( CONTEXT_PATH ) );
```
NJtwentyone: line is past 80 line character limit. same for others

ddiroma: @NJtwentyone There are dozens of lines in the servlets for carte that are longer than 80...for some reason, wingman/checkstyle doesn't care

NJtwentyone: It still needs to be human readable. I would only worry about the actually variable assigned lines (ie all code besides the "out.println(..." ) that this PR changed. The numerous out.println statements makes sense to break the 80 line limit to make the xml readable.

engine/src/main/java/org/pentaho/di/www/StatusServletUtils.java
```diff
@@ -30,6 +30,10 @@
 
 public class StatusServletUtils {
 
+  public static final String RESOURCES_PATH = "/content/common-ui/resources/themes";
+  public static final String STATIC_PATH = "/static";
+  public static final String PENTAHO_ROOT = "/pentaho";
```
pedrofvteixeira: This would still cause problems ( i.e. broken links ) on the following 2 situations:
- _"old"_ DI-Servers ( a.k.a ) flavourized Pentaho Servers, that look-&-feel as the "old" DI-Servers
- OEM specific webapp naming; if the webapp gets rebranded from `pentaho` to `my-oem-webapp`, `kettle/status` page icons could potentially cease to work

Pentaho platform, for example, provides the `fully-qualified-server-url` property via `server.properties` ( which is a user configurable file ), for customers to tweak / change the base server endpoint to be used for things such as static resources.

https://github.com/pentaho/pentaho-platform/blob/8.2.0.0-RC/assemblies/pentaho-solutions/src/main/resources/pentaho-solutions/system/server.properties#L3

Also providing a one-liner to fetch it: `PentahoSystem.getApplicationContext().getFullyQualifiedServerURL();`
- Example: https://github.com/pentaho/pentaho-platform/blob/8.2.0.0-RC/extensions/src/main/java/org/pentaho/platform/web/servlet/PluginDispatchServlet.java#L356

Could it be that, when `!isJettyMode()` , `prefix` should be `PentahoSystem.getApplicationContext().getFullyQualifiedServerURL()` ?

Reaching out to @mbatchelor @pamval on this for confirmation / double-checking.

ddiroma: @pedrofvteixeira -- to speak to your second concern, your suggestion is probably a better implementation. However, I do address this concern by getting the "base server endpoint" as follows:

String root = request.getRequestURI() == null ? StatusServletUtils.PENTAHO_ROOT : request.getRequestURI().substring( 0, request.getRequestURI().indexOf( CONTEXT_PATH ) );
String prefix = isJettyMode() ? StatusServletUtils.STATIC_PATH : root + StatusServletUtils.RESOURCES_PATH;

NJtwentyone: @pedrofvteixeira @ddiroma are all concerns addressed? Can someone hit the "resolve conversation" button.

pedrofvteixeira: Good by me.

engine/src/main/java/org/pentaho/di/www/GetJobStatusServlet.java
NJtwentyone: for the ternary operators I was looking for something like this
`
String root = request.getRequestURI() == null
  ? StatusServletUtils.PENTAHO_ROOT
  : request.getRequestURI().substring( 0, request.getRequestURI() 
     .indexOf( CONTEXT_PATH ) );
`
It's not a big deal now, there are other issues with the code style for this file as you mentioned.

NJtwentyone:
all comments resolved.

pentaho-lmartins:
@NJtwentyone @pentaho/r2d2 please don't merge PRs with check style violations.
This is a really bad practice that won't bring us any good.

ddiroma:
@pentaho-lmartins This checkstyle violation was bogus. I noted that in the PR. The line in question had correct indentation...it was a wingman glitch.

PR #6091, opened 2018-12-14
[PDI-17622] Build Model step works fine in Spoon but fails from Kitchen

ppatricio:
This is a race condition issue with 9 fails in 10 in my environment. The JobEntryType (of BuildModel for example) is not registered when we try to run it in Kitchen and the job ends unsuccessfully due to plugin missing. This PR complements that logic, providing a way of giving more time for that plugin registration process, instead of aborting the execution immediately.  A waitLimit flag is provided, preventing looping issues. Should this be a Const? Configured and visible to users?

This is a wait to get fix. Probably a more architectural approach is needed to ensure we have all plugins we need (or know at least they are already installed in Karaf and then wait for them to be active) before start running those objects in kitchen. But at the same time we should keep our code separated from how OSGI/Karaf bundle management works.
Spoon works fine.

This issue (PDI-17622) has another error solved in this PR,https://github.com/pentaho/pentaho-metadata/pull/199

For the overall fix, both should be considered.
@mkambol 
@mbatchelor 
@pedrofvteixeira
@pentaho-lmartins
@graimundo   
can you please review this?

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/plugins/PluginRegistry.java
+++ b/core/src/main/java/org/pentaho/di/core/plugins/PluginRegistry.java
@@ -94,6 +94,7 @@ public class PluginRegistry {
   private final Map<Class<? extends PluginTypeInterface>, Set<PluginTypeListener>> listeners = new HashMap<>();
 
   private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+  private static final int WAIT_FOR_PLUGIN_TO_BE_AVAILABLE_LIMIT = 3000;
 
 
   /**
@@ -1062,5 +1063,30 @@ public class PluginRegistry {
     } finally {
       lock.writeLock().unlock();
     }
+  }
+
+  public PluginInterface findPluginWithId( Class<? extends PluginTypeInterface> pluginType, String pluginId, boolean waitForPluginToBeAvailable ) {
+    PluginInterface pluginInterface = findPluginWithId(  pluginType,  pluginId );
+    return waitForPluginToBeAvailable && pluginInterface == null
+      ? waitForPluginToBeAvailable( pluginType,  pluginId, WAIT_FOR_PLUGIN_TO_BE_AVAILABLE_LIMIT )
+      : pluginInterface;
+  }
+
+  private PluginInterface waitForPluginToBeAvailable( Class<? extends PluginTypeInterface> pluginType, String pluginId, int waitLimit ) {
+    int timeToSleep = 50;
+    try {
+      Thread.sleep( timeToSleep );
+      waitLimit -= timeToSleep;
+    } catch ( InterruptedException e ) {
+      log.logError( e.getLocalizedMessage(), e );
+      Thread.currentThread().interrupt();
+      return null;
+    }
+    PluginInterface pluginInterface = findPluginWithId( pluginType, pluginId );
+    return  waitLimit <= 0 && pluginInterface == null
+      ? null
+      : pluginInterface != null
+        ? pluginInterface
+        : waitForPluginToBeAvailable( pluginType, pluginId, waitLimit );
   }
 }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/pentaho/di/core/plugins/PluginRegistry.java
```diff
@@ -94,6 +94,7 @@
   private final Map<Class<? extends PluginTypeInterface>, Set<PluginTypeListener>> listeners = new HashMap<>();
 
   private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+  private int waitForPluginToBeAvailableLimit = 2000;
```
mkambol: Should be a static final constant, with upper case / underscore naming convention.

core/src/main/java/org/pentaho/di/core/plugins/PluginRegistry.java
```diff
@@ -1063,4 +1064,26 @@ public void reset() {
       lock.writeLock().unlock();
     }
   }
+
+  public PluginInterface findPluginWithId( Class<? extends PluginTypeInterface> pluginType, String pluginId, boolean waitForPluginToBeAvailable ) {
+    return waitForPluginToBeAvailable && findPluginWithId(  pluginType,  pluginId ) == null
```
mkambol: I'd pull the findPluginWithId call out of the conditional so you don't need to invoke it twice.

core/src/main/java/org/pentaho/di/core/plugins/PluginRegistry.java
```diff
@@ -1063,4 +1064,26 @@ public void reset() {
...
+      ? waitForPluginToBeAvailable( pluginType,  pluginId, waitForPluginToBeAvailableLimit )
+      : findPluginWithId(  pluginType,  pluginId );
+  }
+
+  public PluginInterface waitForPluginToBeAvailable( Class<? extends PluginTypeInterface> pluginType, String pluginId, int waitLimit ) {
```
mkambol: Can this method be private?

core/src/main/java/org/pentaho/di/core/plugins/PluginRegistry.java
```diff
@@ -1063,4 +1064,26 @@ public void reset() {
...
+    int timeToSleep = 1;
```
mkambol: timeToSleep of 1 seems aggressive- for a missing plugin we'll end up calling findPluginWithId 2000 times.  Maybe every 50 or 100 milliseconds?

core/src/main/java/org/pentaho/di/core/plugins/PluginRegistry.java
```diff
@@ -1063,4 +1064,26 @@ public void reset() {
...
+    try {
+      Thread.sleep( timeToSleep );
+      waitLimit -= timeToSleep;
+    } catch ( InterruptedException e ) {
+      throw new RuntimeException( "Waiting for Plugin to be Available interrupted!", e );
```
mkambol: Need to reset the interrupt flag (with Thread.currentThread().interrupt())
Also, I'd suggest logging the interruptedexception and returning null rather than rethrowing.  Or, if you need to rethrow, use a less generic runtime exception - maybe IllegalStateException.

core/src/main/java/org/pentaho/di/core/plugins/PluginRegistry.java
```diff
@@ -1063,4 +1064,31 @@ public void reset() {
...
+  }
+
+  private PluginInterface waitForPluginToBeAvailable( Class<? extends PluginTypeInterface> pluginType, String pluginId, int waitLimit ) {
+    int timeToSleep = 50;
+    try {
+      Thread.sleep( timeToSleep );
+      waitLimit -= timeToSleep;
+    } catch ( InterruptedException e ) {
+      log.logError( e.getLocalizedMessage(), e );
+      if ( Thread.currentThread().isInterrupted() ) {
```
mkambol: When Thread.sleep throws interruptedexception, the interrupted status is cleared.  So I would expect the condition on 1082 to always be false.
To reset the interrupt flag, just call `Thread.currentThread().interrupt();`.
That'll allow any callers up the chain in this thread to deal with it.

Good overview here:  https://www.yegor256.com/2015/10/20/interrupted-exception.html

PR #6092, opened 2018-12-16
[PDI-17764] Checkpoints dropped from Job on Export to XML

smmribeiro:


smmribeiro:
The export is based, not on the actual Job, but on a cloned Job.
The problem was that the attributes Map was not considered when cloning.

smmribeiro:
@pentaho-lmartins, I've cleaned the PR.

PR #6095, opened 2018-12-17
[PDI-17769] Copy Files Job entry - Connection to SFTP server with var…

smmribeiro:
…iables errors but copies files successfully


smmribeiro:
@pentaho-lmartins, I've cleaned the PR.

PR #6096, opened 2018-12-17
[PDI-17768] Get File Names step: vfs error when listing files from sf…

smmribeiro:
…tp site


smmribeiro:
VFS issue [VFS-617](https://issues.apache.org/jira/browse/VFS-617) has an associated [PR#27](https://github.com/apache/commons-vfs/pull/27) that fixes the problem, however it hasn't been commited (and doesn't seem that it'll in the near future) and there's also no plan for a new VFS version.

We don't have a VFS fork anymore but the SFTP step has some implementation that overrides the normal VFS code.

This PR got the code from the uncommited [PR#27](https://github.com/apache/commons-vfs/pull/27) to the SFTP implementation.

PR #6118, opened 2019-01-01
8.2

vandercijr:
I'm using Pentaho Kettle 8.2 on fedora 28 64 bits;

I get this error

"Unable to open dialog for this step
org.eclipse.swt.SWTError: No more handles [MOZILLA_FIVE_HOME="](java.lang.UnsatisfiedLinkError:Could not load SWT library. Reasons: no swt-mozilla-gtk-4335 in java.library.path..."

"...libxpcom.so: cannot open shared object file: No such file or directory..."

I've been looking for an answer but all of I have found do not solve this problem.

The last on was instal xulrunner and set MOZILLA_FIVE_HOME to xulrunner directory.


pamval:
Please open an issue at jira.pentaho.com or post a thread at the community forums: https://community.hitachivantara.com/community/products-and-solutions/pentaho

Did you see this link? http://diethardsteiner.blogspot.com/2014/02/installing-pentaho-kettle-depencencies.html

Seems to be the exact same problem you're having.

PR #6126, opened 2019-01-07
[BACKLOG-26963] widget is disposed error closing trans/job on swt4.9

tgf:


tgf:

PR #6129, opened 2019-01-07
[SP-4837] - jms consumer might lose messages on safe stopping

kurtwalker:


kurtwalker:
@ricardosilva88

ricardosilva88:
Hi @kurtwalker, please change the commit message to : "Backport of BACKLOG-26905 - JMS Consumer can lose messages on safe stop (8.2 Suite)".

It is the same that JIRA case title that I already correct also. Ty

PR #6134, opened 2019-01-10
[BACKLOG-26963] widget dispose error closing job on newer swt

tgf:


tgf:

PR #6139, opened 2019-01-11
[BACKLOG-27125] Preventive fix for when upgrading Linux SWT an extra …

JoaoRMF:
…window is show when opening the help dialog in the connect to repo


JoaoRMF:

PR #6155, opened 2019-01-17
[PPP-4312] .TXT files content are outdated (Report Designer and PDI)

pentaho-lmartins:
@pentaho/tatooine

### DIFF
--- a/assemblies/static/src/main/resources/README.txt
+++ b/assemblies/static/src/main/resources/README.txt
@@ -6,9 +6,9 @@ Installation instructions:
 
   OS X:
     - You can launch Spoon by navigating to the folder this "read me" file is located in and double clicking on the "Data Integration" application icon, depending on your system.
-  Ubuntu 12.04 and later:
+  Ubuntu 16.04 and later:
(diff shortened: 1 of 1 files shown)


assemblies/static/src/main/resources/README.txt
```diff
@@ -6,7 +6,7 @@ Installation instructions:
 
   OS X:
     - You can launch Spoon by navigating to the folder this "read me" file is located in and double clicking on the "Data Integration" application icon, depending on your system.
-  Ubuntu 12.04 and later:
+  Ubuntu 16.04 and later:
     - The libwebkitgtk package needs to be installed.  This can be done by running "apt-get install libwebkitgtk-1.0.0"
```
smmribeiro: Does not seem right: for me, as it is, this mean that if I have a 12.04 version, I don't need to install the libwebkitgtk package!
Or is the 16.04 the minimum version supported?

pentaho-lmartins: Done!

PR #6157, opened 2019-01-18
[BACKLOG-26961] linux swt 4.4+ extra window on db explore progress bar

tgf:


tgf:

PR #6166, opened 2019-01-22
[BACKLOG-26961] have different shell on createDbNodes whether dialog is open

tgf:


tgf:

PR #6192, opened 2019-01-30
[PDI-17843] Lookup Step Output Metadata not set correctly in 8.2

LeonardoCoelho71950:
@mbatchelor 

Unit test for ValueMetaBase.getMetadataPreview method.


core/src/test/java/org/pentaho/di/core/row/value/ValueMetaBaseTest.java
```diff
@@ -1111,4 +1117,368 @@ public void testHashCodes() throws Exception {
...
   }
+
+  @Test
+  public void testGetMetadataPreview() throws SQLException, KettleDatabaseException {
+    ValueMetaBase valueMetaBase = new ValueMetaBase( );
+
+    DatabaseMeta dbMeta = spy( new DatabaseMeta() );
+
+    // SQL CHAR is a Pentaho String
+    ResultSet resultSet = mock( ResultSet.class );
```
pentaho-lmartins: Use a private variable for `resultSet` and initialize it for every test in a `@Before` method.
This results in one less line for every test.

core/src/test/java/org/pentaho/di/core/row/value/ValueMetaBaseTest.java
```diff
@@ -1111,4 +1117,368 @@ public void testHashCodes() throws Exception {
     valueMetaString.type = ValueMetaInterface.TYPE_NONE;
     assertEquals( valueMetaString.hashCode( "any" ),  0 );
   }
+
+  @Test
+  public void testGetMetadataPreview() throws SQLException, KettleDatabaseException {
+    ValueMetaBase valueMetaBase = new ValueMetaBase( );
+
+    DatabaseMeta dbMeta = spy( new DatabaseMeta() );
```
pentaho-lmartins: Place this in a private variable and initialize for every test.

core/src/test/java/org/pentaho/di/core/row/value/ValueMetaBaseTest.java
```diff
@@ -1111,4 +1117,368 @@ public void testHashCodes() throws Exception {
...
+
+    // SQL CHAR is a Pentaho String
+    ResultSet resultSet = mock( ResultSet.class );
+    Mockito.when( resultSet.getInt( "DATA_TYPE" ) ).thenReturn( Types.CHAR );
```
pentaho-lmartins: Please use the syntax `doReturn( Types.CHAR ).when( resultSet ).getInt( "DATA_TYPE" )` instead.
This syntax has given use some problems in the past with tests failing for no apparent reason.
Also use an `import static` instead of `Mockito.method` this is customary for unit tests.

core/src/test/java/org/pentaho/di/core/row/value/ValueMetaBaseTest.java
```diff
@@ -1111,4 +1117,368 @@ public void testHashCodes() throws Exception {
...
+    ValueMetaInterface valueMeta = valueMetaBase.getMetadataPreview( dbMeta, resultSet );
+    Assert.assertTrue( valueMeta.isString() );
```
pentaho-lmartins: Same here `import static`.

core/src/test/java/org/pentaho/di/core/row/value/ValueMetaBaseTest.java
```diff
@@ -1111,4 +1117,368 @@ public void testHashCodes() throws Exception {
...
+
+    // SQL VARCHAR is a Pentaho String
```
pentaho-lmartins: Please write a unit test for each test, something like:
```
@Test
public void sqlVarcharToPentahoString() {
  ...
}
```
If one of these tests fail you'll know exactly when is wrong, instead of having to debug a single test.
This way you can also avoid each comment before the test, the method name will be self explanatory.

pentaho-lmartins:
Despite my comments above, this is exactly what we wanted, with a few tweaks will be perfect.

LeonardoCoelho71950:
Hi @pentaho-lmartins! I performed the requested changes. Furthermore, full coverage of the getMetadataPreview method is guaranteed.

PR #6229, opened 2019-02-07
PDI-17902

jonjarvis:
For the record, I didn't really like adding "data.inputRowMeta" to the data object of TextFileOutput. However, it was the only way to avoid AssertNotInvokedTwice for the call to getInputRowMeta() on the unit tests of TFO. I needed the TFO writeHeader() method to use the inputRowMeta instead of the outputRowMeta for the Concat Fields step, since Concat Fields updates data.outputRowMeta...

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/textfileoutput/TextFileOutput.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/textfileoutput/TextFileOutput.java
@@ -445,7 +445,8 @@ public class TextFileOutput extends BaseStep implements StepInterface {
     Object[] row = getRow(); // This also waits for a row to be finished.
 
     if ( row != null  && first ) {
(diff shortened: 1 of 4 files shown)


jonjarvis:
Rebased per @NJtwentyone request - Also, the reason I commented before regarding adding inputRowMeta to TFO was to pass a general unit test that ensures getInputRowMeta() is not called twice. I believe a good practice is to only ever call this method a single time from the body of if(first) {} in processRow.

NJtwentyone:
@jonjarvis  the commit message needs to follow the format "[PDI-XXX] <some message>"

NJtwentyone:
Also, can you combine the commit with message " Fix checkstyle issue" into an original commit? It doesn't seem necessary to have it as another message.

jonjarvis:
@NJtwentyone - Squashed and updated message

NJtwentyone:
All issues addressed.

PR #6251, opened 2019-02-14
[BACKLOG-27567] Centralize org.osgi dependencies management

nantunes:
@graimundo please review.


plugins/repositories/core/pom.xml
```diff
@@ -109,18 +108,6 @@
       <version>${javax.servlet-api.version}</version>
       <scope>provided</scope>
     </dependency>
-    <dependency>
```
graimundo: This is *not* in a dependency management. Double check removing the dependencies.

PR #6289, opened 2019-02-23
[BACKLOG-24264] PDI Service/Job supporting 'result-set' capability

neeta-rathi:
Bringing in POC work from @pedrofvteixeira's branch to master.

Verified and tested locally, works as expected. PDI-Service's `/ktr` endpoint now takes two new parameters `resultSetStepName` and `resultSetCopyNumber`

@pentaho/rogueone

### DIFF
--- a/engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
+++ b/engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2019 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
--- a/engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
+++ b/engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
@@ -22,15 +22,20 @@
 
 package org.pentaho.di.pan;
 
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.math.NumberUtils;
 import org.pentaho.di.base.AbstractBaseCommandExecutor;
 import org.pentaho.di.base.CommandExecutorCodes;
 import org.pentaho.di.core.Const;
 import org.pentaho.di.core.Result;
+import org.pentaho.di.core.RowMetaAndData;
 import org.pentaho.di.core.exception.KettleException;
+import org.pentaho.di.core.exception.KettleStepException;
 import org.pentaho.di.core.logging.LogChannel;
 import org.pentaho.di.core.logging.LogChannelInterface;
 import org.pentaho.di.core.parameters.NamedParams;
 import org.pentaho.di.core.parameters.UnknownParamException;
+import org.pentaho.di.core.row.RowMetaInterface;
 import org.pentaho.di.core.util.FileUtil;
 import org.pentaho.di.core.util.Utils;
 import org.pentaho.di.core.xml.XMLHandler;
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
```diff
@@ -163,9 +172,30 @@ public Result execute( String repoName, String noRepo, String username, String t
...
       }
 
+      final List<RowMetaAndData> rows = new ArrayList<RowMetaAndData>(  );
+
       // allocate & run the required sub-threads
       try {
-        trans.execute( arguments );
+        trans.prepareExecution( arguments );
+
+        if ( !StringUtils.isEmpty( resultSetStepName ) ) {
```
pdesai16: Why is it looking for stepName? When we run  a ktr, isn't it always returning the resultset returned by the last step?

pdesai16: When we run in spoon, we don't have to specify the step name to get the result.
Are we adding this functionality for testing intermediate steps?

pedrofvteixeira: No, not only from a last step. One can actually fetch a result set from any of the kettle steps.

ktr_sample


In the sample scenario we are using on BACKLOG-24264 ( copy-pasted from a vanilla CDA KTR sample ), the result set is actually being fetched from the penultimate step ( i.e. _"Report Columns"_ ).

On CDA XML files, one states which KTR to use as well as which step name:

cda

pdesai16: I know Spoon lets you do that for testing. Can we make the step name optional? If the steps name is provided, we will get the result from that step, else we get it from the last step

pedrofvteixeira: The "standard" request returns a JSON response with the ETL execution status:

```
{
  "result" : {
    "uuid" : "38ba2cfe-3487-4e80-b2ba-dfd326f74405",
    "inputFile" : "/public/print_received_params.ktr",
    "async" : "false",
    "username" : "admin",
    "status" : "COMPLETED"
  }
}
```

Now, by passing an **extra / optional** form param called `resultSetStepName`, the JSON response with the ETL execution status will carry something extra, which is the result-set:

```
{
  "result" : {
    "uuid" : "38ba2cfe-3487-4e80-b2ba-dfd326f74405",
    "inputFile" : "/public/print_received_params.ktr",
    "async" : "false",
    "username" : "admin",
    "status" : "COMPLETED"
    "resultSet": {
        "metadata": { (...) }
        "rows": { (...) }
     }
  }
}
```

> (...) Can we make the step name optional? If the steps name is provided, we will get the result from that step, else we get it from the last step (...)

To do that, 2 things would need to take place: 
1. One extra ( also optional ) form param such as "wantResultSet" : "true"
2. Create some logic to iterate over all existing steps, locate the "last" one, and use that one's step name 
   2.1. Note: step names are not "known" beforehand; we would need to go and iterate until we find whatever's the step name for the "last" of the steps.

Item 2 would carry additional effort / iteration logic.

My suggestion is that - until there is a use case for such a default/fallback `resultSet` logic, we provide it "as-is" nowadays for CDA and PRD, which is by requiring a "step name" to be passed along.

Thoughts ? @pdesai16 @neeta-rathi

pdesai16: I was thinking of simpler implementation, if the existing stepname parameter is provided, return the result of that step else, return the result of last step

My impression (which might be wrong) , it has the list of steps because it has to feed the result of one step as input to another. So we take the last step from that list and return that result of that step

pdesai16: If you look at the sppon interface, if you click on "execute" button, it return the result of last step 
if you select the step and click on execute, it returns the result of that step

PR #6299, opened 2019-02-27
[PDI-17922] Get File Names step - Only Files filter option works inco…

LeonardoCoelho71950:
…rrectly

@mbatchelor 

File type wasn't checked if there was no wildcard.

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/fileinput/FileInputList.java
+++ b/core/src/main/java/org/pentaho/di/core/fileinput/FileInputList.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2019 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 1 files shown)


smmribeiro:
I don't think this will fix the issue.

The problem seems to be in the handling of the wildcard. I did a quick test on 8.3-QAT-180 and the same happens if one sets "Only folders" or "All files".

LeonardoCoelho71950:
@smmribeiro Actually this bug is when you didn`t specified any wildcard. Right now we check the allowed types when we have a wildcard, but we don`t when we haven`t.

smmribeiro:
Yes, as I said, the problem is with the wildcard handling.
And, as the change is only working with the file type, that is why I have doubts it really fixes the issue.

LeonardoCoelho71950:
So, after talking to @smmribeiro, we realized that the bug described in https://jira.pentaho.com/browse/PDI-17922 is fixed. The question raised by him is a different one.
When we select a file in GetFileNames step, he was expecting only to be returned the folder's content, but it is returned also the select file (if the file is allowed). I wish to know if that is the expected behavior or not (@mbatchelor and @pentaho-lmartins, can you help me on that ).

LeonardoCoelho71950:
With or without wildcard the algorithm to get the fileInputList should be the same.
When onemask is empty, the file will always be included on the list.

PR #6322, opened 2019-03-13
[BACKLOG-27565] Centralize SLF4J dependency

nantunes:
Needs https://github.com/pentaho/maven-parent-poms/pull/111.

@graimundo

### DIFF
--- a/assemblies/core/lib/pom.xml
+++ b/assemblies/core/lib/pom.xml
@@ -504,7 +504,6 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
-            <version>${slf4j.version}</version>
         </dependency>
         <dependency>
             <groupId>org.glassfish.metro</groupId>
--- a/assemblies/core/lib/pom.xml
+++ b/assemblies/core/lib/pom.xml
@@ -39,8 +39,6 @@
         <sapdbc.version>7.4.4</sapdbc.version>
         <sqlite-jdbc.version>3.7.2</sqlite-jdbc.version>
         <geronimo-servlet_3.0_spec.version>1.0</geronimo-servlet_3.0_spec.version>
-        <slf4j-api.version>1.7.7</slf4j-api.version>
-        <slf4j-log4j12.version>1.7.7</slf4j-log4j12.version>
         <apache-log4j-extras>1.2.17</apache-log4j-extras>
         <webservices-api.version>2.3.1</webservices-api.version>
         <webservices-rt.version>2.3.1</webservices-rt.version>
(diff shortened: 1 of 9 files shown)


assemblies/core/lib/pom.xml
```diff
@@ -502,12 +500,11 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
-            <version>${slf4j-api.version}</version>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
-            <version>${slf4j-log4j12.version}</version>
+            <version>${slf4j.version}</version>
```
graimundo: The version is being managed at the parent pom level. No need to define a version here.

assemblies/lib/pom.xml
```diff
@@ -530,12 +528,11 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
-      <version>${slf4j-api.version}</version>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-log4j12</artifactId>
-      <version>${slf4j-log4j12.version}</version>
+      <version>${slf4j.version}</version>
```
graimundo: The version is being managed at the parent pom level. No need to define a version here.

plugins/pur/core/pom.xml
```diff
@@ -166,14 +164,12 @@
...
       <artifactId>slf4j-api</artifactId>
-      <version>${slf4j-api.version}</version>
-      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-log4j12</artifactId>
-      <version>${slf4j-log4j12.version}</version>
-      <scope>provided</scope>
+      <version>${slf4j.version}</version>
```
graimundo: The version is being managed in the parent pom. I'm not 100% sure but I think you don't need to declare it here just to change the scope.

plugins/pur/repository-proxy/pom.xml
```diff
@@ -56,8 +55,6 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
-      <version>${slf4j-api.version}</version>
-      <scope>provided</scope>
```
graimundo: Did you check if the jar continues not being placed in the pur plugin assembly / feature? Asking because you changed the scope of the dep.

nantunes: This is the osgi bundle made for dashboards-next, it doesn't affect the pdi-pur-plugin assembly. But yes, slf4j-api-1.7.7.jar is still there.

PR #6337, opened 2019-03-20
[CLEANUP] Use POJO from Pan and Kitchen to call it's corresponding CommandExecutors

neeta-rathi:
`Pan` and `Kitchen` uses `PanCommandExecutor` and `KitchenCommandExecutor` and passes all the parameters it receives as individual `String`.  This PR introduces pojos that represents `JobParams` and `TransParams` and makes changes to `Pan` and `Kitchen` to use pojos instead.

@pentaho/rogueone 

After merging this PR,  https://github.com/pentaho/pentaho-ee/pull/1929 needs to be merged as well ( re-trigger wingman )

### DIFF
--- a/engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
+++ b/engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2019 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
@@ -24,7 +24,6 @@ package org.pentaho.di.kitchen;
 
 import org.pentaho.di.base.AbstractBaseCommandExecutor;
 import org.pentaho.di.base.CommandExecutorCodes;
-import org.pentaho.di.base.IParams;
 import org.pentaho.di.core.Const;
 import org.pentaho.di.core.Result;
 import org.pentaho.di.core.exception.KettleException;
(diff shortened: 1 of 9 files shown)


engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
```diff
@@ -68,9 +71,7 @@ public KitchenCommandExecutor( Class<?> pkgClazz, LogChannelInterface log, Futur
     setKettleInit( kettleInit );
   }
 
-  public Result execute( String repoName, String noRepo, String username, String trustUser, String password, String dirName,
-                         String filename, String jobName, String listJobs, String listDirs, String exportRepo, String initialDir,
-                         String listRepos, String listParams, NamedParams params, NamedParams customParams, String[] arguments ) throws Throwable {
+  public Result execute( JobParams iParams ) throws Throwable {
```
pedrofvteixeira: **Nitpick**: can we rename `iParams` to just `params` ? It's an odd name for a java variable...

neeta-rathi: Offcourse..

engine/src/main/java/org/pentaho/di/pan/TransParams.java
```diff
@@ -0,0 +1,98 @@
...
+                localFile, localJarFile, localInitialDir, listRepos, safeMode, metrics, listFileParams, logLevel, maxLogLines,
+                maxLogTimeout, logFile, oldLogFile, version, resultSetStepName, resultSetCopyNumber, namedParams, null );
+    }
+
+    @Override
+    public Map<String, String> asMap() {
+
+        Map<String, String> arguments = new HashMap<>();
+
+        putIfNotEmpty( arguments, "uuid", getUuid() );
```
pedrofvteixeira: we can replace these `putIfNotEmpty( arguments, key, value );` lines with `arguments.putIfAbsent( key, value )` and forgo that private `putIfNotEmpty` method

PR #6340, opened 2019-03-22
[BACKLOG-27902] Add snowflake database type to puc with warehouse option

ricksjames:


assemblies/core/lib/pom.xml
```diff
@@ -465,6 +466,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
```
mkambol: We usually don't include jdbc driver implementations here-- there tend to be license issues with distribution.  If needed for a test you can just scope it.

core/src/main/java/org/pentaho/di/core/database/BaseDatabaseMeta.java
```diff
@@ -518,6 +519,23 @@ public void setServername( String servername ) {
     this.servername = servername;
   }
 
+  /**
+   * @return Returns the warehouse.
+   */
+  @Override
```
mkambol: Is there a way to avoid new get/setWarehouse methods?  It's not applicable to any other db that I'm aware of, so it stinks to have to add it to the base.

PR #6342, opened 2019-03-23
[BACKLOG-27098] Fix webpackages' URLs and module ids.

dcleao:
@pentaho/millenniumfalcon please review.

Depends on: https://github.com/pentaho/pentaho-osgi-bundles/pull/320.


plugins/file-open-save/core/src/main/resources-filtered/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -4,14 +4,6 @@
            xmlns:jaxrs="http://cxf.apache.org/blueprint/jaxrs" default-timeout="20000">
 
   <!-- HttpService Whiteboard registration of web assets -->
```
nantunes: This comment can also be removed.

PR #6372, opened 2019-03-31
[PDI-15742] Unable to load a prpt from the repository in the Pentaho Reporting Output step

neeta-rathi:
New implementation of `ResourceData` and `ResourceLoader` for Apache VFS's `FileObject` would load resource of type `FileObject`

@tmorgner @pentaho/rogueone 

To be merged with https://github.com/pentaho/pentaho-reporting/pull/1248


pedrofvteixeira:
PR looks good. 
We should create UTs / ITs to help prove out the fix, as well as to prevent future regressions.

pedrofvteixeira:
Merging, as the related one was also merged: pentaho/pentaho-reporting#1248

PR #6396, opened 2019-04-04
[BACKLOG-28311] NPE in PME, where there is no spoon

tgf:


tgf:

PR #6398, opened 2019-04-05
[PDI-17667] Use of vulnerable component jetty 8.1.15, CVE-2017-9735, …

smmribeiro:
…CVE-2017-7657 CVE-2015-2080

[PPP-4183] Use of Vulnerable Component: Multiple Jetty Components [Listed in Description] (CVE-2017-7656 | CVE-2017-7657 | CVE-2017-7658 | CVE-2015-2080 | CVE-2017-9735 )

Centralized Jetty components: use dependency management.
Cleared some unnecessary exclusions.
Some classes had to be updated to work the way Jetty now works.

Depends on pentaho/maven-parent-poms#121

@pentaho-lmartins @pentaho/tatooine
@graimundo @nantunes

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -80,12 +80,6 @@
     <mstor.version>0.9.13</mstor.version>
     <snappy-java.version>1.1.0</snappy-java.version>
     <commons-cli.version>1.2</commons-cli.version>
-    <jetty-plus.version>8.1.15.v20140411</jetty-plus.version>
-    <jetty-server.version>8.1.15.v20140411</jetty-server.version>
-    <jetty-security.version>8.1.15.v20140411</jetty-security.version>
-    <jetty-servlet.version>8.1.15.v20140411</jetty-servlet.version>
-    <jetty-xml.version>8.1.15.v20140411</jetty-xml.version>
-    <jetty-webapp.version>8.1.15.v20140411</jetty-webapp.version>
     <log4jdbc.version>1.2</log4jdbc.version>
     <javax.servlet-api.version>3.0.1</javax.servlet-api.version>
     <oro.version>2.0.8</oro.version>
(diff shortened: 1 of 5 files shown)


engine/pom.xml
```diff
@@ -760,19 +754,11 @@
...
-        </exclusion>
-      </exclusions>
+      <artifactId>jetty-jaas</artifactId>
     </dependency>
     <dependency>
       <groupId>org.eclipse.jetty</groupId>
       <artifactId>jetty-server</artifactId>
-      <version>${jetty-server.version}</version>
       <exclusions>
         <exclusion>
```
graimundo: @smmribeiro Does the new version still bring in these transitives? Meaning, do we still need this exclusions?

engine/pom.xml
```diff
@@ -930,6 +916,18 @@
       <version>${xmlunit.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <version>3.1.0</version>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
```
graimundo: @smmribeiro Why the exclusions?

smmribeiro: @graimundo, to be honest I don't know/remember why I added those exclusions... even more being a test scoped dependency.
I'll verify it.

engine/pom.xml
smmribeiro: Nope, seems not!
I'll remove it.

engine/pom.xml
smmribeiro: Resolving as I've updated the code.

smmribeiro:
Closing PR: to simplify Karaf's upgrade a different approach was defined!

PR #6420, opened 2019-04-11
[BACKLOG-28227] Expanded Pan functionality to accept and execute a BA…

singletonc:
…SE64 zip representation

### DIFF
--- a/engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
+++ b/engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
@@ -22,8 +22,12 @@
 
 package org.pentaho.di.base;
 
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -232,6 +236,28 @@ public abstract class AbstractBaseCommandExecutor {
 
   public boolean isEnabled( final String value ) {
     return YES.equalsIgnoreCase( value ) || Boolean.parseBoolean( value ); // both are NPE safe, both are case-insensitive
+  }
+
+  /**
+   * Decodes the provided base64String into the specified filePath. Parent directories must already exist.
+   *
+   * @param base64String String BASE64 representation of a file
+   * @param filePath     String The path to which the base64String is to be decoded
+   * @return File the newly created File
+   */
+  protected File decodeBase64StringToFile( String base64String, String filePath ) throws IOException {
+    //Decode base64String to byte[]
+    byte[] decodedBytes = Base64.getDecoder().decode( base64String );
+    File file = new File( filePath );
+
+    //Try-with-resources, write to file, ensure fos is always closed
+    try ( FileOutputStream fos = new FileOutputStream( file ) ) {
+      fos.write( decodedBytes );
+    } catch ( IOException e ) {
+      throw e;
+    }
+
+    return file;
   }
 
   public LogChannelInterface getLog() {
--- a/engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
+++ b/engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
@@ -52,6 +52,8 @@ import org.pentaho.di.trans.step.RowAdapter;
 import org.pentaho.di.trans.step.StepInterface;
 import org.w3c.dom.Document;
 
+import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Calendar;
@@ -127,7 +129,7 @@ public class PanCommandExecutor extends AbstractBaseCommandExecutor {
         // Try to load the transformation from file, even if it failed to load from the repository
         // You could implement some fail-over mechanism this way.
         if ( trans == null ) {
-          trans = loadTransFromFilesystem( p.getLocalInitialDir(), p.getLocalFile(), p.getLocalJarFile() );
+          trans = loadTransFromFilesystem( p.getLocalInitialDir(), p.getLocalFile(), p.getLocalJarFile(), p.getBase64Zip() );
         }
 
       }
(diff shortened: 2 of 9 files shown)


engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
```diff
@@ -234,6 +237,31 @@ public boolean isEnabled( final String value ) {
...
+    if ( deleteOnExit ) {
+      File file = new File( filePath );
+      file.deleteOnExit();
+    }
+
+    //Write to file
+    try ( FileOutputStream fos = new FileOutputStream( filePath ) ) {
+      fos.write( decodedBytes );
+    } catch ( Exception e ) {
+      System.out.println( " Exception writing decoded BASE64 string to file " + e.toString() );
```
pdesai16: Pls use the logger

engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
```diff
@@ -234,6 +237,31 @@ public boolean isEnabled( final String value ) {
...
+   * @param base64String String BASE64 representation of a file
+   * @param filePath     String The path to which the base64String is to be decoded
+   * @param deleteOnExit boolean Delete created file on JVM exit if true
+   */
+  protected void decodeBase64StringToFile( String base64String, String filePath, boolean deleteOnExit ) {
+    //Decode base64String to byte[]
+    byte[] decodedBytes = Base64.getDecoder().decode( base64String );
+
+    //Set file to delete on exit if true
+    if ( deleteOnExit ) {
```
pedrofvteixeira: if I pass `deleteOnExit` as `false` ... would it **not** create the java.io.File ?

engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
```diff
@@ -326,10 +327,18 @@ public Trans loadTransFromRepository( Repository repository, String dirName, Str
...
-  public Trans loadTransFromFilesystem( final String initialDir, final String filename, final String jarFilename ) throws Exception {
+  public Trans loadTransFromFilesystem( String initialDir, String filename, String jarFilename, String base64Zip )
+    throws Exception {
 
     Trans trans = null;
 
+    if ( !Utils.isEmpty( base64Zip ) ) {
+      //expected form of filename "*.ktr"
+      String zipName = java.util.UUID.randomUUID().toString() + ".zip";
+      filename = "zip:file:" + File.separator + File.separator + System.getProperty( "user.dir" ) + File.separator + zipName + "!" + filename;
```
pedrofvteixeira: Use `Const.getUserHomeDirectory()` instead of `System.getProperty( "user.dir" )`

https://github.com/pentaho/pentaho-kettle/blob/master/core/src/main/java/org/pentaho/di/core/Const.java#L2054-L2057

engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
singletonc: it will still create the file on disk, but it won't delete it. FileOutputStream didn't have a mechanism to cleanup the created file, so I used File in that case

engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
```diff
@@ -234,6 +237,31 @@ public boolean isEnabled( final String value ) {
...
   }
 
+  /**
+   * Decodes the provided base64String into the specified filePath. Parent directories must already exist.
+   *
+   * @param base64String String BASE64 representation of a file
+   * @param filePath     String The path to which the base64String is to be decoded
+   * @param deleteOnExit boolean Delete created file on JVM exit if true
+   */
+  protected void decodeBase64StringToFile( String base64String, String filePath, boolean deleteOnExit ) {
```
pedrofvteixeira: To make this method as reusable as possible, have it returning the `File` , instead of void.

That way you can even later one apply `File.getAbsolutePath()`, that you can use when creating that `zip:file://...` string.

engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
singletonc: those are not equivalent, fyi, but I modified the code such that it will work

engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
```diff
@@ -234,6 +237,28 @@ public boolean isEnabled( final String value ) {
...
+  protected File decodeBase64StringToFile( String base64String, String filePath ) {
+    //Decode base64String to byte[]
+    byte[] decodedBytes = Base64.getDecoder().decode( base64String );
+    File file = new File( filePath );
+
+    //Write to file
+    try ( FileOutputStream fos = new FileOutputStream( file ) ) {
+      fos.write( decodedBytes );
+    } catch ( Exception e ) {
+      getLog().logError( " Exception writing decoded BASE64 string to file " + e.toString() );
```
pedrofvteixeira: One of 2 things should occur here:
 - Opt A) throw the exception ( add it to method signature ); not the generic `Exception`, but the `IOException` one
 - Opt B) set file to `null`, so it ends up getting returned as a `null` object, rather than a corrupted one

If none of the above occur, and in a scenario where some exception takes please, the person leveraging this method would never know of it, and rely that the returned `File` object is a solid one.

engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
```diff
@@ -326,10 +327,18 @@ public Trans loadTransFromRepository( Repository repository, String dirName, Str
...
-  public Trans loadTransFromFilesystem( final String initialDir, final String filename, final String jarFilename ) throws Exception {
+  public Trans loadTransFromFilesystem( String initialDir, String filename, String jarFilename, String base64Zip )
+    throws Exception {
 
     Trans trans = null;
 
+    if ( !Utils.isEmpty( base64Zip ) ) {
+      //expected form of filename "*.ktr"
+      String zipPath = Const.getUserHomeDirectory() + File.separator + java.util.UUID.randomUUID().toString() + ".zip";
+      filename = "zip:file:" + File.separator + zipPath + "!" + filename;
```
pedrofvteixeira: Shouldn't this be 

```
"zip:file:" + File.separator + File.separator + zipPath + ... 
```

?

I see only 1 `File.separator` ...

engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
```diff
@@ -326,10 +327,18 @@ public Trans loadTransFromRepository( Repository repository, String dirName, Str
...
+      decodeBase64StringToFile( base64Zip, zipPath ).deleteOnExit();
```
pedrofvteixeira: In the eventuality the returned `File` object is null: NPE on `.deleteOnExit()`

engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
singletonc: yes, somehow it was working, I dunno, but I will change it

engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
singletonc: so I will check != null?

singletonc: and if null throw an exception?

singletonc: I think Pan will already complain if it cannot find the file 🤔 but I could provide more information as to where? is the error log enough?

singletonc: I will throw an exception in the method so it will never get to this point with a null file

engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
singletonc: I will go with A

PR #6422, opened 2019-04-11
[BACKLOG-28338] Add metadata injection support for CONNECTION NAME fi…

baudekin:
See https://jira.pentaho.com/browse/BACKLOG-28338


baudekin:
I tested the code I added. I did not go back and add tests for exiting methods.

PR #6428, opened 2019-04-15
[BACKLOG-28507] Refactored the PdiAction code that is making a connetion to

pdesai16:
repository and adding it as utility so it can be reused

### DIFF
--- a/engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
+++ b/engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
@@ -1,0 +1,164 @@
+/*!
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License, version 2.1 as published by the Free Software
+ * Foundation.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along with this
+ * program; if not, you can obtain a copy at http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html
+ * or from the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details.
+ *
+ * Copyright (c) 2019 Hitachi Vantara.  All rights reserved.
+ */
+
+package org.pentaho.di.repository;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.pentaho.di.core.Const;
+import org.pentaho.di.core.exception.KettleException;
+import org.pentaho.di.core.exception.KettleSecurityException;
+import org.pentaho.di.core.logging.LoggingBuffer;
+import org.pentaho.di.core.plugins.PluginRegistry;
+import org.pentaho.di.core.plugins.RepositoryPluginType;
+import org.pentaho.di.i18n.BaseMessages;
+import org.pentaho.platform.api.engine.ActionExecutionException;
+
+import java.io.ByteArrayInputStream;
+
+/**
+ *  Provides a utility method to connect to repository and return the repository object
+ */
+public class RepositoryConnectionUtils {
+  private static final Log logger = LogFactory.getLog( RepositoryConnectionUtils.class );
+  private static final String SINGLE_DI_SERVER_INSTANCE = "singleDiServerInstance";
+  private static Class<?> PKG = RepositoryConnectionUtils.class;
+
+
+  /**
+   * Connects to the PDI repository. If the  system settings "SINGLE_DI_SERVER_INSTANCE" is set, it will connect to
+   * repository using the embedded repository.xml otherwise it will connect to repository with the given name.
+   *
+   * @param  String RepositoryName
+   * @param  boolean Checking if the property "singleDiServerInstance" in PentahoSystems is set
+   * @param String Repo user name
+   * @param String Fully qualified server Url used for connecting to the repository
+   * @param LoggingBuffer
+   * @return  Repository
+   * @throws KettleException
+   * @throws KettleSecurityException
+   * @throws ActionExecutionException
+   */
+  public static Repository connectToRepository( final String repositoryName, boolean isSingleDiServerInstance,
+             String userName, String fullyQualifiedServerUrl, LoggingBuffer logBuffer ) throws KettleSecurityException,
+             KettleException, ActionExecutionException {
+
+    if ( logger.isDebugEnabled() ) {
+      logger.debug( "Creating Meta-repository" );
+    }
+
+    RepositoriesMeta repositoriesMeta = new RepositoriesMeta();
+
+    if ( logger.isDebugEnabled() ) {
+      logger.debug( " Populating Meta repository" );
+    }
+
+    try {
+      if ( isSingleDiServerInstance ) {
+        if ( logger.isDebugEnabled() ) {
+          logger.debug( "singleDiServerInstance=true, loading default repository" );
+        }
+
+        // Get the embedded repository xml
+        String repositoriesXml = getEmbeddedDefaultRepositoryXml( fullyQualifiedServerUrl );
+
+        ByteArrayInputStream sbis = new ByteArrayInputStream( repositoriesXml.getBytes( Const.XML_ENCODING ) );
+        repositoriesMeta.readDataFromInputStream( sbis );
+      } else {
+        // TODO: add support for specified repositories.xml files...
+        repositoriesMeta.readData(); // Read from the default $HOME/.kettle/repositories.xml file.
+      }
+    } catch ( Exception e ) {
+      throw new ActionExecutionException( BaseMessages.getString(
+        "RepositoryConnectionUtils.ERROR_0018_META_REPOSITORY_NOT_POPULATED" ), e ); //$NON-NLS-1$
+    }
+
+    if ( logger.isDebugEnabled() ) {
+      logger.debug( " Finding repository metadata" );
+    }
+    // Find the specified repository.
+    RepositoryMeta repositoryMeta = null;
+    try {
+      if ( isSingleDiServerInstance ) {
+        repositoryMeta = repositoriesMeta.findRepository( SINGLE_DI_SERVER_INSTANCE );
+      } else {
+        repositoryMeta = repositoriesMeta.findRepository( repositoryName );
+      }
+
+
+    } catch ( Exception e ) {
+      throw new ActionExecutionException( BaseMessages.getString( PKG,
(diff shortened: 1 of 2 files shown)


pdesai16:
@pentaho/rogueone  Please review

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
```diff
@@ -0,0 +1,179 @@
...
+    try {
+      if ( isSingleDiServerInstance ) {
+        if ( logger.isDebugEnabled() ) {
+          logger.debug( "singleDiServerInstance=true, loading default repository" ); //$NON-NLS-1$
+        }
+
+        // Get the embedded repository xml
+        String repositoriesXml = getEmbeddedDefaultRepositoryXml( fullyQualifiedServerUrl );
+
+        ByteArrayInputStream sbis = new ByteArrayInputStream( repositoriesXml.getBytes( "UTF8" ) );
```
pedrofvteixeira: See if we can use https://github.com/pentaho/pentaho-kettle/blob/master/core/src/main/java/org/pentaho/di/core/Const.java#L393

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
```diff
@@ -0,0 +1,179 @@
...
+      + "<repository_location_url>" + fullyQualifiedServerUrl
+      + "</repository_location_url>" //$NON-NLS-1$ //$NON-NLS-2$
+      + "<version_comment_mandatory>N</version_comment_mandatory>" //$NON-NLS-1$
+      + "</repository>" //$NON-NLS-1$
+      + "</repositories>"; //$NON-NLS-1$
+
+    return repositoriesXml;
+  }
+
+  protected static boolean hasExecutePermission(IAuthorizationPolicy authorizationPolicy) {
```
pedrofvteixeira: checkstyle

singletonc:
If this is refactoring/centralizing code from PDIAction, should not some code be removed from there?

engine/src/main/resources/org/pentaho/di/repository/messages/messages_en_US.properties
```diff
@@ -121,3 +121,14 @@ ExportFeedback.Message.RuleViolated=   [{0}]
 
 Repository.Lost.Error.Message=Unable to connect to the repository. The repository may not exist or is not operational at this time. Verify your repository and network settings and try again.
 Repository.Lost.Error.Preface=You don''t seem to be getting a connection to the repository.\r\nAny unsaved changes will be lost.
+
+RepositoryConnectionUtils.DEBUG_META_REPOSITORY=creating meta-repository
+RepositoryConnectionUtils.DEBUG_POPULATING_META=populating meta-repository
+RepositoryConnectionUtils.DEBUG_FINDING_REPOSITORY=Finding repository metadata
+RepositoryConnectionUtils.DEBUG_GETTING_REPOSITORY=Getting repository instance
+RepositoryConnectionUtils.DEBUG_CONNECTING=Connecting to repository
+RepositoryConnectionUtils.DEBUG_FINDING_DIRECTORY=Finding directory
```
pedrofvteixeira: Don't believe "debug" logs would need localizing...

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
```diff
@@ -0,0 +1,179 @@
...
+   */
+  public static Repository connectToRepository( final String repositoryName, boolean isSingleDiServerInstance,
+             IAuthorizationPolicy authorizationPolicy, String userName,
+             String fullyQualifiedServerUrl, LoggingBuffer logBuffer ) throws KettleSecurityException, KettleException,
+              ActionExecutionException {
+
+    if ( !hasExecutePermission( authorizationPolicy ) ) {
+      throw new IllegalStateException(
+        BaseMessages.getString( "RepositoryConnectionUtils.ERROR_0001_NO_PERMISSION_TO_EXECUTE" ) ); //$NON-NLS-1$
+    }
```
pedrofvteixeira: Alternatively, this util method - whose sole purpose is to establish a Repository connection - doesn't do a check for `hasExecutePermission( authorizationPolicy )` , as it does not pertain its realm of responsibilities.  

Other blocks of code which would leverage this `RepositoryConnectionUtils. connectToRepository()` would be the ones tasked with checking if a certain `username` has execute permissions or not.

pedrofvteixeira:
> Naive Question: If this is refactoring/centralizing code from PDIAction, should not some code be removed from there?
> 
> Update: Oh that's from another repo :) carry on!

@singletonc yes, that'll take place in a follow-up PR; 
Other locations ( such as PdiAction and WorkerNodes ) will replace its code with a call to this one centralised method.

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
```diff
@@ -0,0 +1,178 @@
...
+  private static final String SINGLE_DI_SERVER_INSTANCE = "singleDiServerInstance";
+  private static Class<?> PKG = RepositoryConnectionUtils.class;
+
+
+  /**
+   * Connects to the PDI repository. If the  system settings "SINGLE_DI_SERVER_INSTANCE" is set, it will connect to
+   * repository using the embedded repository.xml otherwise it will connect to repository with the given name.
+   *
+   * @param  String RepositoryName
+   * @param LoggingBuffer
```
befc: Params in the javadoc don't match the actual parameters of the method

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
```diff
@@ -0,0 +1,178 @@
...
+    } catch ( Exception e ) {
+      throw new ActionExecutionException( BaseMessages.getString(
+        "RepositoryConnectionUtils.ERROR_0018_META_REPOSITORY_NOT_POPULATED" ), e ); //$NON-NLS-1$
+    }
+
+    if ( logger.isDebugEnabled() ) {
+      logger.debug( " Finding repository metadata" );
+    }
+    // Find the specified repository.
+    RepositoryMeta repositoryMeta = null;
```
befc: **Intellij warning:** Initializing this variable to null value is redundant

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
```diff
@@ -0,0 +1,178 @@
...
+        logger.debug( logBuffer.getBuffer().toString() );
+      }
+      throw new ActionExecutionException( BaseMessages.getString( PKG,
+        "RepositoryConnectionUtils.ERROR_0004_REPOSITORY_NOT_FOUND", repositoryName ) ); //$NON-NLS-1$
+    }
+
+    if ( logger.isDebugEnabled() ) {
+      logger.debug( " Getting repository instance " ); //$NON-NLS-1$
+    }
+    Repository repository = null;
```
befc: **Intellij warning:** Initializing this variable to null value is redundant

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    return repository;
+  }
+
+  private static String getEmbeddedDefaultRepositoryXml( final String fullyQualifiedServerUrl ) {
+    // only load a default enterprise repository. If this option is set, then you cannot load
+    // transformations or jobs from anywhere but the local server.
+
+    String repositoriesXml = "<?xml version=\"1.0\" encoding=\"UTF-8\"?><repositories>" //$NON-NLS-1$
```
befc: **Intellij warning:** Initializing this variable is redundant, the same result can be done with:
*return "<?xml version=\"1.0\" encoding=\"UTF-8\"?><repositories>" ...*

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
pdesai16: It is surprising... My IntelliJ complains that the variables are not initialized

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
pdesai16: I can fix that

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
pdesai16: Same comment as above - I initialized it only because IntelliJ was complaining ..

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
befc: Can be kept as is, just a warning.
But with the current code bellow it always gets initialized ( both **if** and **else** do it ). Only flow it doesn't get initialized is the catch case, but as the exception is thrown, the flow doesn't continue

befc: Attaching screenshot: Screenshot 2019-04-15 at 17 51 37

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
befc: same as above, not a blocker, although it does seem to be redundant and may induce in error

befc: Attaching screenshot: 
Screenshot 2019-04-15 at 17 53 00

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
pdesai16: Fixed

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
befc: The correct javadoc syntax should the name of the param instead of the type:
```
   * @param repositoryName RepositoryName
   * @param isSingleDiServerInstance Checking if the property "singleDiServerInstance" in PentahoSystems is set
   * @param authorizationPolicy  AuthorizationPolicy
   * @param userName Repo user name
   * @param fullyQualifiedServerUrl Fully qualified server Url used for connecting to the repository
   * @param logBuffer
```
Also the first two have an extra space after **@param**

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
pdesai16: I thought we discussed it and decided to keep the check.

engine/src/main/java/org/pentaho/di/repository/RepositoryConnectionUtils.java
pdesai16: localVariablesScreenshot

PR #6437, opened 2019-04-16
[BACKLOG-28227] Unit Test

singletonc:


pentaho-mracine:
Nah I think Charles just updated it too quickly ;) I'll merge it as long as wingman gives us its shaky thumbs up

PR #6442, opened 2019-04-17
[CLEANUP] Refactor Params classes to use builder pattern

befc:
- **Params** is no longer an abstract class, and follows the **builder pattern**
   - Removed **TransParams** and **JobParams** since they are no longer required
Moved the **asMap** method from **TransParams** and **JobParams** to **KettleConstants** as **toJobMap()** and **toTransMap()**

Needs to be merged together with https://github.com/pentaho/pentaho-ee/pull/2024


pedrofvteixeira:
On `KitchenCommandExecutor`, we have:
 - `public Result execute( Params params ) throws Throwable {`

And on `PanCommandExecutor`, we have:
-  `public Result execute( Params p ) throws Throwable {`


As we're doing this cleanup work, could we go ahead - just for consistency sake - and have the variable named the same way on both locations ? Either as `params` or as `p` , I'd be fine with either

befc:
@pedrofvteixeira I will fix this and the checkstyle in a new commit

PR #6445, opened 2019-04-18
[PDI-17977] Set Variables in Job as blank returns null error if varia…

LeonardoCoelho71950:
…ble scope type is set to: Valid in the current job

ConcurrentHashMap does not allow null keys and null values.

@pentaho/tatooine @pentaho-lmartins @ppatricio 
@bantonio82 Can you take a look at this, since this is similar to PDI-17536?

### DIFF
--- a/engine/src/main/java/org/pentaho/di/job/entries/setvariables/JobEntrySetVariables.java
+++ b/engine/src/main/java/org/pentaho/di/job/entries/setvariables/JobEntrySetVariables.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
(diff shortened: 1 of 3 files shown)


ppatricio:
Hey @LeonardoCoelho71950 , could you just add some tests to cover this case?

LeonardoCoelho71950:
> Hey @LeonardoCoelho71950 , could you just add some tests to cover this case?

Sure :)

LeonardoCoelho71950:
I've included the test. 
There was also a missing check for variableName, since this is a minor fix, I've included it here.

PR #6447, opened 2019-04-18
[BACKLOG-28228] As a PDI on Foundry developer, I want Kitchen to supp…

pedrofvteixeira:
…ort running ETL inside zip files, so that a containerized PDI can process ETL inside a zip

@pentaho/rogueone


engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
```diff
@@ -298,13 +300,31 @@ public Job loadJobFromRepository( Repository repository, String dirName, String
...
 
-  public Job loadJobFromFilesystem( final String initialDir, final String filename ) throws Exception {
+  public Job loadJobFromFilesystem( String initialDir, String filename, String base64Zip ) throws Exception {
 
     if ( Utils.isEmpty( filename ) ) {
       System.out.println( BaseMessages.getString( getPkgClazz(), "Kitchen.Error.canNotLoadJob" ) );
       return null;
     }
 
+    if ( !Utils.isEmpty( base64Zip ) ) {
```
befc: Suggestion: promote this chunk of code to an util class to be re-used in both (...)CommandExecutors classes

pedrofvteixeira: Agree. I'll issue a follow-up Cleanup PR promoting what I can, for re-use at Pan and Kitchen

singletonc: Follow up PR will be part of BACKLOG-28228?

engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
```diff
@@ -298,13 +300,31 @@ public Job loadJobFromRepository( Repository repository, String dirName, String
...
+      String zipPath = Const.getUserHomeDirectory() + File.separator + java.util.UUID.randomUUID().toString() + ".zip";
+      filename = "zip:file:" + File.separator + File.separator + zipPath + "!" + filename;
+      File zipFile;
+
+      //responsibly attempt to write to file
+      try {
+        zipFile = decodeBase64StringToFile( base64Zip, zipPath );
+      } catch ( IOException e ) {
+        getLog().logError( e.toString() + "\n" );
+        e.printStackTrace();
```
pdesai16: Do we need to log the stackTrace .. we are loggin the execption string a line before

engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
```diff
@@ -126,7 +128,7 @@ public Result execute( Params params ) throws Throwable {
         if ( job == null ) {
 
           // Try to load the job from file, even if it failed to load from the repository
-          job = loadJobFromFilesystem( params.getLocalInitialDir(), params.getLocalFile() );
+          job = loadJobFromFilesystem( params.getLocalInitialDir(), params.getLocalFile(), params.getBase64Zip() );
```
pdesai16: Wondering if we still need to use the getLocalFile() or can we use the inputFile in which case we may not need to send this parameter from PentahoServer

PR #6471, opened 2019-04-30
[SP-5037] Backport of PDI-17977 - Set Variables in Job as blank retur…

LeonardoCoelho71950:
…ns null error if variable scope type is set to: Valid in the current job (8.2 Suite)

@ssamora @RPAraujo


usbrandon:
nice job.

PR #6507, opened 2019-05-07
[PDI-18033] PDI sample Get File Names - Remove extention.ktr misspelled

cdeptula:
@pentaho-lmartins @pentaho/tatooine


cdeptula:
@smmribeiro I think that makes a lot of sense.  You are correct the "Get File Names - Remove extension from short filename" does not remove the extension.  It instead changes the filename.

smmribeiro:
@cdeptula I think it makes sense doing it under this one.
Can you do it (pushing or force-pushing to this same branch) or do you want me to do it?

smmribeiro:
@cdeptula , renamed the files (and, internally, the samples' names) and pushed to your branch.

PR #6568, opened 2019-05-22
[PDI-17863] Borders are missing in kettle/status page after we perfor…

ddiroma:
…m an upgrade to 8.2 version.

Master: https://github.com/pentaho/pentaho-kettle/pull/6565
8.3.0.0: https://github.com/pentaho/pentaho-kettle/pull/6569

PR #6569, opened 2019-05-22
[PDI-17863] Borders are missing in kettle/status page after we perfor…

ddiroma:
…m an upgrade to 8.2 version.

Master: #6565
8.3: https://github.com/pentaho/pentaho-kettle/pull/6568


ddiroma:
@dkincade @kcruzada

PR #6570, opened 2019-05-22
[PDI-18039] Changing the web application name breaks the display on the kettle/status page.

jpruiz114:
[PDI-18039] Adding work in progress
[PDI-18039] Removing additional double quote from href path

@pentaho-lmartins

CC @gryphendowre  @wseyler

### DIFF
--- a/engine/src/main/java/org/pentaho/di/www/StatusServletUtils.java
+++ b/engine/src/main/java/org/pentaho/di/www/StatusServletUtils.java
@@ -24,6 +24,7 @@ package org.pentaho.di.www;
 
 import org.w3c.dom.Document;
 
+import javax.servlet.http.HttpServletRequest;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import java.io.File;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/pentaho/di/www/StatusServletUtils.java
```diff
@@ -34,7 +35,9 @@
   public static final String STATIC_PATH = "/static";
   public static final String PENTAHO_ROOT = "/pentaho";
 
-  public static String getPentahoStyles() {
+  public static String getPentahoStyles( HttpServletRequest request ) {
+    String root = request.getRequestURI() == null ? StatusServletUtils.PENTAHO_ROOT : request.getRequestURI().substring( 0, request.getRequestURI().indexOf( GetStatusServlet.CONTEXT_PATH ) );
```
pentaho-lmartins: Please use a variable for `uri` so you don't repeat `request.getRequestURI()` so many times.
This would make it easier to read such a long line.

pentaho-lmartins: Can we call this `webapp` instead of `root`?

engine/src/main/java/org/pentaho/di/www/StatusServletUtils.java
```diff
@@ -89,9 +92,9 @@ public static String getPentahoStyles() {
       // log here
     }
 
-    sb.append( "<link rel=\"stylesheet\" type=\"text/css\" href=\"/pentaho/content/common-ui/resources/themes/" + themeName + "/" + themeCss + "\"/>" );
-    sb.append( "<link rel=\"stylesheet\" type=\"text/css\" href=\"/pentaho/mantle/themes/" + themeName + "/" + mantleThemeCss + "\"/>" );
-    sb.append( "<link rel=\"stylesheet\" type=\"text/css\" href=\"/pentaho/mantle/MantleStyle.css\"/>" );
+    sb.append( "<link rel=\"stylesheet\" type=\"text/css\" href=\"" + root + "/content/common-ui/resources/themes/" + themeName + "/" + themeCss + "\"/>" );
```
pentaho-lmartins: Please use multiple `.appen()` calls instead of concatenating string and then appending the result to the buffer.

jpruiz114: Let me know if the change added makes sense. Thanks.

jpruiz114:
This was solved by PDI-17863.

PR #6591, opened 2019-05-27
[PDI-17956] fixed fields toggle between Edit/Add for "Decrypt files w…

NJtwentyone:
…ith PGP" and "Encrypt files with PGP"

- correct indices for Edit listener logic
- added UserID and Action Field for Encrypt PGP
- added Passphrase for Decrypt PGP

### DIFF
--- a/ui/src/main/java/org/pentaho/di/ui/job/entries/pgpencryptfiles/JobEntryPGPEncryptFilesDialog.java
+++ b/ui/src/main/java/org/pentaho/di/ui/job/entries/pgpencryptfiles/JobEntryPGPEncryptFilesDialog.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2019 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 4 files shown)


NJtwentyone:
@pentaho/r2d2 please review

ui/src/main/java/org/pentaho/di/ui/job/entries/pgpencryptfiles/JobEntryPGPEncryptFilesDialog.java
```diff
@@ -520,7 +557,7 @@ public void widgetSelected( SelectionEvent e ) {
     wSourceFileFolder.addModifyListener( lsMod );
     FormData fdSourceFileFolder = new FormData();
     fdSourceFileFolder.left = new FormAttachment( middle, 0 );
-    fdSourceFileFolder.top = new FormAttachment( wSettings, 2 * margin );
+    fdSourceFileFolder.top = new FormAttachment( wAction, margin );
```
cseverino789: Margin transformation (2 * margin) was removed on line 560 but kept on 501.  Is this Intentional?

NJtwentyone: good catch, the label was barely noticeable off center. It will be corrected with the next push.

PR #6594, opened 2019-05-28
[PPP-4374] Random issues with Karaf - Race Condition

ppatricio:
PR with a "wait for provider" solution in KettleVFS:
1-added a wait for provider mechanism
2-check for big data providers, that were wrongly defined as relativePathFilenames

There was also a problem with some trans executions, that required a change in StepMeta following the https://github.com/pentaho/pentaho-kettle/pull/6091 approach.

For 7.1 we have to backport the https://github.com/pentaho/pentaho-kettle/pull/6091 first to use this fix.

@mkambol

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/vfs/KettleVFS.java
+++ b/core/src/main/java/org/pentaho/di/core/vfs/KettleVFS.java
@@ -60,6 +60,8 @@ public class KettleVFS {
 
   private static final KettleVFS kettleVFS = new KettleVFS();
   private final DefaultFileSystemManager fsm;
+  private static final int timeOutLimit = 9000;
+  private static final int timeToSleepStep = 50;
 
   private static VariableSpace defaultVariableSpace;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/pentaho/di/core/vfs/KettleVFS.java
```diff
@@ -129,12 +131,23 @@ public static FileObject getFileObject( String vfsFilename, VariableSpace space,
...
+
+      int timeOut = timeOutLimit;
+      boolean hasScheme = vfsFilename != null && vfsFilename.contains( "://" ); // check for bigData providers
+
+      while ( relativeFilename && hasScheme && timeOut > 0 ) {
+        String[] schemes = fsManager.getSchemes();
+        try {
+          Thread.sleep( timeToSleepStep );
+          timeOut -= timeToSleepStep;
+          relativeFilename = checkForScheme( schemes, relativeFilename, vfsFilename, space, fsOptions );
```
mkambol: Could this while loop just check repeatedly for whether the scheme is available from fsManager.getSchemes()?

The .checkForSchemes() seems to be doing too much to me-- looping each time to find whether or not the scheme is present, and then calling buildFsOptions.

core/src/main/java/org/pentaho/di/core/vfs/KettleVFS.java
```diff
@@ -162,6 +175,19 @@ public static FileObject getFileObject( String vfsFilename, VariableSpace space,
...
   }
 
+  protected static boolean checkForScheme( String[] initialSchemes, boolean relativeFilename, String vfsFilename,
+                                         VariableSpace space, FileSystemOptions fsOptions )
+    throws IOException {
+    for ( int i = 0; i < initialSchemes.length && relativeFilename; i++ ) {
+      if ( vfsFilename.startsWith( initialSchemes[ i ] + ":" ) ) {
+        relativeFilename = false;
+        // We have a VFS URL, load any options for the file system driver
+        fsOptions = buildFsOptions( space, fsOptions, vfsFilename, initialSchemes[ i ] );
```
mkambol: I had to look at this for a couple minutes to realize that buildFsOptions is mutating the fsOptions.  The .checkForScheme method doesn't return fsOptions or do anything else with it, so there isn't a need to assign.

core/src/main/java/org/pentaho/di/core/vfs/KettleVFS.java
```diff
@@ -129,12 +131,23 @@ public static FileObject getFileObject( String vfsFilename, VariableSpace space,
       // If not, we are going to assume it's a file.
       //
       boolean relativeFilename = true;
-      String[] schemes = fsManager.getSchemes();
-      for ( int i = 0; i < schemes.length && relativeFilename; i++ ) {
-        if ( vfsFilename.startsWith( schemes[ i ] + ":" ) ) {
+      String[] initialSchemes = fsManager.getSchemes();
+
+      relativeFilename = checkForScheme( initialSchemes, relativeFilename, vfsFilename, space, fsOptions );
```
mkambol: What does `relativeFilename` mean?  Looks like checkForScheme returns false if the files scheme is present, not sure what makes that "relative".  Also, why false?
Maybe some comments or javadoc could clarify.

PR #6611, opened 2019-05-29
[PDI-18090] - forcing isCoalescing to be false, which is supposed to …

kurtwalker:
…be the default.  some implementations were leaving it null


kurtwalker:
Master PR https://github.com/pentaho/pentaho-kettle/pull/6604
8.3.0.0 PR https://github.com/pentaho/pentaho-kettle/pull/6610

PR #6618, opened 2019-05-29
[PDI-18092] Call endpoint step has null pointer exception in log even…

ddiroma:
… though the request was successful


ddiroma:
@bmorrise

PR #6633, opened 2019-05-30
[BACKLOG-27345] - changes to redshift database meta to support login …

kurtwalker:
…with iam credentials


kurtwalker:
hold for now

PR #6647, opened 2019-06-03
[BACKLOG-27839] Replace In String Metadata Injection Boolean Values E…

cseverino789:
…rror

Updated Replace in String Step, changing the following fields from int arrays to boolean arrays:
- use_regex
- whole_world
- case_sensitive
- is_unicode

Update fixes Metadata injection as well as, accepts "Y", "yes", "true" from the .ktr where previously only "yes" worked. Output of XML uses "yes" / "no" for backwards compatibility.


cseverino789:
@pentaho/r2d2

PR #6678, opened 2019-06-05
[BACKLOG-22179] Decouple blockuntilstepsfinish step from Kettle engine

cseverino789:
Moved Step to the core plugin.


cseverino789:
@pentaho/r2d2

PR #6682, opened 2019-06-06
[BACKLOG-22197] Decouple delay step from Kettle engine

cseverino789:


cseverino789:
See https://github.com/pentaho/pentaho-kettle/pull/6684

PR #6684, opened 2019-06-06
[BACKLOG-22197] Decouple delay step from Kettle engine

cseverino789:
Moved Delay Step to core plugin


cseverino789:
@pentaho/r2d2

PR #6690, opened 2019-06-10
[BACKLOG-22191] Decouple cubeinput step from Kettle engine

cseverino789:
Moved step from engine to plugins/core


cseverino789:
@pentaho/r2d2

PR #6706, opened 2019-06-12
[BACKLOG-22191] Decouple cubeinput step from Kettle engine

cseverino789:
Fixed *.properties files that had incorrect byte-encoding.


cseverino789:
@pentaho/r2d2

PR #6712, opened 2019-06-14
[BACKLOG-17630] Decouple addsequence step from Kettle engine

cseverino789:
Moved step from engine to plugins/core


cseverino789:
@pentaho/r2d2

NJtwentyone:
@cseverino789  where did you get the SEQ.png/svg?

cseverino789:
> @cseverino789 where did you get the SEQ.png/svg?

It's copied from https://github.com/pentaho/pentaho-kettle/blob/master/ui/src/main/resources/ui/images/SEQ.png 

Because multiple steps use the same resource that still exist in kettle/ui, I had to copy it rather than move the file.

PR #6723, opened 2019-06-19
[BACKLOG-29401] uses OS file separator for saving KTR to filesystem

dmoore62:


ccaspanello:
Approved from my side; let's get @apreto or @DFieldFL to do a final review.

PR #6737, opened 2019-06-28
[BACKLOG-27619] Avoid having duplicate API calls when opening artifac…

ddiroma:
…ts using Spoon if versioning and Comment Tracking options are enabled on Server


ddiroma:
@bmorrise

PR #6752, opened 2019-07-08
[PDI-17005] Issues with Carte Jobs API executeJob endpoint in Pentaho…

LeonardoCoelho71950:
… REST API

@bantonio82 identified an error while trying to connect into a repository, via PDI.
That was because I added jaxws-rt in kettle-engine pom. But in compile scope, not in provided scope. Thus, we didn't have jaxws-rt on runtime.   
Nevertheless, I created a KettleClientTransportException. This way, we don't need to have jaxws-rt in kettle-engine runtime.  

@smmribeiro  Can you review this please, since you helped me on that? :)
@pentaho/tatooine @pentaho-lmartins @bantonio82

### DIFF
--- a/engine/src/main/java/org/pentaho/di/repository/KettleRepositoryNotFoundException.java
+++ b/engine/src/main/java/org/pentaho/di/repository/KettleRepositoryNotFoundException.java
@@ -1,0 +1,59 @@
+/*! ******************************************************************************
+ *
+ *
+ * Pentaho Data Integration
+ * Copyright (C) 2019 by Hitachi Vantara : http://www.pentaho.com
+ *
+ *******************************************************************************
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ******************************************************************************/
+package org.pentaho.di.repository;
+
+import org.pentaho.di.core.exception.KettleException;
+
+/**
+ * This exception is thrown when repository wasn't found.
+ */
+public class KettleRepositoryNotFoundException extends KettleException {
+
+  private static final long serialVersionUID = 5920594838763107082L;
+
+  public KettleRepositoryNotFoundException() {
+    super();
+  }
+
+  /**
+   * @param message
+   * @param cause
+   */
+  public KettleRepositoryNotFoundException( String message, Throwable cause ) {
+    super( message, cause );
+  }
+
+  /**
+   * @param message
+   */
+  public KettleRepositoryNotFoundException( String message ) {
+    super( message );
+  }
+
+  /**
+   * @param cause
+   */
+  public KettleRepositoryNotFoundException( Throwable cause ) {
+    super( cause );
+  }
+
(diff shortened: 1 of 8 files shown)


LeonardoCoelho71950:
@graimundo I made some changes here.
Now we only use ClientTransportException at UnifiedRepositoryInvocationHandler.java.
Over there, if we notice a ClientTransportException, we will convert it into KettleAuthenticationException.

I couldn't simple used javax.xml.ws.WebServiceException (as I was thinking this morning), because we wanted to differentiate between an authentication error, due to bad credentials, and other errors like trying to connect to a BA server.

engine/src/main/java/org/pentaho/di/repository/KettleRepositoryNotFoundException.java
```diff
@@ -0,0 +1,59 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ ******************************************************************************/
+package org.pentaho.di.repository;
+
+import org.pentaho.di.core.exception.KettleException;
+
+/**
+ * This exception is thrown when there was an authentication error, while connecting into a repository.
```
graimundo: Description does not match exception name. Copy paste error?

engine/src/main/java/org/pentaho/di/www/ExecuteJobServlet.java
```diff
@@ -202,26 +204,29 @@ public void doGet( HttpServletRequest request, HttpServletResponse response ) th
...
-        // Something unexpected occurred.
-        response.setStatus( HttpServletResponse.SC_INTERNAL_SERVER_ERROR );
-        message = BaseMessages.getString(
-          PKG, "ExecuteJobServlet.Error.UnexpectedError", Const.CR + Const.getStackTracker( ke ) );
-        out.println( new WebResult( WebResult.STRING_ERROR, message ) );
+      out.println( new WebResult( WebResult.STRING_ERROR, message ) );
+      return;
+    } catch ( KettleException ke ) {
+      // Authentication Error.
+      if (  ke.getCause() instanceof ExecutionException ) {
```
graimundo: Are there tests for these cases?

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/UnifiedRepositoryInvocationHandler.java
```diff
@@ -58,6 +64,18 @@ private boolean lookupConnectException( Throwable root ) {
     return false;
   }
 
+  private boolean lookupAuthenticationException( Throwable root ) {
+    while ( root != null ) {
+      if ( root instanceof ClientTransportException ) {
```
graimundo: Where are the tests for this case?

engine/src/main/java/org/pentaho/di/repository/KettleRepositoryNotFoundException.java
LeonardoCoelho71950: Updated! *facepalm*

engine/src/main/java/org/pentaho/di/www/ExecuteJobServlet.java
LeonardoCoelho71950: There is a test to validate response error message, when a KettleAuthenticationException is thrown while connecting into a server.
https://github.com/pentaho/pentaho-kettle/blob/146e2160759081d40b451a35a11d953188a72e6f/engine/src/test/java/org/pentaho/di/www/ExecuteJobServletTest.java#L206

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/UnifiedRepositoryInvocationHandler.java
LeonardoCoelho71950: Done

PR #6767, opened 2019-07-15
[BACKLOG-29819] Adding ability to parameterize Spark Run Configuragtion

ccaspanello:


ccaspanello:
@pentaho/bb-8

DFieldFL:
Try this again @pentaho/bb-8

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/spark/SparkRunConfiguration.java
```diff
@@ -3,7 +3,7 @@
  *
  *  Pentaho Data Integration
  *
- *  Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ *  Copyright (C) 2002-2019 by Hitachi Vantara : http://www.pentaho.com
```
cravobranco: I think this class does not need to be part of the PR because only header was changed.

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/spark/SparkRunConfigurationProvider.java
```diff
@@ -3,7 +3,7 @@
  *
  *  Pentaho Data Integration
  *
- *  Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ *  Copyright (C) 2002-2019 by Hitachi Vantara : http://www.pentaho.com
```
cravobranco: I think this class does not need to be part of the PR because only header was changed.

PR #4955, opened 2018-02-22
[PDI-17407] Add correct crawling multiline message

Valeran86:
Replace zero or more spaces between `BaseMessages.getString(` 
and `PKG` on exactly one space when `Translator2` is crawling translation
keys from sources.

Key like this don't parsed and stayed invisible in Translator2:
miRowInsBef.setText( OsHelper.customizeMenuitemText( BaseMessages.getString(
  PKG, "TableView.menu.InsertBeforeRow" ) ) );

PR #6770, opened 2019-07-15
[PDI-17407] Translator2 can't crawling multiline `BaseMessages.getStr…

smmribeiro:
…ing(` \n ... "message"

The issue had a PR [(pentaho-kettle#4955)](https://github.com/pentaho/pentaho-kettle/pull/4955) contributed by the community (@Valeran86), however, while fixing the problem, it was only considering one possible scanPhrase (by hard-coding the scanPhrase "BaseMessages.getString( PKG,") - something one can't do because, by design, the translation system can work with several scanPhrases!

So I've changed the code in a more aggressive manner to calculate, for each scanPhrase configured, all the ways that it could be split over more than one source-code line. Now, each scanPhrase can have a new line both before and after any Java element and can be spread for more than two source-code line.

As the source change is big, I've added unit tests to guarantee the behaviour with one- to seven-line scenarios.

@pentaho/tatooine @pentaho-lmartins @ssamora

### DIFF
--- a/Translator.bat
+++ b/Translator.bat
@@ -22,6 +22,8 @@ REM limitations under the License.
 REM
 REM *****************************************************************************
 
+setlocal 
+
 set XMLFILE=%1%
 set SRCDIR=%2%
 
@@ -70,8 +72,7 @@ REM ** Set java runtime options                                     **
 REM ** Change 512m to higher values in case you run out of memory   **
 REM ******************************************************************
 
-set PENTAHO_DI_JAVA_OPTIONS="-Xmx512m" "-XX:MaxPermSize=256m"
-set OPT=
+set OPT="-Xmx512m" "-XX:MaxPermSize=256m"
 
 REM ***************
 REM ** Run...    **
--- a/Translator.bat
+++ b/Translator.bat
@@ -1,4 +1,4 @@
-@echo on
+@echo off
 
 REM *****************************************************************************
 REM
@@ -22,19 +22,11 @@ REM limitations under the License.
 REM
 REM *****************************************************************************
 
-REM **************************************************
-REM ** Libraries used by Kettle:                    **
-REM **************************************************
+set XMLFILE=%1%
+set SRCDIR=%2%
 
-set RUNDIR=%CD%\dist
-set XMLFILE=%CD%\%1%
-set SRCDIR=%CD%\%2%
-
-echo RUNDIR=%RUNDIR%
 echo XMLFILE=%XMLFILE%
 echo SRCDIR=%SRCDIR%
-
-cd dist
 
 set IS64BITJAVA=0
 
@@ -64,7 +56,6 @@ REM ===========================================
 REM Using 32bit Java, so include 32bit SWT Jar
 REM ===========================================
 set LIBSPATH=libswt\win32
-
 GOTO :CONTINUE
 :USE64
 REM ===========================================
@@ -72,30 +63,15 @@ REM Using 64bit java, so include 64bit SWT Jar
 REM ===========================================
 set LIBSPATH=libswt\win64
 set SWTJAR=..\libswt\win64
-
 :CONTINUE
-
-
-REM **********************
-REM   Collect arguments
-REM **********************
-
-set _cmdline=
-:TopArg
-if %1!==! goto EndArg
-set _cmdline=%_cmdline% %1
-shift
-goto TopArg
-:EndArg
 
 REM ******************************************************************
 REM ** Set java runtime options                                     **
 REM ** Change 512m to higher values in case you run out of memory   **
-REM ** or set the PENTAHO_DI_JAVA_OPTIONS environment variable      **
 REM ******************************************************************
 
 set PENTAHO_DI_JAVA_OPTIONS="-Xmx512m" "-XX:MaxPermSize=256m"
-set OPT="-Djava.library.path=%LIBSPATH%" "-DKETTLE_HOME=%KETTLE_HOME%" "-DKETTLE_REPOSITORY=%KETTLE_REPOSITORY%" "-DKETTLE_USER=%KETTLE_USER%" "-DKETTLE_PASSWORD=%KETTLE_PASSWORD%" "-DKETTLE_PLUGIN_PACKAGES=%KETTLE_PLUGIN_PACKAGES%" "-DKETTLE_LOG_SIZE_LIMIT=%KETTLE_LOG_SIZE_LIMIT%" "-DKETTLE_JNDI_ROOT=%KETTLE_JNDI_ROOT%"
+set OPT=
 
 REM ***************
 REM ** Run...    **
@@ -105,7 +81,5 @@ REM Eventually call java instead of javaw and do not run in a separate window
 set TRANSLATOR_START_OPTION=start "Translator"
 
 @echo on
-%TRANSLATOR_START_OPTION% java %OPT% -jar launcher\launcher-1.0.0.jar -lib ..\%LIBSPATH% -main org.pentaho.di.ui.i18n.editor.Translator2 %XMLFILE% %SRCDIR%
+%TRANSLATOR_START_OPTION% java %OPT% -jar launcher\launcher.jar -lib ..\%LIBSPATH% -main org.pentaho.di.ui.i18n.editor.Translator2 %XMLFILE% %SRCDIR%
 @echo off
-
-cd %RUNDIR%\..
--- a/ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
+++ b/ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2019 by Hitachi Vantara : http://www.hitachivantara.com
  *
  *******************************************************************************
  *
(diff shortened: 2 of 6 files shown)


ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
```diff
@@ -88,115 +149,77 @@
...
-  private Pattern packagePattern;
-  private Pattern importPattern;
-  private Pattern importMessagesPattern;
-  private Pattern stringPkgPattern;
-  private Pattern classPkgPattern;
+  private Pattern packagePattern = Pattern.compile( PACKAGE_PATTERN );
+  private Pattern importPattern = Pattern.compile( IMPORT_PATTERN );
+  private Pattern importMessagesPattern = Pattern.compile( IMPORT_MESSAGES_PATTERN );
+  private Pattern stringPkgPattern = Pattern.compile( STRING_PKG_PATTERN );
+  private Pattern classPkgPattern = Pattern.compile( CLASS_PKG_PATTERN );
```
pentaho-lmartins: If the constants above are only used here, why declare private constants?

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
```diff
@@ -88,115 +149,77 @@
...
-    importPattern = Pattern.compile( "^\\s*import [a-z\\._0-9]*\\.[A-Z].*;[ \t]*$" );
-    importMessagesPattern = Pattern.compile( "^\\s*import [a-z\\._0-9]*\\.Messages;[ \t]*$" );
-    stringPkgPattern = Pattern.compile( "^.*private static String PKG.*=.*$" );
-    classPkgPattern = Pattern.compile( "^.*private static Class.*\\sPKG\\s*=.*$" );
-  }
-
-  /**
-   * @return The source directories to crawl through
-   */
-  public List<String> getSourceDirectories() {
```
pentaho-lmartins: Don't remove public methods from a public class that someone may have overridden. Remember, this is open source code. If these are not necessary anymore please deprecate them.

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
```diff
@@ -208,38 +231,28 @@ public boolean traverseDescendents( FileSelectInfo info ) throws Exception {
...
-    // Also search for keys in the XUL files...
-    //
     for ( SourceCrawlerXMLFolder xmlFolder : xmlFolders ) {
-      String[] xmlDirs = { xmlFolder.getFolder(), };
-      String[] xmlMasks = { xmlFolder.getWildcard(), };
-      String[] xmlReq = { "N", };
-      boolean[] xmlSubdirs = { true, }; // search sub-folders too
+      String[] xmlDirs = { xmlFolder.getFolder() };
+      String[] xmlMasks = { xmlFolder.getWildcard() };
+      String[] xmlReq = { N_STRING };
```
pentaho-lmartins: Why are we declaring a constant for something that is only used once?

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
```diff
@@ -208,38 +231,28 @@ public boolean traverseDescendents( FileSelectInfo info ) throws Exception {
...
+  protected void crawlXmlFolders() throws Exception {
 
-    // Also search for keys in the XUL files...
-    //
     for ( SourceCrawlerXMLFolder xmlFolder : xmlFolders ) {
-      String[] xmlDirs = { xmlFolder.getFolder(), };
-      String[] xmlMasks = { xmlFolder.getWildcard(), };
-      String[] xmlReq = { "N", };
-      boolean[] xmlSubdirs = { true, }; // search sub-folders too
+      String[] xmlDirs = { xmlFolder.getFolder() };
```
pentaho-lmartins: Why so many variables that are only used once?

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
```diff
@@ -56,15 +39,93 @@
...
 
-  private String[] scanPhrases;
+  /**
+   * When searching for a scanPhrase, it may exist over more than one source line. This string contains the characters
+   * to be considered in the calculation of the different splits that each scanPhrase may have.
+   *
+   * @see #SPLIT_CHARACTERS
+   * @see #SPLITTER_AND_WHITESPACES_PATTERN
+   */
+  private static final String SPLIT_CHARACTERS_STRING = ".(,";
```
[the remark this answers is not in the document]
smmribeiro: See beneath.

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
```diff
@@ -56,15 +39,93 @@
...
+
+  /**
+   * When searching for a scanPhrase, it may exist over more than one source line. These are the characters to be
+   * considered in the calculation of the different splits that each scanPhrase may have.
+   *
+   * @see #SPLIT_CHARACTERS_STRING
+   * @see #SPLITTER_AND_WHITESPACES_PATTERN
+   */
+  private static final char[] SPLIT_CHARACTERS = SPLIT_CHARACTERS_STRING.toCharArray();
```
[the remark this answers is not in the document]
smmribeiro: The translator will process every line of every file it finds under the configured folder(s) and this can be a tight loop!
For every line it will see if it matches any of the scan phrases it was set to recognize and, for each of those, it has to recognize splitted phrases - and that's why we need to define which split characters (where a new line can appear) there are.
Before my changes, these (and others) strings were hard-coded and so, they were being created and destroyed every time with the associated costs on memory and GC... what I did here was create some single reusable static objects.
As we need the split characters in two distinct cases as String and as Character array, I had to differentiate them somehow (and thus the '_STRING'). The Character array object is declared based on the String object so that there's only one place to change if another split character is to be considered.

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
```diff
@@ -56,15 +39,93 @@
...
+  private static final String MESSAGES_PACKAGE_END_STRING = ".Messages;";
+  private static final String ORG_PENTAHO_STRING = "org.pentaho.";
+  private static final String SYSTEM_DOT_STRING = "System.";
+
+  //REGEX expressions to be compiled
+  private static final String PACKAGE_PATTERN = "^\\s*package\\s*.*;\\s*$";
+  private static final String IMPORT_PATTERN = "^\\s*import\\s*[a-z._0-9]*\\.([*]|([A-Z][a-z._0-9]*))\\s*;\\s*$";
+  private static final String IMPORT_MESSAGES_PATTERN = "^\\s*import\\s*[a-z._0-9]*\\.Messages;\\s*$";
+  private static final String STRING_PKG_PATTERN = "^.*private static String PKG.*=.*$";
+  private static final String CLASS_PKG_PATTERN = "^.*private static Class.*\\sPKG\\s*=.*$";
```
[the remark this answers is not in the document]
smmribeiro: As explained in my response above, these strings already existed but were hard-coded in the code itself; I've just extracted them to constants.

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
smmribeiro: It was only moved... it's now near line 740 :-)

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
smmribeiro: As above :-)

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
smmribeiro: This is how the code was, I only removed the extra comma that existed in every array definition (a valid but - IMHO - confusing java spec) and reformatted the lines.
But you're correct: at first I've inlined them all but when I saw that the code readability was somewhat reduced, I've reverted my change and left it as it was.

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
```diff
@@ -74,12 +135,12 @@
   /**
    * Source folder - package name - all the key occurrences in there
    */
-  private Map<String, Map<String, List<KeyOccurrence>>> sourcePackageOccurrences;
+  private Map<String, Map<String, List<KeyOccurrence>>> sourcePackageOccurrences = new HashMap<>();
```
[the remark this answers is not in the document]
smmribeiro: Looks like it, doesn't it? The use of this variable was the most difficult aspect I had in this issue!!!
Could I avoid this? Probably, but that would mean not a refactoring but a complete redoing of all this class... and, as you can see, this refactoring is already huge!

ui/src/main/java/org/pentaho/di/ui/i18n/MessagesSourceCrawler.java
smmribeiro: As you see, I've just inlined the assignment but you're right: as Pattern instances are immutable and already the regex expression used is a static final, these can also be static final!

Translator.bat
```diff
@@ -64,38 +56,22 @@ REM ===========================================
...
 REM ===========================================
 REM Using 64bit java, so include 64bit SWT Jar
 REM ===========================================
 set LIBSPATH=libswt\win64
 set SWTJAR=..\libswt\win64
-
 :CONTINUE
 
-
-REM **********************
```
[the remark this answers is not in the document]
smmribeiro: This batch file seems to be created by making some tweaks on some (very) old "spoon.bat" but without cleaning it very well. As so, there's some dead code, Spoon-only related code, outdated code...
This specific code is unnecessary as we don't need to collect arguments here - nothing was being done with them afterwards.

Translator.bat
```diff
@@ -64,38 +56,22 @@ REM ===========================================
...
-:EndArg
-
 REM ******************************************************************
 REM ** Set java runtime options                                     **
 REM ** Change 512m to higher values in case you run out of memory   **
-REM ** or set the PENTAHO_DI_JAVA_OPTIONS environment variable      **
 REM ******************************************************************
 
 set PENTAHO_DI_JAVA_OPTIONS="-Xmx512m" "-XX:MaxPermSize=256m"
-set OPT="-Djava.library.path=%LIBSPATH%" "-DKETTLE_HOME=%KETTLE_HOME%" "-DKETTLE_REPOSITORY=%KETTLE_REPOSITORY%" "-DKETTLE_USER=%KETTLE_USER%" "-DKETTLE_PASSWORD=%KETTLE_PASSWORD%" "-DKETTLE_PLUGIN_PACKAGES=%KETTLE_PLUGIN_PACKAGES%" "-DKETTLE_LOG_SIZE_LIMIT=%KETTLE_LOG_SIZE_LIMIT%" "-DKETTLE_JNDI_ROOT=%KETTLE_JNDI_ROOT%"
```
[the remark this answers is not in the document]
smmribeiro: All Spoon specific - it's not relevant for translator!

Translator.bat
```diff
@@ -64,38 +56,22 @@ REM ===========================================
...
-if %1!==! goto EndArg
-set _cmdline=%_cmdline% %1
-shift
-goto TopArg
-:EndArg
-
 REM ******************************************************************
 REM ** Set java runtime options                                     **
 REM ** Change 512m to higher values in case you run out of memory   **
-REM ** or set the PENTAHO_DI_JAVA_OPTIONS environment variable      **
```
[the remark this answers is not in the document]
smmribeiro: Dead code - it's not referred again!

PR #6774, opened 2019-07-16
[PDI-18044] - JobEntryHTTP uses java.net.Authenticator.setDefault whi…

bcostahitachivantara:
…ch is not threadsafe

@pentaho-lmartins 
@pentaho/tatooine 

Please review


pentaho-lmartins:
Fantastic fix and awesome tests!! Nicely done!

PR #6780, opened 2019-07-18
[PDI-18153] Text File Input step - Error Handling issue. Failure to write to file due to java.io.UnsupportedEncodingException

gryphendowre:
@pentaho/tatooine @pentaho-lmartins @smmribeiro 

* [PDI-18153] Added in further null/empty string check to ensure that null/empty string encoding does not cause an encoding issue.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/step/errorhandling/AbstractFileErrorHandler.java
+++ b/engine/src/main/java/org/pentaho/di/trans/step/errorhandling/AbstractFileErrorHandler.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2019 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/pentaho/di/trans/step/errorhandling/AbstractFileErrorHandler.java
```diff
@@ -40,6 +40,8 @@
 import org.pentaho.di.i18n.BaseMessages;
 import org.pentaho.di.trans.step.BaseStep;
 
+import static org.apache.commons.lang.StringUtils.isBlank;
```
pentaho-lmartins: We don't import static methods other than in unit tests.

gryphendowre: Is that standard though? @kurtwalker fixed some of my code adding a similar method to ensure that null and empty strings are covered.  I applied the same logic here.  How else would you like me to do this besides manually writing: "if encoding != "" && encoding != null"

Reference: https://github.com/pentaho/pentaho-vertica-bulkloader/pull/76/commits/d0786be799a2fb4bed5126ea4fd37e99707bf337

kurtwalker: There was a time that checkstyle didn't allow it.  I objected and got the dev team at the time to agree.  I use static imports frequently in production code, as does at least all the devs on ackbar.

PR #6789, opened 2019-07-23
[BACKLOG-22188] Decouple constant step from Kettle engine

cseverino789:
Moved from engine to plugins/core.

- Note: This step is used by pdi-dataservice-plugin which is a karaf plugin. This requires additional changes to karaf assembly.


cseverino789:
@pentaho/r2d2

PR #6792, opened 2019-07-29
[BACKLOG-24066] Add Metadata Injection (MDI) support to the String op…

NJtwentyone:
…erations step

 - update access level to static methods


NJtwentyone:
please review @pentaho/r2d2

PR #6794, opened 2019-07-30
[BACKLOG-29981] - snowflake needs the catalog specified when collecti…

kurtwalker:
…ng metadata

It was found Teradata does not support catalogs.  There are possibly others.  So, moving methods into the DatabaseInterface so that each implementation can override as needed

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/database/Database.java
+++ b/core/src/main/java/org/pentaho/di/core/database/Database.java
@@ -1951,7 +1951,7 @@ public class Database implements VariableSpace, LoggingObjectInterface {
   /**
    * Retrieves the table description matching the schema and table name.
    *
-   * @param shema the schema name pattern
+   * @param schema the schema name pattern
    * @param table the table name pattern
    * @return table description row set
    * @throws KettleDatabaseException if DatabaseMetaData is null or some database error occurs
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/pentaho/di/core/database/Database.java
```diff
@@ -1962,7 +1962,8 @@ private ResultSet getTableMetaData( String schema, String table ) throws KettleD
       throw new KettleDatabaseException( BaseMessages.getString( PKG, "Database.Error.UnableToGetDbMeta" ) );
     }
     try {
-      tables = getDatabaseMetaData().getTables( null, schema, table, TABLE_TYPES_TO_GET );
+      tables = databaseMeta.getDatabaseInterface().getTables(
```
mkambol: `databaseMeta.getDatabaseInterface().getTables(getDatabaseMetaData(), databaseMeta, ...` is quite a mouthful.

I wonder- would it make sense to add a .getTables method to DatabaseMeta that is a pass through to the databaseinterface?  That way the call could be simplified to something like:

 `databaseMeta.getTables(getDatabaseMetaData(),schema, table, TABLE_TYPES_TO_GET );`

Even that's a bit confusing with "databaseMeta" and "getDatabaseMetaData", but at least visually scans a bit better.

mkambol: Maybe even bring a .getTables() up in to Database, delegating to the dbinterface.  Pretty easy to grok `Database.getTables(schema, table, types)`

PR #6820, opened 2019-08-07
[PDI-18211] JSON Input Step Performance Degradation in 8.3

ppatricio:
The PPP-4374 case ( https://github.com/pentaho/pentaho-kettle/pull/6641 ) popped up some invalid getObject calls. For instance, it looks likes kettleAnalyzerUtil.java (pentaho-metaverse) calls KettleVFS and then saves the result as resource even if **it is not a valid file path**, so we should be validating if we really want to request a "getFileObject" from a whole json string as argument, and not just validating if it is empty:
https://github.com/pentaho/pentaho-metaverse/blob/master/api/src/main/java/org/pentaho/metaverse/api/analyzer/kettle/KettleAnalyzerUtil.java#L156

At the same time KettleVFS should also be prepared to deal with this. And we have more cases where this also happen. This PR added a scheme validation (getScheme function) that now throws an exception with invalid paths instead of a valid fileObject (that in fact does not exists...)

A refactoring was also made and a few more tests were added/adapted to the new behavior.

@mkambol @pentaho-lmartins

### DIFF
--- a/core/src/test/java/org/pentaho/di/core/vfs/KettleVFSTest.java
+++ b/core/src/test/java/org/pentaho/di/core/vfs/KettleVFSTest.java
@@ -22,17 +22,13 @@
 
 package org.pentaho.di.core.vfs;
 
-
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.assertFalse;
-
-
 import org.junit.Test;
 import org.pentaho.di.core.exception.KettleFileException;
 
-
-import java.io.IOException;
-
+import java.io.File;
 
 public class KettleVFSTest {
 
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/pentaho/di/core/vfs/KettleVFSTest.java
```diff
@@ -52,23 +48,86 @@ public void testStartsWithScheme() {
...
 
-    boolean test = KettleVFS.checkForScheme(schemes, true, vfsFilename, null, null);
-    assertFalse( test );
+    boolean test = "hdfs".equals( KettleVFS.getScheme( vfsFilename ) );
+    assertTrue( test );
+  }
 
+  @Test( expected = KettleFileException.class )
+  public void testCheckForSchemeFailWithException() throws KettleFileException {
+    String invalidVfsFilename = "{\"sport_hitting_tm\":{\"copyRight\":\" Copyright 2019 MLB Advanced Media, L.P.  Use of any"
```
cdeptula: Since this JSON string is copyrighted, should we use it in a public test case in our code?  It might be better to avoid copyright issues and mock up a JSON string instead...

ppatricio: Thank you Chris, i totally agree with you, it went by the test door.

PR #6826, opened 2019-08-12
[BACKLOG-29956] Added cache checkbox and cached row meta so it does n…

ccaspanello:
…ot make a database call when getFields is called


ccaspanello:
Merge conflicts will create a new PR.

PR #6827, opened 2019-08-12
[BACKLOG-29956]  Added cache flag/checkbox so we do not access the da…

ccaspanello:
…tabase every time getFields is called by down streams steps (useful in AEL / Spark)

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/steps/tableinput/TableInputMeta.java
+++ b/engine/src/main/java/org/pentaho/di/trans/steps/tableinput/TableInputMeta.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2018 by Hitachi Vantara : http://www.pentaho.com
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/pentaho/di/trans/steps/tableinput/TableInputMeta.java
```diff
@@ -295,10 +313,35 @@ public String getXML() {
...
+    try {
+      if ( cachedRowMeta != null ) {
+        retval.append( "    " + cachedRowMeta.getMetaXML() );
+      }
+    } catch ( IOException e ) {
+      throw new RuntimeException( "Unexpected error stored cached row meta data.", e );
+    }
+  }
+
+  public void updateCachedRowMeta() {
```
baudekin: Does this need to by synchronized???

ccaspanello: No clue

ccaspanello:
@pentaho/bb-8

PR #6831, opened 2019-08-13
[PDI-18275] Data type issue when upgrading to PDI 8.1 from 6.1

ddiroma:


ddiroma:
@bmorrise

ddiroma:
@mdamour1976

PR #6865, opened 2019-09-03
[SP-5252] Backport of PDI-18205 - 'Copy Rows to Result' step is not p…

ssamora:
…assing rows to later transformation in job if transformations in between do not use 'Copy Rows to Result' (8.3 Suite)

@RPAraujo @smmribeiro


RPAraujo:
Checklist passed locally.

PR #6919, opened 2019-10-04
[PDI-18283] pdi.log: Excuting Jobs/Transformation using Slave Server results to Jobs/Transformation name as null along with duplicate entries.

gryphendowre:

* [PDI-18283] Ensuring that the name of the job/transformation is printed. Either pull from the filename string if it exists, or to pull from the object name and append the appropriate extension to replicate the filename.
* [PDI-18283] Adding test code
* [PDI-18283] Updating license year


gryphendowre:
Wrong Branch/Name. Moving to correct location

PR #6943, opened 2019-10-14
[CLEANUP] - supressing sonar violations for fields with PASSWORD in the

kurtwalker:
name, closing possible vulnerability with client SSL.

### DIFF
--- a/core/src/main/java/org/pentaho/di/cluster/SlaveConnectionManager.java
+++ b/core/src/main/java/org/pentaho/di/cluster/SlaveConnectionManager.java
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2019 by Hitachi Vantara : http://www.pentaho.com
  *
  *******************************************************************************
  *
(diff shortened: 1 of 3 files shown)


buildguy:
### :gear: Build finished in 11m 56s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -P !assemblies -pl core
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|4660|0|7|

|[Details](http://wingman2.pentaho.net/job/projects/job/pdi/job/PR-6943/1/testReport)|
|----|


core/src/main/java/org/pentaho/di/cluster/SlaveConnectionManager.java
```diff
@@ -128,6 +130,7 @@ private static X509TrustManager getDefaultTrustManager() {
     return new X509TrustManager() {
       @Override
       public void checkClientTrusted( X509Certificate[] certs, String param ) throws CertificateException {
+        throw new CertificateException( "Client Authentication not implemented" );
```
mkambol: I'm not familiar with how we're managing SSL here.  Does this method get called by anything that was relying on it being a noop?

kurtwalker: It should not be getting called from anywhere.  It's used during client SSL authentication.  The only usage is in initiating a connection to the server.  Having the exception in place stops us from unknowingly extending the usage of this to a client SSL situation where it wouldn't be secured.

PR #6946, opened 2019-10-15
[BACKLOG-31521] Allow UUIDs to be passed from Pan/Kitchen, add them to the log4j MDC on execution

pentaho-whartman:
@pentaho/rogueone 

- Adds a new `uuid` parameter (this is hidden, like the `base64zip` parameter, so it won't need external documentation)
- Adds the UUID from params into the log4j MDC so it can be inserted into log lines

### DIFF
--- a/engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
+++ b/engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
@@ -23,6 +23,7 @@
 package org.pentaho.di.kitchen;
 
 import com.google.common.annotations.VisibleForTesting;
+import org.apache.log4j.MDC;
 import org.pentaho.di.base.AbstractBaseCommandExecutor;
 import org.pentaho.di.base.CommandExecutorCodes;
 import org.pentaho.di.base.KettleConstants;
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
```diff
@@ -222,6 +223,7 @@ public Result execute( Params params ) throws Throwable {
         return exitWithStatus( CommandExecutorCodes.Kitchen.COULD_NOT_LOAD_JOB.getCode() ); // same as the other list options
       }
 
+      MDC.put( KettleConstants.UUID, params.getUuid() );
```
befc: These MDC commands here seem a bit out of context.
Not a blocker, but would be nice to have a comment to provide context

engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
```diff
@@ -182,6 +183,8 @@ public Result execute( final Params params ) throws Throwable {
 
       final List<RowMetaAndData> rows = new ArrayList<RowMetaAndData>(  );
 
+      MDC.put( KettleConstants.UUID, params.getUuid() );
```
befc: Same as the other MDC comment

engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java
pentaho-whartman: Sure, there's a merge conflict on an import, I can add a comment when I fix that.

PR #6970, opened 2019-10-23
[BACKLOG-31521] Fix UUIDs for log4j in KJBs

pentaho-whartman:
@pentaho/rogueone

### DIFF
--- a/engine/src/main/java/org/pentaho/di/job/Job.java
+++ b/engine/src/main/java/org/pentaho/di/job/Job.java
@@ -44,6 +44,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.commons.vfs2.FileName;
 import org.apache.commons.vfs2.FileObject;
+import org.apache.log4j.MDC;
+import org.pentaho.di.base.KettleConstants;
 import org.pentaho.di.cluster.SlaveServer;
 import org.pentaho.di.core.Const;
 import org.pentaho.di.core.util.Utils;
@@ -204,6 +206,8 @@ public class Job extends Thread implements VariableSpace, NamedParams, HasLogCha
   private String executingUser;
 
   private String transactionId;
+
+  private String uuid;
 
   private Map<String, Object> extensionDataMap;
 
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/pentaho/di/job/Job.java
```diff
@@ -375,6 +379,11 @@ public void setRepository( Repository rep ) {
       copyParametersFrom( jobMeta );
       activateParameters();
 
+      // Add the UUID to log4j MDC so it can be inserted in log lines
+      if( uuid != null ) {
```
singletonc: First Glance Question: in which case would a uuid be null? is that ever expected, and if not should there be error handling?

buildguy:
### :gear: Build finished in 16m 46s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -pl engine
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|3577|0|7|

|[Details](http://wingman2.pentaho.net/job/projects/job/pdi/job/PR-6970/1/testReport)|
|----|


engine/src/main/java/org/pentaho/di/job/Job.java
befc: Good question. We do sanitize the input in such a way that UUID is either passed by us or generated in the **Params** class constructor. But since we don't know the full scope of the **Job** class usage, I'd say it makes sense to make this validation.

pentaho-whartman: Yeah, that was my logic. In the normal worker nodes flow we don't need to worry about it, but because this is a pretty "core" class that's used in lots of different ways, I wanted to keep the changes small/safe.

pdesai16: Are you sure we want to make the change in job.java and not in the CommandExecutor class. My only concern is that we are adding import to log4j in a class which is using the own logging classes

pdesai16: Are we introducing dependency to log4j?

pentaho-whartman: The log4j jars are already included for the project, we add the import of course, but no new jars needed to be added to the pom for this change.

We can't add the UUIDs in just the executor because the MDC is thread specific and the jobs run in their own thread -- this changes is adding the value so that the UUID is visible in the context where the job logs are written.

PR #7506, opened 2020-06-24
[BISERVER-14442] - Database connections using connection pooling will not adhere to changes in the connection details

bcostahitachivantara:
@pentaho-lmartins 
@ssamora 
@ppatricio 
@smmribeiro

### DIFF
--- a/core/src/main/java/org/pentaho/di/core/database/Database.java
+++ b/core/src/main/java/org/pentaho/di/core/database/Database.java
@@ -458,7 +458,8 @@ public class Database implements VariableSpace, LoggingObjectInterface, Closeabl
           String name = databaseMeta.getName();
           try {
             try {
-              this.connection = dsp.getNamedDataSource( name, DatasourceType.POOLED ).getConnection();
+              String datasourceName = ConnectionPoolUtil.getDataSourceName( databaseMeta, partitionId );
+              this.connection = dsp.getNamedDataSource( datasourceName, DatasourceType.POOLED ).getConnection();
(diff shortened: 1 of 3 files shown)


buildguy:
### :white_check_mark: Build finished in 22m
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl core
```
---
:ok_hand: All tests passed!
 
Tests run: **4841**, Failures: **0**, Skipped: **7** &nbsp;&nbsp; [Test Results](http://wingman.pentaho.net/job/projects/job/pdi/job/PR-7506/1/testReport)

 ---
 
 

core/src/main/java/org/pentaho/di/core/database/Database.java
```diff
@@ -458,7 +458,8 @@ public void normalConnect( String partitionId ) throws KettleDatabaseException {
           String name = databaseMeta.getName();
```
ssamora, line 458: Shouldn't we get rid of this name variable? and use the datasourceName insted? 
The name variable is now only being used log if something went wrong. But shouldn't the logged connection name be dataSourceName?
Also probably put String datasourceName where the String name = databaseMeta.getName() is.

bcostahitachivantara: I kept this variable because the name is a real name of datasource connection that was defined on managed datasource dialog. 
The datasourceName variable means a new datasource with variables replaced.

ssamora: Thanks. it makes sense that way.

PR #7678, opened 2020-09-14
[PDI-18881] Calculator Step : Absolute value ABS( A ) function in the…

ppatricio:
… step does not output the right result

@pentaho-lmartins @smmribeiro


smmribeiro (approved):
Looks good.

PR #7826, opened 2021-01-22
[PDI-16579] - Text File Input Not Recognized Escaped Characters At End of String

bcostahitachivantara:
@pentaho-lmartins 
@ssamora 
@smmribeiro


bcostahitachivantara:
Incorrect jira case

PR #7986, opened 2021-05-25
Webspoon 9.0

wseyler:
[TAP-4311] - Added Scripts that run on startup of pod to copy configuration information from NFS to correct location on webspoon pod.

[TAP-4311]: https://insightgroup.atlassian.net/browse/TAP-4311


wseyler:
Got some error I need to fix

PR #8132, opened 2021-10-19
[SP-6059] Backport of PDI-19286 - XML output : The '>' special charac…

ddiroma:
…ter is not escaped when 'Namespace' is set with this special character. (9.2 Suite)


ddiroma:
@pentaho/tatooine Please review/approve. Thanks!

PR #8173, opened 2022-01-14
[PDI-19378] - Pentaho 9.2 has much higher Voluntary Context Switches causing reduced performance

bcostahitachivantara:
@smmribeiro 
@andreramos89 
@renato-s


smmribeiro (approved):
Looks good (writeRowTo will set the variable).

PR #8223, opened 2022-02-25
[BACKLOG-30667] CSRF- /api/scheduler changes

sVarma1992:
[BACKLOG-30667] - CSRF- /api/scheduler changes

### DIFF
--- a/plugins/engine-configuration/assemblies/feature/src/main/feature/feature.xml
+++ b/plugins/engine-configuration/assemblies/feature/src/main/feature/feature.xml
@@ -26,6 +26,7 @@
 
   <feature name="${project.artifactId}" version="${project.version}">
     <details>${project.description}</details>
+    <bundle>mvn:com.hitachivantara.security.web/csrf-token-service-client-java-httpclient-v4/${csrf.token.service.client.version}</bundle>
   </feature>
 
   <feature name="pdi-engine-configuration-ui" version="${project.version}">
--- a/plugins/engine-configuration/assemblies/feature/src/main/feature/feature.xml
+++ b/plugins/engine-configuration/assemblies/feature/src/main/feature/feature.xml
@@ -26,7 +26,7 @@
 
   <feature name="${project.artifactId}" version="${project.version}">
     <details>${project.description}</details>
-    <bundle>mvn:com.hitachivantara.security.web/csrf-token-service-client-java-httpclient-v4/${csrf.token.service.client.version}</bundle>
+    <bundle>mvn:com.hitachivantara.security.web/csrf-token-service-client-java-httpclient-v4/${hv-security-web.version}</bundle>
   </feature>
 
   <feature name="pdi-engine-configuration-ui" version="${project.version}">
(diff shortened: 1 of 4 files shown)


plugins/engine-configuration/pom.xml
```diff
@@ -17,6 +17,7 @@
 
   <properties>
     <pdi.version>9.3.0.0-SNAPSHOT</pdi.version>
+    <csrf.token.service.client.version>0.8.0</csrf.token.service.client.version>
```
dcleao, line 20: There already exists a property in the [pentaho parent pom](https://github.com/pentaho/maven-parent-poms/blob/master/pom.xml#L353) for this that we _must_ use — `hv-security-web.version`.
So this line should be removed.

plugins/engine-configuration/impl/pom.xml
```diff
@@ -77,6 +77,12 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpcore</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.hitachivantara.security.web</groupId>
+      <artifactId>csrf-token-service-client-java-httpclient-v4</artifactId>
+      <version>${csrf.token.service.client.version}</version>
```
dcleao, line 83: Version must be inherited from the parent pom, so please remove the `version` element.

plugins/engine-configuration/assemblies/feature/src/main/feature/feature.xml
```diff
@@ -26,6 +26,7 @@
 
   <feature name="${project.artifactId}" version="${project.version}">
     <details>${project.description}</details>
+    <bundle>mvn:com.hitachivantara.security.web/csrf-token-service-client-java-httpclient-v4/${csrf.token.service.client.version}</bundle>
```
dcleao, line 29: Must use the property `hv-security-web.version` inherited from the parent pom, instead.

dcleao (changes requested):
Some POM issues

plugins/engine-configuration/pom.xml
sVarma1992: addressed the comments

plugins/engine-configuration/impl/pom.xml
sVarma1992: addressed the comments

plugins/engine-configuration/assemblies/feature/src/main/feature/feature.xml
sVarma1992: addressed the comments

plugins/engine-configuration/impl/pom.xml
dcleao: You should remove the version element altogether.
> so please remove the version element.

dcleao (changes requested):
Please see note.

plugins/engine-configuration/impl/pom.xml
sVarma1992: Removed the version as requested

dcleao (approved):

PR #8247, opened 2022-04-01
[BISERVER-14832] Memory leak by executing data transformation

smmribeiro:
@bcostahitachivantara @renato-s @andreramos89


bcostahitachivantara (approved):
Looks good

PR #8312, opened 2022-05-18
[SP-6118] Backport of PDI-19426 Calculator step: No error occurs even…

ddiroma:
… if a negative value is specified for SQRT(A) (9.2.0.4)


smmribeiro (approved):

PR #8338, opened 2022-05-26
[SP-6110][BAD-1920] Optional automatic updates of Hadoop *-site.xml i…

tkafalas:
…n cluster definition


smmribeiro (approved):

PR #8365, opened 2022-06-09
[BACKLOG-35902] Move Monet DB bulk loader from core to plugin

spravasahu:


plugins/monet-db-bulk-loader/impl/src/main/java/org/pentaho/di/trans/steps/monetdbbulkloader/MonetDBBulkLoader.java
```diff
@@ -38,20 +37,15 @@
...
 import org.pentaho.di.i18n.BaseMessages;
 import org.pentaho.di.trans.Trans;
 import org.pentaho.di.trans.TransMeta;
-import org.pentaho.di.trans.step.BaseStep;
-import org.pentaho.di.trans.step.StepDataInterface;
-import org.pentaho.di.trans.step.StepInterface;
-import org.pentaho.di.trans.step.StepMeta;
-import org.pentaho.di.trans.step.StepMetaInterface;
-import org.pentaho.di.trans.steps.monetdbagilemart.MonetDBRowLimitException;
+import org.pentaho.di.trans.step.*;
```
peterrinehart, line 44: This needs to be expanded to include only the classes being used from this package.

peterrinehart (changes requested):

plugins/monet-db-bulk-loader/impl/src/main/java/org/pentaho/di/trans/steps/monetdbbulkloader/MonetDBBulkLoader.java
spravasahu: Addressed review comments

peterrinehart (approved):

PR #8398, opened 2022-07-06
Backlog 35839 Move RSS plugins  out of core

GopiKrishnaAngadi:


plugins/pom.xml
```diff
@@ -84,6 +84,7 @@
         <module>salesforce</module>
         <module>pur</module>
         <module>ms-access</module>
+        <module>rss</module>
```
peterrinehart, line 87: If this plugin is getting moved from kettle-engine to a plugin, this should go in the lowdeps section above, not in highdeps.

peterrinehart (changes requested):

PR #8426, opened 2022-07-21
[PDI-19562] 'Delete Files' step does not delete files from Windows mapped drive

andreramos89:


andreramos89:
@smmribeiro 
@bcostahitachivantara 
@renato-s

PR #8445, opened 2022-08-01
[BACKLOG-35829] Move meta-inject-plugin out of OSGi

spravasahu:
[BACKLOG-35829] Move meta-inject-plugin out of OSGi


peterrinehart (changes requested):
Everything here looks good, however we're missing a class similar to https://github.com/pentaho/pentaho-kettle/blob/no-osgi/plugins/excel/impl/src/main/java/org/pentaho/di/lifecycle/KettleExcelPluginLifecycleListener.java to register the metaverse analyzer code (MetaInjectAnalyzer.java).

assemblies/plugins/pom.xml
```diff
@@ -680,6 +681,18 @@
             </exclusion>
           </exclusions>
         </dependency>
+        <dependency>
```
peterrinehart, line 684: This dependency should be up in the main group of dependencies, not inside any of the profiles.

peterrinehart (changes requested):

assemblies/plugins/pom.xml
spravasahu: addressed review comments

peterrinehart (approved):

PR #8461, opened 2022-08-03
[BACKLOG-35812] - Removing excel and reporting unnecessary jars from …

sVarma1992:
Removing excel and reporting unnecessary jars from engine


engine/src/main/java/org/pentaho/di/shared/SharedObjects.java
```diff
@@ -34,7 +34,7 @@
 
 import org.apache.commons.vfs2.FileObject;
```
ddiroma, line 35: @sVarma1992 Please update the copyright

sVarma1992: Done the requested changes

ddiroma (approved):

PR #8500, opened 2022-08-22
[BACKLOG-36234] Fix for interacting with files without any extension

urena-kenny:


peterrinehart (changes requested):
Any chance we can consolidate some of the repeated code in there?  That same boolean check is done a couple times, for instance.

peterrinehart (approved):

PR #8517, opened 2022-08-26
[PDI-19160]-Set variables(job entry): When "Variable substitution" is N, the value of the externally set variable is replaced.

krishnaalluhitachi:
Code is added to respect Variable Substitution flag so that variables declared in the Set Variables step operated as below

Variable will be overridden if the same variable is available in the property file and Variable Substitution flag is true
Variable will not be overridden if the Variable Substitution flag is false
Variable will be added if the variable is not available in the property file and Variable Substitution flag is true/false.


buildguy:
### :white_check_mark: Build finished in 18m 50s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl engine
```
---
 :exclamation: No tests found!
 ---
 
 

PR #8553, opened 2022-09-08
[PDI-18807] : Using Create Folder step to create a folder in S3 - displays a file being created until manually refreshed

SreejaSridhar:
Previously FileUtils.copyContent() method is used which is unable to copy a folder and could copy only files. And also the usage of FileUtils class was deprecated. Now it is able to copy folders as well.


buildguy:
### :white_check_mark: Build finished in 18m 28s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl engine
```
---
 :exclamation: No tests found!
 ---
 
 

PR #8563, opened 2022-09-12
[BACKLOG-36316] - Test cases failing inside QAT build 238 (after PDI-19357)

renato-s:
Changing the initialization of the new parameter to the beginning of Job/Transformation execution to avoid unit tests issues.

@andreramos89 
@bcostahitachivantara 
@smmribeiro

### DIFF
--- a/engine/src/main/java/org/pentaho/di/job/Job.java
+++ b/engine/src/main/java/org/pentaho/di/job/Job.java
@@ -307,7 +307,6 @@ public class Job extends Thread implements VariableSpace, NamedParams, HasLogCha
     this.log = new LogChannel( this, parentLogging );
     this.logLevel = log.getLogLevel();
     this.log.setHooks( this );
-    this.logBufferStartLine = KettleLogStore.getAppender().getLastBufferLineNr();
 
     if ( this.containerObjectId == null ) {
       this.containerObjectId = log.getContainerObjectId();
@@ -319,7 +318,6 @@ public class Job extends Thread implements VariableSpace, NamedParams, HasLogCha
     this.log = new LogChannel( this );
     this.logLevel = log.getLogLevel();
     this.log.setHooks( this );
-    this.logBufferStartLine = KettleLogStore.getAppender().getLastBufferLineNr();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/pentaho/di/job/Job.java
```diff
@@ -460,6 +458,8 @@ private void emergencyWriteJobTracker( Result res ) {
    */
   private Result execute() throws KettleException {
     try {
+      this.logBufferStartLine = KettleLogStore.getAppender().getLastBufferLineNr();
```
bcostahitachivantara, line 461: Why not just call setInitialLogBufferStartLine()? In future, you could need to add extra code to that method

engine/src/main/java/org/pentaho/di/trans/steps/jobexecutor/JobExecutor.java
```diff
@@ -222,6 +222,7 @@ private void executeJob() throws KettleException {
 
     ExtensionPointHandler.callExtensionPoint( log, KettleExtensionPoint.JobStart.id, data.executorJob );
 
+    data.executorJob.setInitialLogBufferStartLine();
```
bcostahitachivantara, line 225: Why you need this here? The method execute on job is not call?

ui/src/main/java/org/pentaho/di/ui/spoon/trans/TransGraph.java
```diff
@@ -4277,6 +4277,8 @@ private synchronized void prepareTrans( final Thread parentThread, final String[
       @Override
       public void run() {
         try {
+
+          trans.setInitialLogBufferStartLine();
```
bcostahitachivantara, line 4281: Why you need this here? The method execute on trans is not call?

engine/src/main/java/org/pentaho/di/job/Job.java
renato-s: Changed as suggested

ui/src/main/java/org/pentaho/di/ui/spoon/trans/TransGraph.java
renato-s: Some transformations aren't started from the execute method

engine/src/main/java/org/pentaho/di/trans/steps/jobexecutor/JobExecutor.java
renato-s: Some jobs aren't started from the execute method

bcostahitachivantara (approved):

PR #8647, opened 2022-10-12
[BACKLOG-36464] - Skip KettleThin kdbs for bare PDI installation

hv-vamsi:


buildguy:
### :white_check_mark: Build finished in 7m 29s
Build command:
```console
mvn clean verify -B -e -Daudit -amd -pl plugins/pur/core
```
---
:ok_hand: All tests passed!
 
Tests run: **280**, Failures: **0**, Skipped: **0** &nbsp;&nbsp; [Test Results](https://wingman.orl.eng.hitachivantara.com/job/projects/job/pdi-11/job/PR-8647/1/testReport)

 ---
 
 

krishnaalluhitachi (approved):
KettleThin type repos are ignored when there is no Hadoop plugin(System folder will  not be created in this case) in Kettle.

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepository.java
```diff
@@ -1492,13 +1497,17 @@ private LinkedHashMap<RepositoryObjectType, List<RepositoryFile>> getFilesByType
...
       List<DatabaseMeta> dbMetas = new ArrayList<DatabaseMeta>();
 
       for ( RepositoryFile file : children ) {
         DataNode node;
         node = pur.getDataForRead( file.getId(), NodeRepositoryFileData.class ).getNode();
-
+        if ( !hasOsgiFolder && StringUtils.equals( node.getProperty( "TYPE" ).getString(), "KettleThin" ) ) {
+          log.logDetailed( "Unable to find database {" + file.getName() + "}" );
+          continue;
+        }
```
dcleao, line 1510: **Post Merge Comment**
Imo, this would better be done in the error handler, where the severity of the error would be evaluated, and a decision of logging and continuing or throwing would be made.

Also, I think that rather than hard-coding the `KettleThin` connection type, it would be better to recognize that, generally, a repository may store connections whose type the current PDI client/engine does not know of. This is expected, given the different flavors of PDI clients/engines which exist, and in an application following a plugin architecture.

In this case, the `KettleThin` connection type will not be used by any other feature installed in the slimmed-down PDI client/engine, and, if it were, that dependency would likely be manifested as a JAVA or OSGi dependency between plugins of the engine, and, when unsatisfied, would be caught at engine boot time. 

Execution of transformations/jobs which would require the specific connection type or connection instance to be defined (not sure this kind of dependency exists for this specific case) will not be possible in this PDI client/engine. That's an unavoidable condition for a pluggable application with documents. The only question is how and when these dependencies are detected and their unsatisfaction handled.

Concluding, the particular condition of a connection type being unknown — for any unknown connection type — should be detectable by either catching a specific exception class or inspecting the caught error. That specific condition would be logged and connection loading would be resumed.

hv-vamsi: Hi, Thanks for reviewing and sharing the inputs.
Actually in DatabaseMeta exception is thrown if the plugin type is not available
https://github.com/pentaho/pentaho-kettle/blob/master/core/src/main/java/org/pentaho/di/core/database/DatabaseMeta.java#L592

are you suggesting to make change at DatabaseMeta to just log and remove this line ? If yes, we can revert the code changes for this PR

PR #8753, opened 2022-11-28
[PDI-19128]- replaced hard coded string with variable

jsaraf:
[PDI-19128]- replaced hard coded string with variable

PR #8823, opened 2023-02-24
[SP-6286]-Backport of PDI-19128 - Job Executor step: The value of "fi…

jsaraf:
…eld name in result file" is not referenced from subsequent steps. (9.3 Suite)


smmribeiro (approved):
Consistent with the original PR ([pentaho-kettle#8753](https://github.com/pentaho/pentaho-kettle/pull/8753)).

PR #8843, opened 2023-03-24
[PDI-19754] - Job entry "Delete files" fails if the files that needs to be deleted do not exist

krishnaalluhitachi:
PR will fix the below issue
If files exist, the delete is performed. But if files do not exist, then execution fails.


smmribeiro:
Closing as it's no longer to be backported to this branch.

PR #8845, opened 2023-03-24
[SP-6275] [PDI-19754] - Job entry "Delete files" fails if the files that needs to be deleted do not exist

krishnaalluhitachi:
This PR fixes below issue
If files exist, the delete is performed. But if files do not exist, then execution fails.


smmribeiro:
Closing as it no longer is valid (outdated).

PR #8870, opened 2023-04-19
[BACKLOG-37391] Sample VfsMetastoreProvider

niehusa:


plugins/metastore-locator/core/src/main/java/org/pentaho/metastore/locator/impl/vfs/VfsMetastoreProvider.java
```diff
@@ -0,0 +1,68 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@ServiceProvider( id = "VfsMetastoreProvider", description = "Provides access to a VFS metastore", provides = MetastoreProvider.class )
+public class VfsMetastoreProvider implements MetastoreProvider, ServiceProviderInterface<MetastoreProvider> {
+  private static final Logger logger = LoggerFactory.getLogger( VfsMetastoreProvider.class );
+  private final MetastoreSupplier supplier;
+
+  public VfsMetastoreProvider() {
+    this( MetaStoreConst::openPentahoVfsMetaStore );
```
pdesai16, line 42: Just to confirm - This is a placeholder until we get the XmlVfsMetastore in place so  it can be instantiated here.

niehusa: Look at MetaStoreConst.openPentahoVfsMetaStore() method, I believe that's the place we need to instantiate XmlVfsMetastore, not here.

PR #8875, opened 2023-05-02
[PDI-19754] - Job entry "Delete files" fails if the files that needs to be deleted do not exist

jsaraf:
(9.3.0.4 Suite)


jsaraf:
Closed as changes were required.

PR #8876, opened 2023-05-02
[PDI-19754] - Job entry "Delete files" fails if the files that needs …

jsaraf:
…to be deleted do not exist (9.3.0.4 Suite)


smmribeiro:
Closing as it's no longer to be backported like this.

PR #8860, opened 2023-04-04
[PDI-19725] Avoid NPE in dispose

abryant-hv:

PR #8883, opened 2023-05-08
[SP-6317] backport of PDI-19725: Avoid NPE in dispose

abryant-hv:


smmribeiro (approved):
Consistent with the original code (#8860)

PR #8884, opened 2023-05-09
[BACKLOG-37379] Replaced the reference to local metastore with the

pdesai16:
metastore locator in ConnectionUtil
- Removed the setting of localMetastore as supplier in ProviderServiceService

### DIFF
--- a/plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/service/ProviderServiceService.java
+++ b/plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/service/ProviderServiceService.java
@@ -55,6 +55,17 @@ public class ProviderServiceService implements ProviderServiceInterface, Service
     RepositoryFileProvider repoProvider = new RepositoryFileProvider();
     RecentFileProvider recentProvider = new RecentFileProvider();
     VFSFileProvider vfsProvider = new VFSFileProvider();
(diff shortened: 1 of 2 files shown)


plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/service/ProviderServiceService.java
```diff
@@ -56,16 +56,6 @@ private static void initProviderService() {
...
-    ConnectionManager.getInstance().setMetastoreSupplier( () -> {
-
-      try {
-        return MetaStoreConst.openLocalPentahoMetaStore();
-      } catch ( Exception e ) {
-        // Error getting metastore
-        throw new RuntimeException( e );
-      }
-    } );
-
```
abryant-hv, line 68: Something like this may still be needed, maybe for pan or kitchen? I'm not actually sure. We probably need this in place using the MetastoreLocator, in the end.

abryant-hv (approved):

aburgos-hv (approved):

plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/service/ProviderServiceService.java
pdesai16: Agreed. Will do more testing to see if it need the supplier in this case

PR #8895, opened 2023-05-18
[BACKLOG-37509] - refactor metastore access to go through MetastoreLocator

abryant-hv:


core/src/main/java/org/pentaho/di/metastore/MetaStoreConst.java
```diff
@@ -53,6 +60,27 @@ public class MetaStoreConst {
...
+    private volatile MetastoreLocatorOsgi metastoreLocator;
+
+    public IMetaStore get() {
+      if ( metastoreLocator == null ) {
+        synchronized ( this ) {
+          try {
+            if ( metastoreLocator == null ) {
+              Collection<MetastoreLocatorOsgi> metastoreLocators =
+                  PluginServiceLoader.loadServices( MetastoreLocatorOsgi.class );
+              metastoreLocator = metastoreLocators.stream().findFirst().orElse( null );
```
pdesai16, line 73: When we pick the first one, how do we know which metastoreLocator it is going to return

abryant-hv: There should generally be only one.

pdesai16: nvm, I was thinking about MetastoreProvider

pdesai16: We should try and use the MetastoreLocator instead of MetastoreLocatorOsgi as MetastoreLocatorOsgi  is meant to be used when working with karaf

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -609,7 +609,7 @@ public class Spoon extends ApplicationWindow implements AddUndoPositionInterface
 
   private CommandLineOption[] commandLineOptions;
 
-  public DelegatingMetaStore metaStore;
+  /* package for testing */ Supplier<IMetaStore> metaStoreSupplier = MetaStoreConst.getDefaultMetastoreSupplier();
```
pdesai16, line 612: At this point, since the plugins are not initialized, would this method - getDefaultMetastoreSupplier return the MetastoreLocator

abryant-hv: if the supplier were called before plugins were initialized, yes. But all the uses of metastore that we've found only occur after plugins are loaded.

core/src/main/java/org/pentaho/di/metastore/MetaStoreConst.java
abryant-hv: Using MetastoreLocator would require kettle core to depend on the plugin... I don't think we want that.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
abryant-hv: oh nevermind, there's a separate API jar. That should work.

core/src/main/java/org/pentaho/di/metastore/MetaStoreConst.java
pdesai16: There is separate api jar - pentaho-metastore-locator-api

engine/src/main/java/org/pentaho/di/www/SlaveServerConfig.java
```diff
@@ -107,27 +107,15 @@ public SlaveServerConfig() {
...
-      //
-      try {
-        MemoryMetaStore memoryStore = new MemoryMetaStore();
-        memoryStore.setName( "Memory metastore" );
-        metaStore.addMetaStore( memoryStore );
-        metaStore.setActiveMetaStoreName( memoryStore.getName() );
-      } catch ( MetaStoreException e2 ) {
-        throw new RuntimeException( "Unable to add a default memory metastore to the delegating store", e );
+    metaStoreSupplier = () -> {
+      IMetaStore metastore = MetaStoreConst.getDefaultMetastoreSupplier().get();
```
pdesai16, line 111: This removes two exception handling block and error message from log. In the new code, how is the error handled? Did we test this scenario by comparing the behavior in existing code and in the updated code to make sure we are consistent

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepository.java
```diff
@@ -380,16 +379,6 @@ public RootRef getRootRef() {
   @Override public void disconnect() {
     connected = false;
     metaStore = null;
-    IMetaStore activeMetaStore = null;
```
pdesai16, line 383: If we are connected to repository metastore, then disconnecting from repository should trigger a call to go back to MetastoreLocator so it can find which metastore to connect

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -8945,36 +8921,6 @@ public Repository getRepository() {
   public void setRepository( Repository rep ) {
     this.rep = rep;
     this.repositoryName = rep != null ? rep.getName() : null;
-    try {
```
pdesai16, line 8948: This block remove some log messages and error dialog. How are we handling these cases in updated code

engine/src/main/java/org/pentaho/di/www/SlaveServerConfig.java
abryant-hv: Most of the logging will have to be in MetastoreLocatorImpl. 
We could put in a log message that we're using the memory metastore here. 

There are no more exceptions being thrown, though. Those were only coming from DelegatingMetastore, which isn't needed anymore.

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepository.java
abryant-hv: hrm. The next time the Supplier is called, it should return a non-repository metastore, since the repository isn't connected anymore. 
But, if anything is holding on to a metastore instance instead of the supplier, that wouldn't switch over like the DelegatingMetastore did. hrm. I'll dig into this.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
abryant-hv: This is just the same logic that already exists in MetastoreLocatorImpl.

pdesai16:
I noticed that line:344 in PurRepository is not changed
 metaStore = new PurRepositoryMetaStore( this );
 
Is this intentional?

abryant-hv:
> I noticed that line:344 in PurRepository is not changed metaStore = new PurRepositoryMetaStore( this );
> 
> Is this intentional?

Nope, good catch. The call to reset the ConnectionManager metastoreSupplier is bad, too.

engine/src/main/java/org/pentaho/di/www/SlaveServerConfig.java
abryant-hv: OK, so looking more closely, neither of the failure cases here that used to log can still happen, or at least, can't be detected here. 

If the local metastore is attempted and fails in the MetastoreLocator, this code can't tell. 

The second log message was only based on exceptions thrown by DelegatingMetaStore, and the methods never actually threw anything, even though they declared that they could, so it could never have failed before.

abryant-hv: Ah, we can still log when we use the memory metastore, it just won't have an exception. I'll add that.

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepository.java
abryant-hv: See updated PR

abryant-hv:
> I noticed that line:344 in PurRepository is not changed metaStore = new PurRepositoryMetaStore( this );
> 
> Is this intentional?

This should now be fixed.

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepository.java
```diff
@@ -343,11 +342,6 @@ public RootRef getRootRef() {
           log.logDetailed( BaseMessages.getString( PKG, "PurRepositoryMetastore.Create.Message" ) );
         }
         metaStore = new PurRepositoryMetaStore( this );
```
pdesai16, line 344: If PurRepository creates an instance of PurRepositoryMetastore, how would it connect to VFS metastore

abryant-hv: This metaStore doesn't go anywhere. Everywhere else still uses the metaStore from MetastoreLocatorImpl.

pdesai16: This metastore is used when connected to repository. Is this story handling the usecase of using VFS Metastore when connected to repository

abryant-hv: No, in this story repository still uses the repository metastore, just not hard-coded.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -8945,36 +8921,6 @@ public Repository getRepository() {
...
-
-      // Keep one metastore here...
-      //
-      if ( metaStore.getMetaStoreList().size() > 1 ) {
-        metaStore.getMetaStoreList().remove( 0 );
-        metaStore.setActiveMetaStoreName( metaStore.getMetaStoreList().get( 0 ).getName() );
-      }
-
-      if ( rep != null ) {
```
pdesai16, line 8957: I am so sorry for adding comments in bit and pieces but now I understand the code little more and I feel like the line 8957 -8958 should stay as it is not related to metastore
My understanding is as User is switching from  one repository to another, the RepositoryCapabilities should be set to current repository. With this change, our RepositoryCapabilities won't be set to the newly connected repository.

abryant-hv: Good catch.

pdesai16 (approved):

niehusa (approved):

PR #8910, opened 2023-06-01
[PPP-4828] - Component Rome is outdated - missing dependencies for Rome

andreramos89:
@renato-s @bcostahitachivantara @smmribeiro


smmribeiro (approved):
While not optimal, seems that it's what is needed here (at least for now).

PR #8911, opened 2023-06-02
[BACKLOG-37509] add an accessor for the current metastore

abryant-hv:
and a couple other build fixes for other repos.

### DIFF
--- a/core/src/main/java/org/pentaho/di/metastore/MetaStoreConst.java
+++ b/core/src/main/java/org/pentaho/di/metastore/MetaStoreConst.java
@@ -28,7 +28,6 @@ import java.util.function.Supplier;
 
 import org.pentaho.di.core.Const;
 import org.pentaho.di.core.exception.KettlePluginException;
-import org.pentaho.di.core.logging.LogChannel;
 import org.pentaho.di.core.service.PluginServiceLoader;
 import org.pentaho.di.core.util.Utils;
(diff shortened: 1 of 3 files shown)


rkimmel-hv (approved):

core/src/main/java/org/pentaho/di/metastore/MetaStoreConst.java
```diff
@@ -72,7 +75,7 @@ public MetastoreLocator get() {
               metastoreLocator = metastoreLocators.stream().findFirst().orElse( null );
             }
           } catch ( KettlePluginException e ) {
-            LogChannel.GENERAL.logError( "Error getting metastore locator", e );
+            logger.error( "Error getting metastore locator", e );
```
niehusa, line 78: Why we can't continue to use `LogChannel.GENERAL.logError` ?

abryant-hv: The pentaho-platform tests were failing because of something like "General LogChannel not initialized". This error caused the tests to crash. I think either we never hit this exception block in the pentaho-kettle build, *or* the LogChannel stuff is somehow initialized in those cases. 
In either case, this type of logging seems more common. I also tried some other things to get the LogChannel stuff to work, but couldn't get it working. MetaStoreConst being a static-only class added complication.

niehusa (approved):

PR #8922, opened 2023-06-07
[SP-6347] - Backport of PDI-19808 - Connections keeps increasing when…

renato-s:
… the transformation is alternately executed on the Server and through REST API when connection pooling is enabled. (9.3 Suite)

@andreramos89 
@bcostahitachivantara 
@smmribeiro


smmribeiro (approved):

PR #8925, opened 2023-06-08
BACKLOG-37701 Update pentaho server with code to use vfs metastore

rkimmel-hv:


plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepository.java
```diff
@@ -341,7 +340,7 @@ public RootRef getRootRef() {
         if ( log.isDetailed() ) {
           log.logDetailed( BaseMessages.getString( PKG, "PurRepositoryMetastore.Create.Message" ) );
         }
-        metaStore = new PurRepositoryMetaStore( this );
+        IMetaStore metaStore = new PurRepositoryMetaStore( this );
```
pdesai16, line 343: Is anyone using this metastore? Is not, can we remove it because it is confusing that we are creating an instance of PurRepositoryMetastore but we really want is use the one created by MetastoreConst.getDefaultMetastore()

niehusa (approved):

PR #8928, opened 2023-06-09
BACKLOG-37701 Refactor all Metastore usage to use MetastoreLocator in…

rkimmel-hv:
… PentahoServer


engine/src/main/java/org/pentaho/di/job/Job.java
```diff
@@ -697,7 +698,7 @@ private Result execute( final int nr, Result prev_result, final JobEntryCopy job
       ( (VariableSpace) cloneJei ).copyVariablesFrom( this );
       cloneJei.setRepository( rep );
       if ( rep != null ) {
-        cloneJei.setMetaStore( rep.getMetaStore() );
```
abryant-hv, line 700: why does this need to be in an if block that checks rep?

engine/src/main/java/org/pentaho/di/trans/steps/injector/InjectorMeta.java
```diff
@@ -260,7 +261,7 @@ public void check( List<CheckResultInterface> remarks, TransMeta transMeta, Step
   public void check( List<CheckResultInterface> remarks, TransMeta transMeta, StepMeta stepMeta,
     RowMetaInterface prev, String[] input, String[] output, RowMetaInterface info ) {
     check( remarks, transMeta, stepMeta, prev, input, output, info, transMeta, repository, repository != null
-      ? repository.getMetaStore() : null );
+      ? MetaStoreConst.getDefaultMetastore() : null );
```
abryant-hv, line 264: same, why check repository here?

plugins/metastore-locator/core/src/main/java/org/pentaho/metastore/locator/api/impl/MetastoreLocatorImpl.java
```diff
@@ -70,10 +70,11 @@ public IMetaStore getMetastore() {
 
   @Override
   public IMetaStore getMetastore( String providerKey ) {
-    IMetaStore metaStore = getExplicitMetastore( MetastoreLocator.REPOSITORY_PROVIDER_KEY );
+    //Look for VFS metastore.
+    IMetaStore metaStore = getExplicitMetastore( MetastoreLocator.VFS_PROVIDER_KEY );
+    System.err.println( "This is running the current code totally dude" );
```
abryant-hv, line 75: probably don't want to check this in. :)

abryant-hv (changes requested):

plugins/metastore-locator/core/src/main/java/org/pentaho/metastore/locator/api/impl/MetastoreLocatorImpl.java
rkimmel-hv: Yep oops haha

engine/src/main/java/org/pentaho/di/job/Job.java
rkimmel-hv: I wasn't really sure what this was doing so for all instances that were calling getRepositoryMetastore I converted it to get default unless it was directly responsible for setting up the repository metastore instance

engine/src/main/java/org/pentaho/di/trans/steps/injector/InjectorMeta.java
rkimmel-hv: Same as before I just replaced all calls to getRepositoryMetastore with getDefaultMetastore where I could without making extra changes

engine/src/main/java/org/pentaho/di/repository/Repository.java
```diff
@@ -759,7 +759,7 @@ public void saveDatabaseMetaJobEntryAttribute( ObjectId id_job, ObjectId id_jobe
    * @return the Metastore that is implemented in this Repository. Return null if this repository doesn't implement a
    *         Metastore.
    */
-  public IMetaStore getMetaStore();
+  public IMetaStore getRepositoryMetaStore();
```
pdesai16, line 762: I see lot of places we are changing the name from getMetastore() to getRepositoryMetastore(), not sure why?

rkimmel-hv: This change is to make the metastore flow clearer, the getRepositoryMetastore method should only be used when we are actively trying to get an instance of the Repository Metastore specifically. I have changed all other calls where it makes sense to get the default metastore.

pdesai16: Earlier there was no distinction between the two. Are we are calling it with two different names because they can be out of sync

rkimmel-hv: Yes exactly, pretty much the only place that should be calling getRepositoryMetastore should be places like export/import from repository specifically or the RepositoryProvider

abryant-hv (approved):
I recommend squashing these commits

pdesai16 (approved):

PR #9125, opened 2023-12-06
[PDI-19994] - UDJC Crashes Spoon if it contains uninitialized variable used for output

hv-shiva:
Updated the janino compiler version to resolve the Spoon crash error


smmribeiro:
Closing as 9.5.0.0 is not a maintenance branch.

PR #9226, opened 2024-03-15
[BACKLOG-39983] Feature List API should be handled

varuntangirala:


core/src/main/java/org/pentaho/di/core/database/BaseDatabaseMeta.java
```diff
@@ -174,6 +175,15 @@ public abstract class BaseDatabaseMeta implements Cloneable, DatabaseInterfaceEx
   public static final String AUTOINCREMENT_SQL_FOR_BATCH_ID = "AUTOINCREMENT_SQL_FOR_BATCH_ID";
   public static final String NAMED_CLUSTER_ID = "NAMED_CLUSTER_ID";
 
+  public static final List<String> ATTRIBUTE_KEYLIST = Arrays.asList( ATTRIBUTE_PORT_NUMBER,
```
hv-vamsi, line 178: Suggest to use Map/Enum.
If going ahead with map then map can be initialized. and values can be put from static block.

engine/src/main/java/org/pentaho/di/www/DatabaseFeatureListServlet.java
```diff
@@ -0,0 +1,161 @@
...
+          String value = ( String ) jsonObject.get( "code" );
+          validateAndAddAttribute( attribute, value, databaseMeta );
+        }
+      }
+
+      JSONArray parametersArray = ( JSONArray ) dbConnectionJson.get( "parameters" );
+      if( !Objects.isNull( parametersArray ) ) {
+        for ( Object object : parametersArray ) {
+          JSONObject jsonObject = ( JSONObject ) object;
+          String parameter = ( String ) jsonObject.get("parameter");
```
hv-vamsi, line 138: Suggest to use String.valueOf(...)  to avoid `ClassCastException` (runtime/unchecked)

https://stackoverflow.com/questions/16815279/difference-between-casting-to-string-and-string-valueof#:~:text=Casting%20means%20that%20the%20object,take%20other%20types%20as%20well.&text=Both%20generates%20same%20output%20in,object%20is%20Not%20a%20string.

engine/src/main/java/org/pentaho/di/www/DatabaseFeatureListServlet.java
```diff
@@ -0,0 +1,161 @@
+/*! ******************************************************************************
+ *
+ * Pentaho Data Integration
+ *
+ * Copyright (C) 2002-2024 by Hitachi Vantara : http://www.pentaho.com
```
hv-vamsi, line 5: New class just should have 2024

hv-vamsi (approved):

addagudi (approved):

PR #9328, opened 2024-05-20
[BACKLOG-40781]:Salesforce dialog modifications

addagudi:


hv-vamsi (approved):

plugins/salesforce/core/src/main/java/org/pentaho/di/trans/steps/salesforce/SalesforceStep.java
```diff
@@ -191,8 +192,7 @@ public boolean testConnection() {
         try {
           connection.close();
         } catch ( Exception e ) {
-          successConnection = false;
-          logError( e.getMessage() );
+          //Ignore
```
hv-vamsi, line 195: Better to add a logger please.

addagudi: This is old code that's just moved out of SWT dialog class.

sVarma1992 (approved):

PR #9419, opened 2024-06-13
[BACKLOG-40272] - Bowl Variables in UI

abryant-hv:
- added getADefaultVariableSpace to Bowl, to enable starting with Bowl-specific variables
- Changed ExecutionConfigurations to use the current used variables from all open files, rather than used variables from previous executions.
- Changed the "Set Environment Variables" UI to only keep (in Spoon.variables) values that were modified by the user.
- Changed "Set Environment Variables" and "Show Used Environment Variables" dialogs to include values from the Bowl
- As a result, the Run dialog, and at runtime, project variables are included as expected.


pdesai16 (approved):

pdesai16:
Are they changes needed to SpoonJobDelegate like SpoonTransformationDelegate?

abryant-hv:
> Are they changes needed to SpoonJobDelegate like SpoonTransformationDelegate?

Surprisingly, no. SpoonJobDelegate never had the behavior that had to be fixed.

tgf (approved):

PR #9422, opened 2024-06-17
[BACKLOG-40258] Add project option to pan and kitchen

niehusa:


abryant-hv (changes requested):
close... but I think we need to move a bit more into the plugin.

engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
```diff
@@ -318,4 +336,17 @@ public Result getResult() {
   public void setResult( Result result ) {
     this.result = result;
   }
+
+  public int validateBowlPath( String path ) throws KettleException {
```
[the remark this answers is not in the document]
niehusa: I thought we're trying to avoid the word `project` in pentaho-kettle, so it's named bowl path, instead of project path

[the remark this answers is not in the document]
abryant-hv: Right... ideally, only the plugin knows about the path at all.

engine/src/main/java/org/pentaho/di/base/Params.java
```diff
@@ -62,6 +62,16 @@ public class Params implements IParams {
...
 
+  public String getBowlPath() {
+    return bowlPath;
+  }
+
+  public void setBowlPath( String bowlPath ) {
+    this.bowlPath = bowlPath;
+  }
+
+  private String bowlPath;
```
abryant-hv, line 73: This is still project-specific. Bowls in general don't have a path.

engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
```diff
@@ -22,36 +22,37 @@
...
-import java.io.IOException;
-import java.io.Serializable;
-import java.text.SimpleDateFormat;
-import java.util.ArrayList;
-import java.util.Base64;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
-
 import org.apache.commons.lang.StringUtils;
```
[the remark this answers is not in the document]
abryant-hv: how is this resolved?

engine/src/main/java/org/pentaho/di/base/CommandExecutorCodes.java
```diff
@@ -31,17 +31,19 @@ public class CommandExecutorCodes {
...
-    CMD_LINE_PRINT ( 9, "Command line usage printing" );
+    SUCCESS( 0, "The transformation ran without a problem" ),
+    ERRORS_DURING_PROCESSING( 1, "Errors occurred during processing" ),
+    UNEXPECTED_ERROR( 2, "An unexpected error occurred during loading / running of the transformation" ),
+    UNABLE_TO_PREP_INIT_TRANS( 3, "Unable to prepare and initialize this transformation" ),
+    KETTLE_VERSION_PRINT( 6, "Kettle Version printing" ),
+    COULD_NOT_LOAD_TRANS( 7, "The transformation couldn't be loaded from XML or the Repository" ),
+    ERROR_LOADING_STEPS_PLUGINS( 8, "Error loading steps or plugins (error in loading one of the plugins mostly)" ),
+    CMD_LINE_PRINT( 9, "Command line usage printing" ),
+    ERRORS_INVALID_PROJECT( 10, "The project path is invalid, please provide a valid one" );
```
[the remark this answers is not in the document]
abryant-hv: kettle shouldn't know about project paths.

engine/src/main/java/org/pentaho/di/kitchen/Kitchen.java
```diff
@@ -221,6 +220,23 @@ public KettleException call() throws Exception {
...
+      exitJVM( 9 );
+    }
+
+    CommandLineOption.parseArguments( args, options, log );
+
+    if ( !Utils.isEmpty( optionRepname.toString() ) && !Utils.isEmpty( pluginParam.toString() ) ) {
+      log.logError( BaseMessages.getString( PKG, "Kitchen.Error.ComdLine.RepAndProject" ) );
+      blockAndThrow( kettleInitFuture );
+      exitJVM( 9 );
+    }
```
abryant-hv, line 239: If we can't do this in the plugin, we should just skip the check. 

It seems like repository+project is going to be required shortly anyway.

engine/src/main/java/org/pentaho/di/pan/CommandExecutorResult.java
```diff
@@ -0,0 +1,37 @@
+package org.pentaho.di.pan;
+
+public interface CommandExecutorResult {
+
+  public static enum Result {
+    SUCCESS( 0, "Success" ),
+    ERRORS_INVALID_PROJECT( 10, "The project path is invalid, please provide a valid one" );
```
abryant-hv, line 7: Should not have project or path mentioned in kettle.

engine/src/main/java/org/pentaho/di/pan/Pan.java
```diff
@@ -213,7 +219,10 @@ public static void main( String[] a ) throws Exception {
         log.logMinimal( BaseMessages.getString( PKG, "Pan.Log.Loglevel", log.getLogLevel().getDescription() ) );
       }
 
-      // ///////////////////////////////////////////////////////////////////////////////////////////////////
+      if ( !Utils.isEmpty( optionRepname.toString() ) && !Utils.isEmpty( pluginParam.toString() ) ) {
+        log.logError( BaseMessages.getString( PKG, "Pan.Error.ComdLine.RepAndProject" ) );
+        exitJVM( 9 );
+      }
```
abryant-hv, line 225: same as above

abryant-hv (changes requested):

engine/src/main/java/org/pentaho/di/base/AbstractBaseCommandExecutor.java
```diff
@@ -22,36 +22,37 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.Serializable;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
```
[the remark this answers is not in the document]
niehusa: No change from me.

engine/src/main/java/org/pentaho/di/pan/CommandExecutorResult.java
niehusa: This `Result` ENUM is part of the `CommandExecutorResult` interface.
Pan.java/Kitchen.java need the code/description from `Result` for exit.
That's why the ENUM is part of a interface in kettle.
Not sure how to work around it.
Your input/idea will be highly appreciated.

abryant-hv: Have the validation method return a string, or throw an exception, and log that instead? 

The validation message for plugins can't come from an enum in kettle...

engine/src/main/java/org/pentaho/di/base/Params.java
```diff
@@ -96,12 +96,18 @@ public static class Builder {
     private String base64Zip;
     private NamedParams namedParams;
     private NamedParams customNamedParams;
+    private String bowlPath;
+
+    public Builder bowlPath( String bowlPath ) {
+      this.bowlPath = bowlPath;
+      return this;
+    }
```
abryant-hv, line 104: This seems like it's unused? I don't see anywhere in the PR that uses this value.

engine/src/main/java/org/pentaho/di/pan/Pan.java
```diff
@@ -239,48 +244,86 @@ public static void main( String[] a ) throws Exception {
...
       exitJVM( CommandExecutorCodes.Pan.UNEXPECTED_ERROR.getCode() );
     }
   }
 
+  private static CommandExecutorResult validateAndSetPluginParam( LogChannelInterface log, String path ) {
+//    String returnMsg = "";
+    CommandExecutorResult result = null;
+      try {
+        Collection<CommandLineOptionProvider> commandlineProviders = PluginServiceLoader.loadServices( CommandLineOptionProvider.class );
+        CommandLineOptionProvider provider = commandlineProviders.stream().findFirst().orElse( null );
```
abryant-hv, line 302: why are we only using one? Shouldn't we use all the plugin-provided CommandLineOptions?

engine/src/main/java/org/pentaho/di/pan/Pan.java
```diff
@@ -239,48 +244,86 @@ public static void main( String[] a ) throws Exception {
...
+        result = provider.handleParameter( log, path );
+
+      } catch ( KettlePluginException ke ) {
```
abryant-hv, line 305: why are we ignoring this exception?

engine/src/main/java/org/pentaho/di/pan/Pan.java
```diff
@@ -239,48 +244,86 @@ public static void main( String[] a ) throws Exception {
...
+
+      exitJVM( rslt.getExitStatus() );
 
     } catch ( Throwable t ) {
       t.printStackTrace();
       exitJVM( CommandExecutorCodes.Pan.UNEXPECTED_ERROR.getCode() );
     }
   }
 
+  private static CommandExecutorResult validateAndSetPluginParam( LogChannelInterface log, String path ) {
```
abryant-hv, line 297: is "path" here specifically a path? If not, please change the variable name

engine/src/main/java/org/pentaho/di/pan/Pan.java
```diff
@@ -239,48 +244,86 @@ public static void main( String[] a ) throws Exception {
...
+
+  private static List<CommandLineOption> getAdditionalCommandlineOption( CommandLineOption[] options,
+                                                                         StringBuilder param ) {
+    List<CommandLineOption> modifiableList = new ArrayList<>();
+
+    Collections.addAll( modifiableList, options );
+    try {
+      for ( CommandLineOptionProvider provider : PluginServiceLoader.loadServices( CommandLineOptionProvider.class ) ) {
+        provider.prepareAdditionalCommandlineOption( modifiableList, param );
+      }
```
abryant-hv, line 320: it's still not clear to me how this can re-use the param multiple times. What is param doing?

engine/src/main/java/org/pentaho/di/pan/PanCommandExecutor.java
```diff
@@ -110,29 +111,38 @@ public Result execute( final Params params, String[] arguments  ) throws Throwab
           /**
            * if set, _trust_user_ needs to be considered. See pur-plugin's:
            *
-           * @link https://github.com/pentaho/pentaho-kettle/blob/8.0.0.0-R/plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepositoryConnector.java#L97-L101
-           * @link https://github.com/pentaho/pentaho-kettle/blob/8.0.0.0-R/plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/WebServiceManager.java#L130-L133
+           * @link https://github.com/pentaho/pentaho-kettle/blob/8.0.0
+           * .0-R/plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepositoryConnector.java#L97-L101
+           * @link https://github.com/pentaho/pentaho-kettle/blob/8.0.0
+           * .0-R/plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/WebServiceManager.java#L130-L133
```
abryant-hv, line 117: this looks like it broke these links

engine/src/main/resources/org/pentaho/di/pan/messages/messages_en_US.properties
```diff
@@ -69,4 +69,5 @@ Pan.ComdLine.JarFile=specifies the jar filename
 Pan.Log.LoadingTransXML=Loading transformation from XML file [{0}]
 Pan.Error.TransJVMExitCodeInvalid=The transformation variable {0} is not an integer.  It is set to "{1}".
 Pan.Log.JVMExitCode=Exiting the JVM with exit code {0}.
-Pan.ComdLine.Metrics=Gather metrics during execution
\ No newline at end of file
+Pan.ComdLine.Metrics=Gather metrics during execution
+Pan.Error.ComdLine.RepAndProject=ERROR\: Cannot specify repository and project path at the same time.
```
abryant-hv, line 73: I would still rip this out. We're going to add projects to repos, so this is moot.

abryant-hv (changes requested):

engine/src/main/java/org/pentaho/di/pan/Pan.java
niehusa: currently it's not handling multiple command line option as it's not a specified requirement in the story. 
Maybe we can add later?!

abryant-hv (approved):

pdesai16 (approved):

pdesai16:
Pls update the copyright header

niehusa:
> Pls update the copyright header

Updated! Please see if I missed anything.
Appreciated it.

engine/src/main/java/org/pentaho/di/pan/Pan.java
niehusa: currently it's not handling multiple command line option as it's not a specified requirement in the story. 
Maybe we can add later?!

abryant-hv: add a comment, then, so a future developer won't be confused. :)

PR #9436, opened 2024-06-24
[BACKLOG-40760] Add doAction method for getVariable dialog

kdarshana12:


engine/src/main/java/org/pentaho/di/trans/steps/getvariable/GetVariable.java
```diff
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2017 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2024 by Hitachi Vantara : http://www.pentaho.com
```
sVarma1992, line 5: change 2024 to 2002-2024

engine/src/main/java/org/pentaho/di/trans/steps/getvariable/GetVariable.java
```diff
@@ -38,9 +42,6 @@
 
 /**
  * Get information from the System or the supervising transformation.
- *
```
sVarma1992, line 41: why is this deleted we should not delete anything existing

engine/src/main/java/org/pentaho/di/trans/steps/getvariable/GetVariable.java
```diff
@@ -51,6 +52,7 @@ public GetVariable( StepMeta stepMeta, StepDataInterface stepDataInterface, int
     super( stepMeta, stepDataInterface, copyNr, transMeta, trans );
   }
 
+  @Override
```
sVarma1992, line 55: why is this added this is not our changes, same thing for the other two methods as well

engine/src/main/java/org/pentaho/di/trans/steps/getvariable/GetVariable.java
```diff
@@ -140,8 +143,40 @@ public boolean init( StepMetaInterface smi, StepDataInterface sdi ) {
     return false;
   }
 
+  @Override
   public void dispose( StepMetaInterface smi, StepDataInterface sdi ) {
     super.dispose( smi, sdi );
   }
 
+  public JSONObject doAction( String fieldName, StepMetaInterface stepMetaInterface, TransMeta transMeta,
```
sVarma1992, line 151: Map<String, String> queryParams 
The above param has been added to the doAction method
Please rebase with latest feature branch

engine/src/main/java/org/pentaho/di/trans/steps/getvariable/GetVariable.java
kdarshana12: Added because of check-style suggestion. Removed in latest push.

PR #9457, opened 2024-07-03
[PDI-20094] - fix VFS recent file for file-open-save

abryant-hv:
Can't assume LOCAL and REPOSITORY are the only options

### DIFF
--- a/plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/dialog/FileOpenSaveDialog.java
+++ b/plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/dialog/FileOpenSaveDialog.java
@@ -529,15 +529,8 @@ public class FileOpenSaveDialog extends Dialog implements FileDetails {
   private void setPreviousSelection() {
     String targetPath = this.fileDialogOperation.getPath();
     if ( StringUtils.isNotEmpty( targetPath ) ) {
-      // If the path is file, set the parent to be the target path
       if ( targetPath.startsWith( "file:///" ) ) {
         targetPath = targetPath.substring( 7 );
-      }
-      java.io.File filePath = new java.io.File( targetPath );
-      if ( !filePath.exists() && !ProviderFilterType.REPOSITORY.toString().equalsIgnoreCase( fileDialogOperation.getProvider() ) ) {
-        return;
-      } else if ( filePath.isFile() ) {
-        targetPath = filePath.getParent();
       }
     }
 
--- a/ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
+++ b/ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
@@ -8978,7 +8978,8 @@ public class Spoon extends ApplicationWindow implements AddUndoPositionInterface
     } else {
       this.repositoryName = null;
       lastFileOpened = props.getLastUsedLocalFile();
-      lastFileOpenedProvider = ProviderFilterType.LOCAL.toString();
+      // we don't know what the provider is.
+      lastFileOpenedProvider = null;
     }
       if ( rep != null ) {
         this.capabilities = rep.getRepositoryMeta().getRepositoryCapabilities();


bconlon-hv (approved):

plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/dialog/FileOpenSaveDialog.java
```diff
@@ -534,7 +534,7 @@ private void setPreviousSelection() {
         targetPath = targetPath.substring( 7 );
       }
       java.io.File filePath = new java.io.File( targetPath );
-      if ( !filePath.exists() && !ProviderFilterType.REPOSITORY.toString().equalsIgnoreCase( fileDialogOperation.getProvider() ) ) {
+      if ( !filePath.exists() && ProviderFilterType.LOCAL.toString().equalsIgnoreCase( fileDialogOperation.getProvider() ) ) {
```
pdesai16, line 537: Why was the ProviderFilterType set to Repository

abryant-hv: it used to say: "If the provider is anything other than REPOSITORY, and the (java.io.) File doesn't exist, return"

That broke because not the only non-REPOSITORY provider that actually works with java.io.File is LOCAL. Other providers, like VFS (this bug) would just never return true for "exists()". 

So the new check is: "if the provider is LOCAL and the File doesn't exist, return"

plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/dialog/FileOpenSaveDialog.java
```diff
@@ -534,7 +534,7 @@ private void setPreviousSelection() {
         targetPath = targetPath.substring( 7 );
       }
       java.io.File filePath = new java.io.File( targetPath );
```
NJtwentyone, line 536: I don't agree with the solution as it doesn't address the core of the problem using` java.io.File` which is for local files URI(`"C:\someDirectory"` or `"/someDirectory"` ) for PVFS URI's ie (`"pvfs://someConneciton/someFile"`)

**It may address** the original case by changing (ProviderFilterType.REPOSITORY to  ProviderFilterType.LOCAL), but a new developer should question why am I using java.io.File for PVFS URI. Then you'd have to either add a java comment and/or they would have to surmise all the edge cases and determine if the code is still valid.

If ever there are changes to this function in the future, the core problem will still be there. hopefully they don't need  to use PVFS URI ~ exists() or PVFS URI ~ getParent() function

if these changes are a means to unblock efforts for "project-profile" by all means go ahead. But I would not consider this issue addressed

running a simple unit test below, the first two asserts pass, and the last two fail

`java.io.File#exists` doesn't behave properly for PVFS URI
`java.io.File#getParent` doesn't behave properly for the PVFS URI

```
@Test
public void testJavaIOFIle() throws Exception {

  Assert.assertFalse( new java.io.File(  "pvfs://someConnection/folder-that-does-not-exist").exists() );
  Assert.assertFalse( new java.io.File(  "pvfs://someConnection/folder-that-does-not-exist").isFile() );
  Assert.assertEquals("pvfs://someConnection", new java.io.File(  "pvfs://someConnection/folder-that-does-not-exist").getParent() );
  Assert.assertTrue( new java.io.File(  "pvfs://someConnection/valid-folder").exists() );

}
```

NJtwentyone (commented):
Talked a bit offline, I don't think the solution addresses the spirit of jira issue's section "Technical Observations bullet 1.b" - https://hv-eng.atlassian.net/browse/PDI-20094

see comment - https://github.com/pentaho/pentaho-kettle/pull/9457/files#r1664303073

plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/dialog/FileOpenSaveDialog.java
abryant-hv: I've pushed a newer update that makes it so we never use java.io.File to parse any non-LOCAL URIs.

abryant-hv:
After more debugging and testing with various file sources, it appears that the original problem that added the java.io.File code has been solved in different ways since it was reported. The entire section of code is now redundant, and can be removed.

pdesai16 (approved):

PR #9470, opened 2024-07-12
[BACKLOG-40258][BACKLOG-41446] Fix test failure

niehusa:


bconlon-hv (changes requested):
Why are we removing error-handling code from kitchen itself?

niehusa:
> Why are we removing error-handling code from kitchen itself?

We don't need it at the end of Kitchen.main() method so remove it.

bconlon-hv (approved):

bconlon-hv:
Approved.  Was concerned that we were removing some old kitchen logic, but it's just something new that we've added only on our side branch.

pdesai16 (approved):

PR #9485, opened 2024-07-17
[BACKLOG-40796] Add new DB Connections with Project Profile Rules. If…

pdesai16:
… the project is selected, it adds the connection at the project level, otherwise it will add the connection at the global level.

- Added a new DatabaseManagementInterface that will be used when performing crud operation from UI.

### DIFF
--- a/core/src/main/java/org/pentaho/di/shared/VfsSharedObjectsIO.java
+++ b/core/src/main/java/org/pentaho/di/shared/VfsSharedObjectsIO.java
@@ -79,6 +79,8 @@ public class VfsSharedObjectsIO implements SharedObjectsIO {
    */
   public VfsSharedObjectsIO( String rootFolder ) {
     this.rootFolder = rootFolder;
+    // Get the complete path to shared.xml
+    this.sharedObjectFile = getSharedObjectFilePath( rootFolder );
   }
 
 
@@ -92,7 +94,7 @@ public class VfsSharedObjectsIO implements SharedObjectsIO {
   public synchronized Map<String, Node> getSharedObjects( String type ) throws KettleXMLException {
     if ( !isInitialized ) {
       // Load the shared.xml file
-      loadSharedObjectNodeMap( rootFolder );
+      loadSharedObjectNodeMap( sharedObjectFile );
       isInitialized = true;
     }
     return getNodesMapForType( type );
@@ -103,16 +105,14 @@ public class VfsSharedObjectsIO implements SharedObjectsIO {
    * where key can be {"connection", "slaveserver", "partitionschema" or  clusterschema"} and
    * value will be xml Node.
    *
-   * @param rootFolder The path to the shared object file
+   * @param pathToSharedObjectFile The path to the shared object file
    * @throws KettleXMLException
    */
-  private void loadSharedObjectNodeMap( String rootFolder ) throws KettleXMLException {
-    // Get the complete path to shared.xml
-    this.sharedObjectFile = getSharedObjectFilePath( rootFolder );
+  private void loadSharedObjectNodeMap( String pathToSharedObjectFile ) throws KettleXMLException {
 
     try {
       // Get the FileObject
-      FileObject file = KettleVFS.getFileObject( sharedObjectFile );
+      FileObject file = KettleVFS.getFileObject( pathToSharedObjectFile );
 
       // If we have a shared file, load the content, otherwise, just keep this one empty
       if ( file.exists() ) {
@@ -132,7 +132,7 @@ public class VfsSharedObjectsIO implements SharedObjectsIO {
       }
     } catch ( Exception e ) {
       throw new KettleXMLException( BaseMessages.getString( PKG, "SharedOjects.ReadingError",
-        sharedObjectFile ), e );
+        pathToSharedObjectFile ), e );
     }
 
   }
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/pentaho/di/shared/VfsSharedObjectsIO.java
```diff
@@ -186,18 +186,15 @@ private static String  getDefaultSharedObjectFileLocation() {
...
-      FileObject fileObject = KettleVFS.getFileObject( sharedObjectFile );
-      Optional<String> backupFileName = createOrGetFileBackup( fileObject );
-      writeToFile( fileObject, backupFileName );
-      isInitialized = false;
-    }
-    // TODO need to handle the case when the new shared object is being added
+    // Add or Update the map entry for this name
+    nodeMap.put( name, node );
+
+    FileObject fileObject = KettleVFS.getFileObject( sharedObjectFile );
```
abryant-hv, line 194: I would add a comment about this not needing a bowl

engine/src/main/java/org/pentaho/di/shared/DatabaseConnectionManager.java
```diff
@@ -47,7 +50,11 @@ private DatabaseConnectionManager( Bowl bowl ) throws KettleXMLException {
     this.sharedObjectIO = bowl.getSharedObjectsIO();
   }
 
-
+  /**
+   * Get the list of databases that are stored in file system
+   * @return List<DatabaseMeta> Returns a List of DatabaseMeta
+   * @throws KettleXMLException
+   */
   public List<DatabaseMeta> getDatabases() throws KettleXMLException {
```
abryant-hv, line 58: this isn't overriding anything?

engine/src/main/java/org/pentaho/di/shared/DatabaseConnectionManager.java
```diff
@@ -68,12 +75,38 @@ private void populateDbMetaMap( Map<String, Node> nodesMap ) throws KettleXMLExc
 
   }
 
+  /**
+   * Add a new database connection. The new connection is added to xml file and also in the in memory map.
+   * @param databaseMeta
+   * @return boolean
+   * @throws KettleException
+   */
+  public synchronized boolean addDatabase( DatabaseMeta databaseMeta ) throws KettleException {
```
abryant-hv, line 84: yeah, these should have `@Override`, and probably have `@inheritdoc` in their javadoc.

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonDBDelegate.java
```diff
@@ -474,12 +475,33 @@ public void saveConnection( DatabaseMeta db, String versionComment ) {
...
+      databaseMeta.setName( con_name );
+      databaseMeta.setDisplayName( con_name );
+      databaseMeta = getDatabaseDialog().getDatabaseMeta();
+      try {
+        DatabaseManagementInterface databaseManagementInterface = spoon.getBowl().getManager( DatabaseManagementInterface.class );
+
+        if ( databaseMeta.findDatabase( databaseManagementInterface.getDatabases(), con_name ) == null ) {
+          databaseManagementInterface.addDatabase( databaseMeta );
+          if ( spoon.rep != null ) {
+            saveConnectionToRepository( databaseMeta );
```
abryant-hv, line 493: why are we saving to the repository? I thought the repository was "Local".

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonDBDelegate.java
```diff
@@ -525,6 +535,23 @@ public void newConnection( HasDatabasesInterface hasDatabasesInterface ) {
     }
   }
 
+  private void saveConnectionToRepository( DatabaseMeta databaseMeta ) {
```
abryant-hv, line 538: this should only be needed on Edit or Undo

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonDBDelegate.java
pdesai16: This is an existing flow. Currently, the new connection gets saved in repository if the user is connected to repository. I kept the flow since we haven't groomed the repository stories

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonDBDelegate.java
pdesai16: same as above

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonDBDelegate.java
abryant-hv: I think this is in this story. Repository == local, and syncing to local is precisely what this story is supposed to be removing.

abryant-hv: oh, that isn't this story, it's the "rip out shared/unshared" story that is already supposed to be dev complete. This would add back behavior that the other story removed.

core/src/main/java/org/pentaho/di/shared/VfsSharedObjectsIO.java
pdesai16: Not sure what kind of comment you like to see added

abryant-hv: Was thinking something like "Does not need a bowl because the path to a project can't use a project VFS connection"

abryant-hv: but really not a big deal.

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonDBDelegate.java
pdesai16: I thought we are not changing/breaking the code that works with the repository i.e leaving it as is. 
That is why I didn't touch it in previous "rip out share/unshare" story and this story

abryant-hv: The plan for removing share/unshare, plus this story, means there should no longer be any way to create "Local" connections anywhere. This isn't "breaking" repository, it's just moving forward with the feature.

abryant-hv (approved):

bconlon-hv (approved):

PR #9488, opened 2024-07-22
[BACKLOG-40339] - more plugin web functions

abryant-hv:
- Allow plugins to add web functions with names
- add a function to check whether another function is defined


bconlon-hv (approved):

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -1860,7 +1867,7 @@ public void run() {
         }
       };
 
-      HashMap<String, Runnable> functions = new HashMap<String, Runnable>();
+      HashMap<String, Runnable> functions = new HashMap<String, Runnable>( browserFunctions );
```
tgf, line 1870: would it be better to add at the end, to have the option to override functions?

tgf (approved):

PR #9508, opened 2024-07-30
[BACKLOG-40797] - work in progress on edit db connections

abryant-hv:
- Updated right-click menu mapping to new layout
- refactored database connections in AbstractMeta to delegate to Bowls as appropriate
- separated read and write paths for AbstractMeta db connections
- implemented Edit and Delete in Db Connection UI


abryant-hv:
DRAFT ONLY. DO NOT MERGE.

core/src/main/java/org/pentaho/di/shared/DelegatingSharedObjectsIO.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.List;
+import java.util.Map;
+import org.w3c.dom.Node;
+
+
+/**
+ * A Read-only wrapper around multiple other SharedObjectsIO implementations that combines them by precedence order.
+ *
+ */
+public class DelegatingSharedObjectsIO implements SharedObjectsIO {
```
pdesai16, line 35: This looks like a new file. We already have a DelegatingShareObjectsIO

abryant-hv: ah, the existing one was misspelled... "Share" vs "Shared"... that's why I didn't see it. 

It's unused and doesn't have any tests. 

The two are extremely similar, but the newer one uses `getSharedObject()` instead of `getSharedObjects` for the single-item API, which should be more efficient. I'll delete the old one.

ui/src/main/java/org/pentaho/di/ui/spoon/DatabasesCollector.java
```diff
@@ -40,36 +39,23 @@
...
     dbMetaMap = new HashMap<String, DatabaseMeta>();
     List<DatabaseMeta> dbMetaList;
 
-    // SharedObjects
     if ( dbManager != null ) {
       dbMetaList = dbManager.getDatabases();
       addToMetaMap( dbMetaList );
     }
 
-    // local databases
```
pdesai16, line 67: This is going to break how we display the local db connections in UI

core/src/main/java/org/pentaho/di/shared/DelegatingSharedObjectsIO.java
pdesai16: It is used in project-plugin. You may have to check that

abryant-hv: I did check there... I don't see any usage... where is it used specifically? 
I *do* use the new class there.

ui/src/main/java/org/pentaho/di/ui/spoon/DatabasesCollector.java
abryant-hv: See the calling code. Now AbstractMeta *also* uses DatabaseManagementInterface, so there isn't a reason to use both modes.

ui/src/main/java/org/pentaho/di/ui/spoon/tree/provider/DBConnectionFolderProvider.java
```diff
@@ -99,7 +99,8 @@ public void refresh( Optional<AbstractMeta> meta, TreeNode treeNode, String filt
 
       // Local Db connection
       if ( meta.isPresent() ) {
-        collector = new DatabasesCollector( null, meta.get(), null );
+        DatabaseManagementInterface dbManager = meta.get().getDatabaseManagementInterface();
+        collector = new DatabasesCollector( dbManager,  null );
```
abryant-hv, line 103: This is where local databases are now read for the UI

plugins/pur/core/src/test/java/com/pentaho/repository/importexport/StreamToTransNodeConverterTest.java
```diff
@@ -54,13 +57,19 @@
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 import static org.pentaho.di.core.util.Assert.assertTrue;
+import org.pentaho.di.shared.MemorySharedObjectsIO;
```
abryant-hv, line 60: fix import order

ui/src/main/java/org/pentaho/di/ui/core/dialog/EnterSelectionDialog.java
```diff
@@ -162,11 +162,11 @@ public EnterSelectionDialog( Shell parent, String[] choices, String shellText, S
...
-  public EnterSelectionDialog( Shell parent, String[] choices, String shellText, String message,
-    HasDatabasesInterface databasesInterface ) {
-    this( parent, choices, shellText, message );
-    this.databasesInterface = databasesInterface;
-  }
+  //public EnterSelectionDialog( Shell parent, String[] choices, String shellText, String message,
+  //  HasDatabasesInterface databasesInterface ) {
+  //  this( parent, choices, shellText, message );
+  //  this.databasesInterface = databasesInterface;
+  //}
```
abryant-hv, line 169: delete

pdesai16 (approved):

tgf (approved):

PR #9538, opened 2024-08-19
[BACKLOG-41332] Implemented Move/Copy DB Connections and remaining DB…

pdesai16:
… Actions

on right click in the left hand configuration/database node based on project profile rule

### DIFF
--- a/ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
+++ b/ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
@@ -2749,21 +2749,26 @@ public class Spoon extends ApplicationWindow implements AddUndoPositionInterface
     try {
       dbManager = getDbManager( leveledSelection.getLevel() );
       databaseMeta = dbManager.getDatabase( leveledSelection.getName() );
+      delegates.db.editConnection( dbManager, databaseMeta );
     } catch ( Exception e ) {
       new ErrorDialog( shell, "Error", "Unexpected error retrieving Database Connection", e );
     }
-    delegates.db.editConnection( dbManager, databaseMeta );
   }
 
   public void dupeConnection() {
-    final DatabaseMeta databaseMeta = (DatabaseMeta) selectionObject;
-    final AbstractMeta abstractMeta = (AbstractMeta) selectionObjectParent;
-    delegates.db.dupeConnection( abstractMeta, databaseMeta );
-  }
+    SpoonTreeLeveledSelection leveledSelection = (SpoonTreeLeveledSelection) selectionObject;
+    DatabaseManagementInterface dbManager = null;
+    DatabaseMeta databaseMeta = null;
+    try {
+      dbManager = getDbManager( leveledSelection.getLevel() );
+      databaseMeta = dbManager.getDatabase( leveledSelection.getName() );
 
-  public void clipConnection() {
-    final DatabaseMeta databaseMeta = (DatabaseMeta) selectionObject;
-    delegates.db.clipConnection( databaseMeta );
+      getLog().logBasic( "Duplicating the connection " + leveledSelection.getName() );
+      delegates.db.dupeConnection( databaseMeta, dbManager );
+    } catch ( Exception e ) {
+      new ErrorDialog( shell, BaseMessages.getString( PKG, "Spoon.ErrorDialog.Title" ),
+        BaseMessages.getString( PKG, "Unexpected error retrieving Database Connection", databaseMeta.getName() ), e );
+    }
   }
 
   public void delConnection() {
@@ -2779,6 +2784,7 @@ public class Spoon extends ApplicationWindow implements AddUndoPositionInterface
       databaseMeta = dbManager.getDatabase( leveledSelection.getName() );
     } catch ( Exception e ) {
       new ErrorDialog( shell, "Error", "Unexpected error retrieving Database Connection", e );
+      return;
     }
 
     MessageBox mb = new MessageBox( shell, SWT.YES | SWT.NO | SWT.ICON_QUESTION );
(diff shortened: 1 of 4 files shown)


ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -2795,16 +2800,35 @@ public void delConnection() {
...
+    DatabaseManagementInterface dbManager = null;
+    DatabaseMeta databaseMeta = null;
+    try {
+      dbManager = getDbManager( leveledSelection.getLevel() );
+      databaseMeta = dbManager.getDatabase( leveledSelection.getName() );
+    } catch ( Exception e ) {
+      new ErrorDialog( shell, BaseMessages.getString( PKG, "Spoon.ErrorDialog.Title" ),
+        BaseMessages.getString( PKG, "Unexpected error retrieving Database Connection", databaseMeta.getName() ), e );
+    }
     delegates.db.sqlConnection( databaseMeta );
```
bconlon-hv, line 2813: All of these calls to delegates.db should be inside the try block now right?  Some of them are and some aren't.

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonDBDelegate.java
```diff
@@ -124,29 +131,52 @@ private DatabaseDialog getDatabaseDialog() {
...
+        selectedName = selectedName.trim();
+        // check if the selectedName already exist
+        if ( selectedName.equals( originalName )
+            && databaseMeta.findDatabase( dbManager.getDatabases(), selectedName ) != null ) {
+          databaseMetaCopy.setName( selectedName );
+          DatabaseDialog.showDatabaseExistsDialog( spoon.getShell(), databaseMetaCopy );
+          return;
+        }
+        databaseMetaCopy.setName( selectedName );
+        databaseMetaCopy.setName( selectedName );
```
abryant-hv, line 160: probably don't need this twice. Was once supposed to be DisplayName?

abryant-hv (approved):

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
pdesai16: I reused the existing code and the SpoonDbDelegate code is not throwing any exception.

abryant-hv: Right, but if there was an exception in the try block, and the dialog was displayed, we probably do not want to proceed with the call to the delegate.

abryant-hv: And I see that you copied this from editConnection(), which I messed up in the first place. ;)

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -2795,16 +2800,35 @@ public void delConnection() {
...
+      DatabaseManagementInterface dbManager = null;
+      DatabaseMeta databaseMeta = null;
+      try {
+        dbManager = getDbManager( leveledSelection.getLevel() );
+        databaseMeta = dbManager.getDatabase( leveledSelection.getName() );
+      } catch ( Exception e ) {
+        new ErrorDialog( shell, BaseMessages.getString( PKG, "Spoon.ErrorDialog.Title" ),
+          BaseMessages.getString( PKG, "Unexpected error retrieving Database Connection", databaseMeta.getName() ), e );
+      }
       delegates.db.clearDBCache( databaseMeta );
```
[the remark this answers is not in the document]
pdesai16: Some of existing menu items that are  calling delegate method  like sqlConnection () and this method are not throwing any exception

niehusa (approved):

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
abryant-hv: I'm actually surprised the compiler isn't complaining that databaseMeta might not have been initialized.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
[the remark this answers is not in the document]
abryant-hv: right, see above. The risk is not that clearDBCache will throw, the risk is that getDatabase() could throw, and then clearDBCache would still be called, when it shouldn't. 

Alternatively, we could just `return` after showing the ErrorDialog.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -2756,14 +2756,19 @@ public void editConnection() {
...
+    DatabaseMeta databaseMeta = null;
+    try {
+      dbManager = getDbManager( leveledSelection.getLevel() );
+      databaseMeta = dbManager.getDatabase( leveledSelection.getName() );
+    } catch ( Exception e ) {
+      new ErrorDialog( shell, BaseMessages.getString( PKG, "Spoon.ErrorDialog.Title" ),
+        BaseMessages.getString( PKG, "Unexpected error retrieving Database Connection", databaseMeta.getName() ), e );
+    }
+    getLog().logBasic( "Duplicating the connection " + leveledSelection.getName() );
+    delegates.db.dupeConnection( databaseMeta, dbManager );
```
abryant-hv, line 2770: please move this one into the try block as well.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -2869,8 +2888,84 @@ public void exploreDatabase() {
...
+    DatabaseManagementInterface dbManager = null;
+    DatabaseMeta databaseMeta = null;
+    try {
+      dbManager = getDbManager( leveledSelection.getLevel() );
+      databaseMeta = dbManager.getDatabase( leveledSelection.getName() );
+    } catch ( Exception e ) {
+      new ErrorDialog( shell, BaseMessages.getString( PKG, "Spoon.ErrorDialog.Title" ),
+        BaseMessages.getString( PKG, "Unexpected error retrieving Database Connection", databaseMeta.getName() ), e );
+    }
+    delegates.db.exploreDB( databaseMeta, dbManager, true );
```
abryant-hv, line 2901: please move this one into the try block as well.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -2756,14 +2756,19 @@ public void editConnection() {
   }
```
abryant-hv, line 2756: please move the one here in editConnection() into the try block as well.

bconlon-hv (approved):

PR #9627, opened 2024-09-30
[BACKLOG-40522] Added changes to display the SlavesServes at Global, …

pdesai16:
…Project and local levels in the View Tab

 Added a Common SharedObjectManager and SharedObjectManagementInterface
 that support CRUD operation on the UI.


core/src/main/java/org/pentaho/di/core/database/DatabaseMeta.java
```diff
@@ -3142,4 +3142,8 @@ public String getNamedCluster() {
   public void setNamedCluster( String namedCluster ) {
     databaseInterface.setNamedCluster( namedCluster );
   }
+
+  public Object makeClone() {
+    return clone();
+  }
```
abryant-hv, line 3148: This doesn't override or implement some interface? 
If this is an override, it needs `@Override`

engine/src/main/java/org/pentaho/di/job/JobMeta.java
```diff
@@ -417,7 +419,7 @@ public Object realClone( boolean doClear ) {
         jobMeta.jobhops = new ArrayList<JobHopMeta>();
         jobMeta.notes = new ArrayList<NotePadMeta>();
         jobMeta.localSharedObjects.clear();
-        jobMeta.slaveServers = new ArrayList<SlaveServer>();
+        //jobMeta.slaveServers = new ArrayList<SlaveServer>();
```
abryant-hv, line 422: delete dead code.

engine/src/main/java/org/pentaho/di/job/JobMeta.java
```diff
@@ -436,8 +438,11 @@ public Object realClone( boolean doClear ) {
...
         jobMeta.localSharedObjects.saveSharedObject( dbType, entry.getKey(), entry.getValue().cloneNode( true ) );
       }
-      for ( SlaveServer slave : slaveServers ) {
-        jobMeta.getSlaveServers().add( (SlaveServer) slave.clone() );
+      //SlaveServers
+      String slaveServerType = SharedObjectsIO.SharedObjectType.SLAVESERVER.getName();
+      for ( Map.Entry<String, Node> entry : localSharedObjects.getSharedObjects( slaveServerType ).entrySet() ) {
+        // cloneNode is *probably* overkill
+        jobMeta.localSharedObjects.saveSharedObject( slaveServerType, entry.getKey(), entry.getValue().cloneNode( true ) );
       }
```
abryant-hv, line 446: This feels really verbose, like we're coupling all the meta classes with the shared object implementation even more tightly than it was before.

engine/src/main/java/org/pentaho/di/shared/PassthroughSharedObjectsManager.java
```diff
@@ -0,0 +1,90 @@
...
+        break;
+      case SLAVESERVER:
+        sharedObjectInterface = new SlaveServer( node );
+        break;
+      case PARTITIONSCHEMA:
+        // partitionSchemas
+        break;
+      case CLUSTERSCHEMA:
+        // clusterSchemas
+        break;
```
abryant-hv, line 83: I don't love this style. It would be better if we didn't have hard-coded logic for each type.

engine/src/main/java/org/pentaho/di/shared/SharedObjectsManagementInterface.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.pentaho.di.core.exception.KettleException;
+import java.util.List;
+
+/**
+ * This is the management interface used by the UI to perform CRUD operation for all shared objects. The implementors of this interface will
+ * be scoped based on the bowl and can be retrieved using bowl's getManager()
+ *
+ */
+public interface SharedObjectsManagementInterface {
```
abryant-hv, line 11: why can't this be generic? So we can have exact types in the calling code?

engine/src/main/java/org/pentaho/di/shared/SharedObjectsManagementInterface.java
```diff
@@ -0,0 +1,62 @@
+package org.pentaho.di.shared;
```
abryant-hv, line 1: Kettle code all needs copyright statements.

engine/src/main/java/org/pentaho/di/shared/SharedObjectsManager.java
```diff
@@ -0,0 +1,223 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * This class uses the SharedObjectsIO to retrieve and save shared objects. This is used by the UI.
+ * <p>
+ * This class caches the state of the underlying SharedObjectsIO, and does not re-read from the source. Only changes
+ * written through this interface will be reflected.
+ */
+public class SharedObjectsManager implements SharedObjectsManagementInterface {
```
abryant-hv, line 27: this should really be generic

engine/src/main/java/org/pentaho/di/shared/SharedObjectsManager.java
```diff
@@ -0,0 +1,223 @@
...
+            if ( !slaveServerMap.containsKey( name ) ) {
+              slaveServerMap.put( name, slaveServerSharedObject );
+            }
+          }
+          sharedObjectsMap.put( SLAVESERVER_TYPE, slaveServerMap );
+        }
+      }
+      slaveServer_initialized = true;
+    }
+  }
```
abryant-hv, line 125: This is a lot of copy-paste code. The only variables here are the type string and the constructor call. Those could be refactored out as arguments.

engine/src/main/java/org/pentaho/di/shared/SharedObjectsManager.java
```diff
@@ -0,0 +1,223 @@
...
+   */
+  @Override
+  public synchronized void add( String type, SharedObjectInterface sharedObjectInterface )  throws KettleException {
+    populateSharedObjectMap( type );
+    String name = sharedObjectInterface.getName();
+    Node node = toNode( type, sharedObjectInterface );
+    sharedObjectsIO.saveSharedObject( type, name, node );
+
+    // Add it to the map
+    Map<String, SharedObjectInterface> localSharedObjectsMap = sharedObjectsMap.get( type );
```
abryant-hv, line 142: doesn't this need to computeIfAbsent? 

the calls to computeIfAbsent should maybe be refactored out to a helper method.

engine/src/main/java/org/pentaho/di/shared/SharedObjectsManager.java
```diff
@@ -0,0 +1,223 @@
...
+   * Get the SharedObjectInterface object for the type using the name
+   * @param type The type of sharedObject
+   * @param name name of the SharedObject
+   * @return
+   * @throws KettleException
+   */
+  @Override
+  public SharedObjectInterface getSharedObject( String type, String name ) throws KettleException {
+    populateSharedObjectMap( type );
+    Map<String, SharedObjectInterface> localSharedObjectsMap = sharedObjectsMap.get( type );
```
abryant-hv, line 171: same, doesn't this need computeIfAbsent?

engine/src/main/java/org/pentaho/di/shared/SharedObjectsManager.java
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+
+  @Override
+  public synchronized void removeSharedObject( String type, String sharedObjectName ) throws KettleException {
+    populateSharedObjectMap( type );
+
+    this.sharedObjectsIO.delete( type, sharedObjectName );
+    // Remove the sharedObject from the map
+    Map<String, SharedObjectInterface> localSharedObjectsMap = sharedObjectsMap.get( type );
```
abryant-hv, line 194: same, doesn't this need computeIfAbsent?

engine/src/main/java/org/pentaho/di/shared/SharedObjectsManager.java
```diff
@@ -0,0 +1,223 @@
...
+    localSharedObjectsMap.remove( sharedObjectName );
+  }
+
+  @Override
+  public synchronized void clear( String type ) throws KettleException {
+    this.sharedObjectsIO.clear( type );
+    // reset the initialized flags so sharedObjectMap can be repopulated
+    dbconn_initialized = false;
+    slaveServer_initialized = false;
+
```
abryant-hv, line 204: should we also clear the sharedObjectsMap? Would release some memory, if nothing else.

engine/src/main/java/org/pentaho/di/trans/TransMeta.java
```diff
@@ -2526,10 +2529,13 @@ public String getXML( boolean includeSteps, boolean includeDatabase, boolean inc
...
       retval.append( "    " ).append( XMLHandler.openTag( XML_TAG_SLAVESERVERS ) ).append( Const.CR );
-      for ( int i = 0; i < slaveServers.size(); i++ ) {
-        SlaveServer slaveServer = slaveServers.get( i );
+      for ( SharedObjectInterface slaveServer : localSharedObjectsMgr.getSharedObjects( SharedObjectsManager.SLAVESERVER_TYPE ) ) {
         retval.append( slaveServer.getXML() );
       }
+      /*for ( int i = 0; i < slaveServers.size(); i++ ) {
+        SlaveServer slaveServer = slaveServers.get( i );
+        retval.append( slaveServer.getXML() );
+      }*/
```
abryant-hv, line 2538: remove dead code.

ui/src/main/java/org/pentaho/di/ui/spoon/tree/provider/SlavesFolderProvider.java
```diff
@@ -52,23 +63,62 @@ public SlavesFolderProvider() {
...
-    if ( !meta.isPresent() ) {
-      return;
-    }
-    List<SlaveServer> servers = meta.get().getSlaveServers();
+    Bowl currentBowl = Spoon.getInstance().getBowl();
+    try {
+      Set<String> projectSlaveServerNames = new HashSet<>();
+      // Bowl specific
+      if ( currentBowl != DefaultBowl.getInstance() ) {
+        //SlaveServerManagementInterface slaveServerManager = currentBowl.getManager( SlaveServerManagementInterface.class );
```
abryant-hv, line 71: remove dead code

ui/src/main/java/org/pentaho/di/ui/spoon/tree/provider/SlavesFolderProvider.java
```diff
@@ -52,23 +63,62 @@ public SlavesFolderProvider() {
...
+        org.pentaho.di.shared.SharedObjectsManagementInterface slaveServerManager = currentBowl.getManager( org.pentaho.di.shared.SharedObjectsManagementInterface.class );
```
abryant-hv, line 72: why the full package? Is there more than one SharedObjectsManagementInterface?

ui/src/main/java/org/pentaho/di/ui/spoon/tree/provider/SlavesFolderProvider.java
```diff
@@ -52,23 +63,62 @@ public SlavesFolderProvider() {
...
+
+        globalServerNames.add( slaveServer.getName() );
+        TreeNode childTreeNode = createTreeNode( treeNode, slaveServer.getName(), guiResource.getImageSlaveTree(),
+          LeveledTreeNode.LEVEL.GLOBAL, projectSlaveServerNames.contains( slaveServer.getName() ) );
       }
 
+      // Local SlaveServers
+      if ( meta.isPresent() ) {
+        SharedObjectsManagementInterface localServerMgr = meta.get().getSharedObjectsManagementInterface();
+        List<SharedObjectInterface> localServers = localServerMgr.getSharedObjects( SharedObjectsManager.SLAVESERVER_TYPE );
```
abryant-hv, line 107: This style, where the calling code has to pass a constant, rather than there being a specific manager instance for each type, will make it harder to refactor this to use the metastore later. I think we need a manager for each type. Those managers could be almost entirely implemented in a base class, but they have to each have a separate instance, I think, for the later conversion to the metastore to work the way we have discussed. 

Like, with the current DatabaseManagementInterface that we added, refactoring DatabaseManagementInterface to be implemented on top of a metastore is trivial, because the calling code doesn't know any of the internal implementation. This code will not be as easy because of the exposed internals.

abryant-hv (changes requested):

ui/src/main/java/org/pentaho/di/ui/spoon/tree/provider/SlavesFolderProvider.java
```diff
@@ -52,23 +62,58 @@ public SlavesFolderProvider() {
...
-
-    servers.sort( ( s1, s2 ) -> String.CASE_INSENSITIVE_ORDER.compare( s1.getName(), s2.getName() ) );
+    Bowl currentBowl = Spoon.getInstance().getBowl();
+    try {
+      Set<String> projectSlaveServerNames = new HashSet<>();
+      // Bowl specific
+      if ( currentBowl != DefaultBowl.getInstance() ) {
+        SlaveServerManagementInterface slaveServerManager = currentBowl.getManager( SlaveServerManagementInterface.class );
+        List<SlaveServer> slaveServers = slaveServerManager.getAll();
+        slaveServers.sort( ( s1, s2 ) -> String.CASE_INSENSITIVE_ORDER.compare( s1.getName(), s2.getName() ) );
```
abryant-hv, line 72: why are we sorting this? The tree should sort itself.

ui/src/test/java/org/pentaho/di/ui/spoon/SharedObjectSyncUtilTest.java
```diff
@@ -116,7 +116,7 @@ public void tearDown() throws Exception {
     }
   }
 
-  @Test
+ /* @Test
```
abryant-hv, line 119: ???

ui/src/test/java/org/pentaho/di/ui/spoon/SharedObjectSyncUtilTest.java
```diff
@@ -225,7 +225,7 @@ public void synchronizePartitionSchemasDeleteFromRepository() throws Exception {
   }
 
 
-  @Test
+  /*@Test
   public void synchronizeSlaveServers() throws Exception {
```
abryant-hv, line 229: ??? 

should leave a TODO or something.

ui/src/test/java/org/pentaho/di/ui/spoon/SpoonRefreshSlavesSubtreeTest.java
```diff
@@ -22,21 +22,24 @@
 
 package org.pentaho.di.ui.spoon;
 
-import org.junit.Assert;
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.*;
```
abryant-hv, line 25: don't import .*

abryant-hv (changes requested):

engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
```diff
@@ -307,6 +306,7 @@ public void clear() throws KettleException {
   }
 
   protected final DatabaseManagementInterface localDbMgr = new MetaDatabaseManager( localSharedObjects );
+  protected final SlaveServerManagementInterface localSharedObjectsMgr = new PassthroughSlaveServerManager( localSharedObjects );
```
abryant-hv, line 309: This is still named weirdly. Maybe localSlaveServerManager?

engine/src/main/java/org/pentaho/di/job/JobMeta.java
```diff
@@ -1167,7 +1158,7 @@ public void loadXML( Node jobnode, String fname, Repository rep, IMetaStore meta
         Node entrynode = XMLHandler.getSubNodeByNr( entriesnode, "entry", i );
         // System.out.println("Reading entry:\n"+entrynode);
 
-        JobEntryCopy je = new JobEntryCopy( entrynode, getDatabases(), slaveServers, rep, metaStore );
+        JobEntryCopy je = new JobEntryCopy( entrynode, getDatabases(), getSlaveServers(), rep, metaStore );
```
abryant-hv, line 1161: is this returning *all* the slave servers, or just the local ones? I suppose the same risk is true for getDatabases().

engine/src/main/java/org/pentaho/di/partition/PartitionSchema.java
```diff
@@ -26,8 +26,10 @@
 import java.util.Date;
 import java.util.List;
 
+import org.pentaho.di.cluster.ClusterSchema;
```
abryant-hv, line 29: unused?

engine/src/main/java/org/pentaho/di/shared/PassthroughSlaveServerManager.java
```diff
@@ -0,0 +1,67 @@
+package org.pentaho.di.shared;
```
abryant-hv, line 1: needs copyright

ui/src/main/java/org/pentaho/di/ui/cluster/dialog/SlaveServerDialog.java
```diff
@@ -603,4 +610,10 @@ public void test() {
         + BaseMessages.getString( PKG, "SlaveServer.ExceptionUnableGetReplay.Error2" ), e );
     }
   }
+
+  private void showMessage( String message ){
```
abryant-hv, line 614: `){` -> `) {`

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -2971,6 +2969,87 @@ public void copyConnectionToProject() {
...
+    SlaveServer slaveServer = null;
+    try {
+      slaveServerManager = getSlaveServerManager( leveledSelection.getLevel() );
+      slaveServer = slaveServerManager.get( leveledSelection.getName() );
+      getLog().logBasic( "Moving the slave server " + leveledSelection.getName() + " to project" );
+      delegates.slaves.copyToProject( slaveServerManager, slaveServer );
+    } catch ( Exception e ) {
+      new ErrorDialog( shell, "Error", "Unexpected error retrieving slave servers", e );
+    }
+  }
```
abryant-hv, line 3038: this is a lot of copy-paste code. Can we combine these?

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -3064,20 +3143,66 @@ public void monitorClusterSchema() throws KettleException {
...
+   * Edit the Slave server configuration
+   */
   public void editSlaveServer() {
-    final SlaveServer slaveServer = (SlaveServer) selectionObject;
-    editSlaveServer( slaveServer );
+    SpoonTreeLeveledSelection leveledSelection = (SpoonTreeLeveledSelection) selectionObject;
+    editSlaveServer( leveledSelection );
+  }
+
+  private void editSlaveServer(SpoonTreeLeveledSelection leveledSelection ) {
```
abryant-hv, line 3154: space in the parens

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -3064,20 +3143,66 @@ public void monitorClusterSchema() throws KettleException {
...
+      return managementBowl.getManager( SlaveServerManagementInterface.class );
+    } else if ( level == LeveledTreeNode.LEVEL.GLOBAL ) {
+      return DefaultBowl.getInstance().getManager( SlaveServerManagementInterface.class );
+    } else if ( level == LeveledTreeNode.LEVEL.LOCAL ) {
+      AbstractMeta meta = getActiveAbstractMeta();
+      return meta != null ? meta.getSlaveServerManagementInterface() : null;
+    } else {
+      // shouldn't happen.
+      return null;
+    }
```
abryant-hv, line 3177: This could almost be combined with getDBManager. Just need a way to get a generic manager from the meta.

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonSlaveDelegate.java
```diff
@@ -70,80 +80,194 @@ public void addSpoonSlave( SlaveServer slaveServer ) {
...
-    hasSlaveServersInterface.getSlaveServers().remove( slaveServer );
-    refreshTree();
+    spoon.getLog().logBasic( "Deleting the slave server " +  slaveServer.getName() );
+
+    try {
+      slaveServerManager.remove( slaveServer );
 
+      Repository rep = spoon.getRepository();
+      if ( rep != null && slaveServer.getObjectId() != null ) {
+        // remove the slave server from the repository too...
```
abryant-hv, line 100: This is going to be interesting. See the recent spike findings. Nothing to worry about in this PR, though.

engine/src/main/java/org/pentaho/di/job/JobMeta.java
pdesai16: This is returning all slave servers. I tried looking at the usage  of various JobEntryInterface implementation to see how they are using the list of databases and SlaveServers and only one I found was in JobEntryCheckDbConnections which was trying to find the database connection name in the list of databaseConnections that was passed in. Haven't found any usage of slave server list.
But looking at the fact that it is job entry, if the KJB  has a KTR which has the db connection, the question is can it refer to any DB connection or only that is defined in the parent KJB. I think it should be  all connections and slave servers not just local.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
pdesai16: I did look at the different options. I could move the logic to delegate so the delegate method would be
 delegates.slaves.copyToProject( SpoonTreeLeveledSelection leveledSelection );
But I am just moving the code from Spoon to delegate class, not reusing it. 
If we combine, then we would need a big switch statement to calls different method on delegate based on parameter.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
pdesai16: Yeah... but not sure how to return a generic manager from AbstractMeta. May be we need to some logic like BaseBowl.getManager() ??

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
abryant-hv: Could do something like: 
```
  @FunctionalInterface
  private static interface SlaveServerOp {
    void op( SlaveServerManager ssm, slaveServer ss ) throws SomethingUsefulException;
  }

  private void withSlaveServer( String message, SlaveServerOp sso ) {
    SpoonTreeLeveledSelection leveledSelection = (SpoonTreeLeveledSelection) selectionObject;
    SlaveServerManagementInterface slaveServerManager = null;
    SlaveServer slaveServer = null;
    try {
      slaveServerManager = getSlaveServerManager( leveledSelection.getLevel() );
      slaveServer = slaveServerManager.get( leveledSelection.getName() );
      getLog().logBasic(  );
      sso.op( slaveServerManager, slaveServer );
    } catch ( Exception e ) {
      new ErrorDialog( shell, "Error", "Unexpected error retrieving slave servers", e );
    }
  }

  /**
   * Copy the slaveServer to project
   */
  public void copySlaveServerToProject() {
    withSlaveServer( "Copying the slave server " + leveledSelection.getName() + " to project", 
      ( ssm, ss ) -> delegates.slaves.copyToProject( ssm, ss ) );
  }
```

NOTE: the log messages are wrong for copy, too.

engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
```diff
@@ -263,6 +260,8 @@ public void setAlwaysShowRunOptions( boolean alwaysShowRunOptions ) {
       new DelegatingSharedObjectsIO( bowl.getSharedObjectsIO(), localSharedObjects );
   protected DatabaseManagementInterface readDbManager = new PassthroughDbConnectionManager( combinedSharedObjects );
 
+  protected SlaveServerManagementInterface readSlaveServerManager = new PassthroughSlaveServerManager( combinedSharedObjects );
+
   protected void initializeLocalDatabases() {
```
abryant-hv, line 265: I just realized that this needs to be updated to reset all the shared object managers, not just databases. 

See the comment for why. :)

engine/src/main/java/org/pentaho/di/job/JobMeta.java
abryant-hv: I still see this dead code...

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
abryant-hv: I don't think we need the registration part. Maybe just a switch/case?

engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
abryant-hv: line 266-267. 
```// NOTE: this has to assign new objects, not just clear existing ones, because it is used in clone(), and
    // updating the original objects will update the source of the clone.```

PR #9677, opened 2024-10-24
[BACKLOG-40523] - update ClusterSchemas for projects

abryant-hv:
- refactor Database managers into the new SharedObjectInterface stack
- refactored a few things in Spoon to make them use common code for shared objects
- refactor TransMeta to use local managers instead of a list for ClusterSchemas


ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -2662,9 +2665,7 @@ public void newDatabasePartitioningSchema() {
 
   public void newClusteringSchema() {
     TransMeta transMeta = getActiveTransformation();
-    if ( transMeta != null ) {
-      newClusteringSchema( transMeta );
-    }
+    newClusteringSchema( transMeta );
```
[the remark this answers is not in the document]
abryant-hv: No, so we now check this in the Delegate class instead.

abryant-hv:
> Do we still need Spoon.setDatabaseMenuItems(). Can the Database right click menu items use the TreePopupMenuProvider to provide the right click menu actions like the rest of Shared objects

It probably could, but there were enough subtle differences that I didn't switch it over. Same with the Delegates.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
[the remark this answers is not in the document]
pdesai16: Is there a difference between getting the slaveServers from transMeta vs getting from slaveServerManagementInterface.getAll(). 
I have to do something similar for PartitionSchema and want to make sure I am not missing anything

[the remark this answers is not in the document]
abryant-hv: For *Execution* (including selecting from a list to be referenced by something else) they should always be gotten from the combined list, which is generally the Meta, but could be the Execution Bowl if there is no Meta. 

For duplicate checks, only the current level should be used.

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonClustersDelegate.java
```diff
@@ -85,27 +104,42 @@ private void showSaveError( ClusterSchema clusterSchema, KettleException e ) {
...
     }
   }
 
-  public void delClusterSchema( TransMeta transMeta, ClusterSchema clusterSchema ) {
+  public void delClusterSchema( ClusterSchemaManagementInterface manager, ClusterSchema clusterSchema ) {
     try {
-
-      int idx = transMeta.getClusterSchemas().indexOf( clusterSchema );
-      transMeta.getClusterSchemas().remove( idx );
+      manager.remove( clusterSchema );
```
pdesai16, line 142: Is it  prompting for confirmation dialog before deleting

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonClustersDelegate.java
```diff
@@ -85,27 +104,42 @@ private void showSaveError( ClusterSchema clusterSchema, KettleException e ) {
...
+      // execution bowl for listing all slave servers
+      SlaveServerManagementInterface slaveServerManagementInterface =
+        spoon.getExecutionBowl().getManager( SlaveServerManagementInterface.class );
+
+      ClusterSchemaDialog dialog =
+          new ClusterSchemaDialog( spoon.getShell(), clusterSchema, manager.getAll(),
+                                   slaveServerManagementInterface.getAll() );
+      if ( dialog.open() ) {
+        String newName = clusterSchema.getName().trim();
+        manager.add( clusterSchema );
```
pdesai16, line 119: Is there a check to make sure there is no deplicate before adding

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonClustersDelegate.java
abryant-hv: It does not appear so. This is the existing behavior, and the story doesn't say it should be any different.

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonClustersDelegate.java
abryant-hv: The ClusterSchemaDialog handles that.

pdesai16 (approved):

PR #9697, opened 2024-11-07
[BACKLOG- 42745] Fix: select partition on step only includes Local

pdesai16:
- Fixed to ensure the getAll() is not called in loop

### DIFF
--- a/engine/src/main/java/org/pentaho/di/trans/TransMeta.java
+++ b/engine/src/main/java/org/pentaho/di/trans/TransMeta.java
@@ -2543,7 +2543,8 @@ public class TransMeta extends AbstractMeta
     //
     if ( includePartitions ) {
       retval.append( "    " ).append( XMLHandler.openTag( XML_TAG_PARTITIONSCHEMAS ) ).append( Const.CR );
-      for ( PartitionSchema partitionSchema : localPartitionSchemaMgr.getAll() ) {
+      List<PartitionSchema> partitionSchemas = localPartitionSchemaMgr.getAll();
+      for ( PartitionSchema partitionSchema : partitionSchemas ) {
         retval.append( partitionSchema.getXML() );
       }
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/pentaho/di/trans/TransMeta.java
```diff
@@ -2543,7 +2543,8 @@ public String getXML( boolean includeSteps, boolean includeDatabase, boolean inc
     //
     if ( includePartitions ) {
       retval.append( "    " ).append( XMLHandler.openTag( XML_TAG_PARTITIONSCHEMAS ) ).append( Const.CR );
-      for ( PartitionSchema partitionSchema : localPartitionSchemaMgr.getAll() ) {
+      List<PartitionSchema> partitionSchemas = localPartitionSchemaMgr.getAll();
+      for ( PartitionSchema partitionSchema : partitionSchemas ) {
```
abryant-hv, line 2547: This type isn't an issue. getAll() is only called once here.

engine/src/main/java/org/pentaho/di/trans/TransMeta.java
```diff
@@ -3576,8 +3580,9 @@ public void loadXML( Node transnode, String fname, IMetaStore metaStore, Reposit
 
         // Have all step clustering schema meta-data reference the correct cluster schemas that we just loaded
         //
+        List<ClusterSchema> clusterSchemas = localClusterSchemaManager.getAll();
         for ( int i = 0; i < nrSteps(); i++ ) {
-          getStep( i ).setClusterSchemaAfterLoading( localClusterSchemaManager.getAll() );
+          getStep( i ).setClusterSchemaAfterLoading( clusterSchemas );
```
abryant-hv, line 3585: This is an issue because it's called inside the loop. good fix.

niehusa (approved):

abryant-hv (approved):

bconlon-hv (approved):

PR #9750, opened 2024-12-05
[BACKLOG-42827] - CRUD for Global objects in repository

abryant-hv:
- RepositorySharedObjectsIO and test
    - fixed file-open-save for repositories (partial backport)
    - stopped saving or loading shared objects to the repository as part of
      transformations or jobs.
    - Still links steps to databases
    - Keep ObjectId in various Node-based memory structures, but don't store on
      disk in VfsSharedObjectsIO
    - Reset DefaultBowl managers on connect/disconnect from Repository

### DIFF
--- a/engine/src/main/java/org/pentaho/di/repository/filerep/KettleFileRepository.java
+++ b/engine/src/main/java/org/pentaho/di/repository/filerep/KettleFileRepository.java
@@ -958,7 +958,6 @@ public class KettleFileRepository extends AbstractRepository {
     jobMeta.setRepository( this );
     jobMeta.setMetaStore( MetaStoreConst.getDefaultMetastore() );
 
-    readDatabases( jobMeta, true );
     jobMeta.clearChanged();
 
     return jobMeta;
@@ -1118,44 +1117,9 @@ public class KettleFileRepository extends AbstractRepository {
     transMeta.setName( transname );
     transMeta.setObjectId( new StringObjectId( calcObjectId( repdir, transname, EXT_TRANSFORMATION ) ) );
 
-    readDatabases( transMeta, true );
     transMeta.clearChanged();
 
     return transMeta;
-  }
-
-  /**
-   * Read all the databases from the repository, insert into the has databases object, overwriting optionally
-   *
-   * @param TransMeta
-   *          The transformation to load into.
-   * @param overWriteShared
-   *          if an object with the same name exists, overwrite
-   * @throws KettleException
-   */
-  public void readDatabases( AbstractMeta transMeta, boolean overWriteShared ) throws KettleException {
-    try {
-      ObjectId[] dbids = getDatabaseIDs( false );
-      for ( int i = 0; i < dbids.length; i++ ) {
-        DatabaseMeta databaseMeta = loadDatabaseMeta( dbids[i], null ); // reads last version
-        if ( transMeta instanceof VariableSpace ) {
-          databaseMeta.shareVariablesWith( (VariableSpace) transMeta );
-        }
-
-        DatabaseMeta check = transMeta.findDatabase( databaseMeta.getName() ); // Check if there already is one in the
-                                                                               // transformation
-        if ( check == null || overWriteShared ) { // We only add, never overwrite database connections.
-          if ( databaseMeta.getName() != null ) {
-            transMeta.getDatabaseManagementInterface().add( databaseMeta );
-            if ( !overWriteShared ) {
-              databaseMeta.setChanged( false );
-            }
-          }
-        }
-      }
-    } catch ( KettleException e ) {
-      throw e;
-    }
   }
 
   public ValueMetaAndData loadValueMetaAndData( ObjectId id_value ) throws KettleException {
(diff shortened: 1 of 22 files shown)


engine/src/main/java/org/pentaho/di/repository/filerep/KettleFileRepository.java
```diff
@@ -958,7 +958,6 @@ public JobMeta loadJob( String jobname, RepositoryDirectoryInterface repdir, Pro
     jobMeta.setRepository( this );
     jobMeta.setMetaStore( MetaStoreConst.getDefaultMetastore() );
 
-    readDatabases( jobMeta, true );
```
pdesai16, line 961: Why is this removed?

engine/src/main/java/org/pentaho/di/shared/RepositorySharedObjectsIO.java
```diff
@@ -0,0 +1,239 @@
...
+  public RepositorySharedObjectsIO( Repository repository, SlaveServersSupplier slaveServerSupplier ) {
+    this.repository = Objects.requireNonNull( repository );
+    this.slaveServerSupplier = slaveServerSupplier;
+  }
+
+  @Override
+  public Map<String, Node> getSharedObjects( String type ) throws KettleException {
+    SharedObjectType objectType = SharedObjectType.valueOf( type.toUpperCase() );
+    List<? extends SharedObjectInterface> objects = null;
+    if ( repository instanceof RepositoryExtended ) {
```
pdesai16, line 60: In what case do you get the repositry that is instance of RepositoryExtended

engine/src/test/java/org/pentaho/di/shared/RepositorySharedObjectsIOTest.java
```diff
@@ -0,0 +1,249 @@
...
+  public static List<Object[]> repositories() {
+    ArrayList<Object[]> reps = new ArrayList<>();
+    reps.add( new Object[] { new MemoryRepository() } );
+    reps.add( new Object[] { new MemoryRepositoryExtended() } );
+    return reps;
+  }
+
+  @Before
+  public void setup() throws Exception {
+    //rep = new MemoryRepository();
```
pdesai16, line 64: comented code

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/ISharedObjectsTransformer.java
```diff
@@ -26,10 +26,13 @@
 import org.pentaho.di.shared.SharedObjects;
 
 public interface ISharedObjectsTransformer extends ITransformer {
+
+  @Deprecated
```
pdesai16, line 30: why is it deprecated. Can you add a comment

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/ISharedObjectsTransformer.java
```diff
@@ -26,10 +26,13 @@
...
   void loadSharedObjects( final RepositoryElementInterface element,
       final Map<RepositoryObjectType, List<? extends SharedObjectInterface>> sharedObjectsByType )
     throws KettleException;
 
+  @Deprecated
```
pdesai16, line 35: same as above

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/JobDelegate.java
```diff
@@ -180,43 +180,18 @@ public JobDelegate( final Repository repo, final IUnifiedRepository pur ) {
   // ~ Methods =========================================================================================================
   @SuppressWarnings( "unchecked" )
   @Override
+  @Deprecated // Shared Object reads should now go through a SharedObjectsIO
```
pdesai16, line 183: I think PUC uses these classes. Is this tested from PUC?

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/JobDelegate.java
```diff
@@ -180,43 +180,18 @@ public JobDelegate( final Repository repo, final IUnifiedRepository pur ) {
...
-
-    // Repository objects take priority so let's overwrite them...
-    //
-    readDatabases( jobMeta, true, (List<DatabaseMeta>) sharedObjectsByType.get( RepositoryObjectType.DATABASE ) );
-    readSlaves( jobMeta, true, (List<SlaveServer>) sharedObjectsByType.get( RepositoryObjectType.SLAVE_SERVER ) );
+    // NO-OP
   }
 
+  @Override
+  @Deprecated // Shared Object writes should now go through a SharedObjectsIO
```
pdesai16, line 191: same comment as above

engine/src/main/java/org/pentaho/di/repository/filerep/KettleFileRepository.java
abryant-hv: We no longer need to load shared objects from the repository into Metas. Shared objects from the repository are now accessed through the DefaultBowl.

engine/src/main/java/org/pentaho/di/shared/RepositorySharedObjectsIO.java
abryant-hv: FileRepository does not implement RepositoryExtended, but PurRepository does.

engine/src/test/java/org/pentaho/di/shared/RepositorySharedObjectsIOTest.java
abryant-hv: Good catch!

engine/src/test/java/org/pentaho/di/shared/RepositorySharedObjectsIOTest.java
```diff
@@ -0,0 +1,249 @@
...
+    shared = new RepositorySharedObjectsIO( rep, () -> Collections.emptyList() );
+  }
+
+  @Test
+  public void saveTest() throws Exception {
```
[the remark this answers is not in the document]
abryant-hv: I think it doesn't matter to junit, only the annotation does. I copied the style of some other existing file, but I forget which. 

I'll switch to testFoo to be more consistent.

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/ISharedObjectsTransformer.java
abryant-hv: added.

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/JobDelegate.java
abryant-hv: All the calls to loadSharedObjects I can find are in kettle. I think you are probably right about their use for import/export. The only remaining call I can see is in RepositoryImporter. 

*saveSharedObjects* is still used in a couple more places, though, so we need to be more careful of that. I think we should leave this deleted for now, and revisit it in the story where we do import/export.

pdesai16: I think it is also called during execution from PUC

abryant-hv: That should now also go through the bowl. We should probably have that as a story, though.

pdesai16 (approved):

PR #9798, opened 2025-01-27
[BACKLOG-42915] - Repository Bowl

abryant-hv:
Rather than set a SharedObjectsIO on the DefaultBowl, this switches to using a RepositoryBowl for the Repository. Spoon now tracks 3 different bowls: the existing management bowl and execution bowl, and now a global management bowl.

Earlier code that compared the management bowl to the DefaultBowl now compare to the global management bowl.

Repositories now have a Bowl. This ensures there's only one instance per Repository so caches work correctly.

AbstractMeta and Spoon both start using the Repository's bowl when a Repository is assigned, rather than assuming something else will set the bowl directly. The Project Plugin will need to set the bowl again where appropriate.

Lots of tests updated to have a Repository Bowl.

Refactored Import to use the Repository Bowl rather than a DefaultBowl with different SharedObjectsIO.

Added SharedObjectUtil with helper methods to move/copy all objects between AbstractMeta and Bowls.

### DIFF
--- a/engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
+++ b/engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
@@ -461,6 +461,11 @@ public abstract class AbstractMeta implements ChangedFlagInterface, UndoInterfac
   @Override
   public void setRepository( Repository repository ) {
     this.repository = repository;
+    if ( repository != null ) {
+      setBowl( repository.getBowl() );
+    } else {
+      setBowl( DefaultBowl.getInstance() );
+    }
   }
 
   /**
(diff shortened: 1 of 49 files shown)


engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
```diff
@@ -2,7 +2,7 @@
  *
  * Pentaho Data Integration
  *
- * Copyright (C) 2002-2024 by Hitachi Vantara : http://www.pentaho.com
+ * Copyright (C) 2002-2025 by Hitachi Vantara : http://www.pentaho.com
```
bconlon-hv, line 5: Aren't we supposed to leave these alone?

abryant-hv: On master we are. I'm just being paranoid. These will presumably all be replaced by the devops script.

bconlon-hv (approved):

tgf (approved):

PR #9811, opened 2025-02-04
[BACKLOG-43082] Make Shared Objects config case-insensitive for Manag…

pdesai16:
…ement and Execution

Changes include -
- Updated the left hand navigation tree to grey out the items if the item exist wth name (case  insensitive) at the higher precedence level
- The drop down for Db connection name, partitionSchema name, SlaveServer names are now sorted and deduplicated based on case insensitivity -move operation on all config objects in left hand navigation tree is updated to handle case insensitive checks
-Tested during execution that the sharedobjects are picked up based on the precedence level (project/global/local)

Additional changes
- Updated the SharedObjectsIO api to make the get(), save() and delete()   case insensitive 
- Added unit test for API changes

### DIFF
--- a/core/src/main/java/org/pentaho/di/connections/ConnectionManager.java
+++ b/core/src/main/java/org/pentaho/di/connections/ConnectionManager.java
@@ -699,7 +699,13 @@ public class ConnectionManager {
    */
   public ConnectionDetails getConnectionDetails( String name ) {
     initialize();
-    return clone( detailsByName.get( name ) );
+    // Get the details in ignoring the case
+    for ( String key : detailsByName.keySet() ) {
+      if ( key.equalsIgnoreCase( name ) ) {
+        return clone( detailsByName.get( key ) );
+      }
+    }
+    return null;
   }
 
   /* The following are sugar methods: `getDetails` and `getExistingDetails`. These perform casting of the result to the
@@ -939,4 +945,5 @@ public class ConnectionManager {
       return (MetaStoreException) super.getCause();
     }
   }
+
 }
--- a/core/src/main/java/org/pentaho/di/shared/DelegatingSharedObjectsIO.java
+++ b/core/src/main/java/org/pentaho/di/shared/DelegatingSharedObjectsIO.java
@@ -15,7 +15,6 @@ package org.pentaho.di.shared;
 
 import org.pentaho.di.core.exception.KettleException;
 
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -48,8 +47,11 @@ public class DelegatingSharedObjectsIO implements SharedObjectsIO {
     Map<String, Node> retMap = new HashMap<>();
     for ( SharedObjectsIO store : stores ) {
       Map<String, Node> storeMap = store.getSharedObjects( type );
-      for ( Map.Entry<String,Node> entry : storeMap.entrySet() ) {
-        retMap.putIfAbsent( entry.getKey(), entry.getValue() );
+      for ( Map.Entry<String, Node> entry : storeMap.entrySet() ) {
+        // case in-sensitive check to skip adding entry with same name
+        if ( !retMap.entrySet().stream().anyMatch( e -> e.getKey().equalsIgnoreCase( entry.getKey() ) ) ) {
+          retMap.put( entry.getKey(), entry.getValue() );
+        }
       }
     }
     return retMap;
--- a/core/src/main/java/org/pentaho/di/shared/VfsSharedObjectsIO.java
+++ b/core/src/main/java/org/pentaho/di/shared/VfsSharedObjectsIO.java
@@ -179,7 +179,8 @@ public class VfsSharedObjectsIO implements SharedObjectsIO {
   }
 
   /**
-   * Save the name and the node for the given Shared Object Type in the file
+   * Save the name and the node for the given Shared Object Type in the file. If the SharedObject name exist but in a
+   * different case, the existing entry will be deleted and the new entry will be saved with provided name
    * @param type The SharedObject type
    * @param name The name of the SharedObject for a given type
    * @param node The xml node that containing the details of the SharedObject
@@ -197,6 +198,12 @@ public class VfsSharedObjectsIO implements SharedObjectsIO {
       if ( childNode.getNodeName().equalsIgnoreCase( SharedObjectInterface.OBJECT_ID ) ) {
         node.removeChild( childNode );
       }
+    }
+    // Before adding, verify that the SharedObject with the name (case insensitive) does not exist,
+    // if it exist, delete the entry from map
+    String existingName = SharedObjectsIO.findSharedObjectIgnoreCase( name, nodeMap );
+    if ( existingName != null ) {
+      nodeMap.remove( existingName );
     }
 
     // Add or Update the map entry for this name
@@ -282,9 +289,10 @@ public class VfsSharedObjectsIO implements SharedObjectsIO {
   }
 
   /**
-   * Return the node for the given SharedObject type and name.
-   * @param type
-   * @param name
+   * Return the node for the given SharedObject type and name. The lookup for the SharedObject
+   * using the name will be case-insensitive
+   * @param type The type is shared object type for example, "connection", "slaveserver", "partitionschema" and clusterschema"
+   * @param name The name is the name of the sharedObject
    * @return
    * @throws KettleXMLException
    */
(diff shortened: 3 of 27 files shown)


bconlon-hv (approved):

core/src/main/java/org/pentaho/di/connections/ConnectionManager.java
```diff
@@ -939,4 +939,19 @@ public synchronized MetaStoreException getCause() {
       return (MetaStoreException) super.getCause();
     }
   }
+
+  /**
+   * Get the ConnectionDetails using the name in case insensitive manner
+   * @param name
+   * @return ConnectionDetails
+   */
+  public ConnectionDetails getConnectionDetailsIgnoreCase( String name ) {
```
abryant-hv, line 948: why do we need two APIs that do the same thing?

core/src/main/java/org/pentaho/di/connections/ConnectionManager.java
```diff
@@ -939,4 +939,19 @@ public synchronized MetaStoreException getCause() {
...
+    for (String key : detailsByName.keySet() ) {
```
abryant-hv, line 949: formatting

core/src/main/java/org/pentaho/di/connections/ConnectionManager.java
```diff
@@ -939,4 +939,19 @@ public synchronized MetaStoreException getCause() {
...
+      if ( key.equalsIgnoreCase( name ) ) {
+        return clone( detailsByName.get( key ) );
```
abryant-hv, line 951: Should we make sure there aren't any collisions in `initialize()`? Or do metastores already not allow duplicate keys with the same name but different case?

core/src/main/java/org/pentaho/di/shared/DelegatingSharedObjectsIO.java
```diff
@@ -49,7 +49,10 @@ public Map<String, Node> getSharedObjects( String type ) throws KettleException
     for ( SharedObjectsIO store : stores ) {
       Map<String, Node> storeMap = store.getSharedObjects( type );
       for ( Map.Entry<String,Node> entry : storeMap.entrySet() ) {
-        retMap.putIfAbsent( entry.getKey(), entry.getValue() );
+        // case in-sensitive check to skip adding entry with same name
+        if ( !retMap.entrySet().stream().anyMatch( e->e.getKey().equalsIgnoreCase( entry.getKey() ) ) ) {
+          retMap.put( entry.getKey(), entry.getValue() );
+        }
```
abryant-hv, line 55: Do we need to also update getSharedObject()? 
And delete()?

engine/src/main/java/org/pentaho/di/cluster/ClusterSchema.java
```diff
@@ -56,6 +57,7 @@ public class ClusterSchema extends ChangedFlag implements Cloneable, SharedObjec
   public static final String XML_TAG = "clusterschema";
 
   public static final RepositoryObjectType REPOSITORY_ELEMENT_TYPE = RepositoryObjectType.CLUSTER_SCHEMA;
+  public static final Comparator<ClusterSchema> comparator = Comparator.comparing( ClusterSchema::getName, String.CASE_INSENSITIVE_ORDER );
```
abryant-hv, line 60: constants should be all caps. Should be COMPARATOR (same in the other files)

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/ui/RunConfigurationDelegate.java
```diff
@@ -201,10 +201,12 @@ private void moveCopy( RunConfigurationManager srcManager, RunConfiguration runC
                          boolean deleteSource ) {
     CheckedMetaStoreSupplier ms = () -> targetBowl.getMetastore();
     RunConfigurationManager targetManager = RunConfigurationManager.getInstance( ms );
-    if ( targetManager.getNames().contains( runConfiguration.getName() ) ) {
+    if ( targetManager.getNames().stream().anyMatch( element->element.equalsIgnoreCase( runConfiguration.getName() ) ) ) {
```
abryant-hv, line 204: I think we usually do " -> " instead of "->" (i.e. surround with spaces)

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonDBDelegate.java
```diff
@@ -63,7 +63,7 @@
 import org.pentaho.di.ui.spoon.dialog.GetSQLProgressDialog;
 import org.pentaho.di.ui.spoon.tree.provider.DBConnectionFolderProvider;
 
-public class SpoonDBDelegate extends SpoonDelegate {
+public class SpoonDBDelegate extends SpoonSharedObjectDelegate {
```
abryant-hv, line 66: nice

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonSharedObjectDelegate.java
```diff
@@ -72,10 +72,15 @@ protected void moveCopy( SharedObjectsManagementInterface<T> srcManager,
...
     try {
       // If object already exist, prompt for overwrite
-      if ( findObject( targetManager.getAll(), object.getName() ) != null ) {
+      T existingObject = findObject( targetManager.getAll(), object.getName() );
+      if ( existingObject != null ) {
         if ( !shouldOverwrite( BaseMessages.getString( PKG, overWritePromptKey, object.getName() ) ) ) {
           return;
+        } else {
+          // Remove the existing object from target
+          targetManager.remove( existingObject );
```
abryant-hv, line 81: it strikes me that these could use a comment explaining *why* we need to remove before adding the new one. (i.e. that the case might change)

core/src/main/java/org/pentaho/di/connections/ConnectionManager.java
pdesai16: Metastore does not allow storing two entries with different case.

ui/src/main/java/org/pentaho/di/ui/spoon/delegates/SpoonSharedObjectDelegate.java
abryant-hv: Also, I think this is bad in another way. The APIs should be consistent. If get() is case insensitive, then so should remove() and add() be. 
add() should just work, replacing the existing object with the new object, if their names are the same in a case insensitive manner. We shouldn't need to remove first *only* if the case is different.

core/src/main/java/org/pentaho/di/shared/DelegatingSharedObjectsIO.java
pdesai16: Since DelegatingSharedObject is called during execution, the getSharedObject() and delete() should be doing a case sensitive comparison

abryant-hv:
I think we need unit tests on the case-insensitive API changes, too.

core/src/main/java/org/pentaho/di/shared/DelegatingSharedObjectsIO.java
abryant-hv: Wasn't this already case-insensitive during execution? Regardless, it should be case-insensitive for execution, now. 

(And I was wrong about delete(). DelegatingSharedObjectsIO doesn't support any write operations).

core/src/main/java/org/pentaho/di/connections/ConnectionManager.java
```diff
@@ -699,7 +699,13 @@ public ConnectionDetails getConnectionDetails( String key, String name ) {
    */
   public ConnectionDetails getConnectionDetails( String name ) {
```
abryant-hv, line 700: Since this API is now case-insensitive, the other ConnectionManager APIs should be as well, including save() (especially where it handles existing objects), delete(), exists(), getConnectionDetails( String key, String name), getDetails(), getExistingDetails(), the other flavors of getConnectionDetails(), and copy() (where existing objects are handled).

core/src/main/java/org/pentaho/di/shared/DelegatingSharedObjectsIO.java
abryant-hv: You are right that getSharedObject() does not need to be updated here, but the other implementing classes need to be updated, like VfsSharedObjectsIO. Currently that is case-sensitive. All the APIs on VfsSharedObjectsIO, MemorySharedObjectsIO, and RepositorySharedObjectsIO need to be updated to be case insensitive.

core/src/main/java/org/pentaho/di/connections/ConnectionManager.java
pdesai16: Isn't there another story https://hv-eng.atlassian.net/browse/BACKLOG-43064 that deals with metastore changes related to case insensitive checks

abryant-hv: Yes, but you've done a bunch of the work here. 
I don't recommend just ripping out what you've already done. 

Maybe add 
// TODO: BACKLOG-43064

To the relevant methods?

core/src/main/java/org/pentaho/di/shared/VfsSharedObjectsIO.java
```diff
@@ -365,4 +381,12 @@ protected String toString( Map<String, Node> nodesMap ) throws KettleXMLExceptio
     return xml.toString();
   }
 
+  private String findSharedObjectIgnoreCase( String name, Map<String, Node> nodeMap ) {
```
abryant-hv, line 384: Since we're copying this around, it could instead be added as a static method on the SharedObjectsIO interface.

core/src/test/java/org/pentaho/di/shared/DelegatingSharedObjectsIOTest.java
```diff
@@ -108,6 +109,33 @@ public void testDeleteDisabled() throws Exception {
     }
   }
 
+  /**
+   * Test to verify that getSharedObjects() returns the deduplicated list of items from
+   * different SharedObjectsIO
+   * @throws Exception
+   */
+  @Test
+  public void testGetSharedObjectsDeduplicated() throws Exception {
```
abryant-hv, line 118: You should also have a test for a case-insensitive getSharedObject(). testOverrides() above could have a test for getSharedObject( "B" ) and "A"

abryant-hv (approved):
Just a couple really minor things.

PR #9818, opened 2025-02-12
[BACKLOG-43082] - case insensitive for RepositorySharedObjectsIO

abryant-hv:


bconlon-hv (approved):

engine/src/main/java/org/pentaho/di/shared/RepositorySharedObjectsIO.java
```diff
@@ -113,39 +113,8 @@ public Map<String, Node> getSharedObjects( String type ) throws KettleException
...
-          if ( id != null ) {
-            object = repository.loadClusterSchema( id, slaveServerSupplier.get(), null );
-          }
-          break;
-    }
-    if ( object != null ) {
-      return object.toNode();
-    }
-    return null;
+    Map<String, Node> nodeMap = getSharedObjects( type );
```
pdesai16, line 116: This can cause performance issues

engine/src/main/java/org/pentaho/di/shared/RepositorySharedObjectsIO.java
```diff
@@ -183,25 +152,36 @@ public void clear( String type ) throws KettleException {
   @Override
   public void delete( String type, String name ) throws KettleException {
     SharedObjectType objectType = SharedObjectType.valueOf( type.toUpperCase() );
+
+    Map<String, Node> nodeMap = getSharedObjects( type );
```
pdesai16, line 156: Same comment as above

engine/src/main/java/org/pentaho/di/shared/RepositorySharedObjectsIO.java
```diff
@@ -211,8 +191,15 @@ public void delete( String type, String name ) throws KettleException {
 
   @Override
   public void saveSharedObject( String type, String name, Node node ) throws KettleException {
-    RepositoryElementInterface repoElement = null;
     SharedObjectType objectType = SharedObjectType.valueOf( type.toUpperCase() );
+
+    Map<String, Node> nodeMap = getSharedObjects( type );
```
pdesai16, line 196: This can cause performance issues

engine/src/main/java/org/pentaho/di/shared/RepositorySharedObjectsIO.java
abryant-hv: This should still all be in memory. I'm not sure how it could be improved and satisfy the acceptance criteria. 

See https://github.com/pentaho/pentaho-kettle/blob/master/plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/PurRepository.java#L1953 for where the caching happens that prevents additional network traffic.

PR #9829, opened 2025-02-21
[BACKLOG-42977] - convert global to local on save to repository

abryant-hv:
From spoon.

Also eliminate "setPrivateDatabases" which isn't used anymore.


tgf (approved):

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/JobDelegate.java
```diff
@@ -446,13 +422,6 @@ public DataNode elementToDataNode( final RepositoryElementInterface element ) th
 
     DataNode rootNode = new DataNode( NODE_JOB );
 
-    if ( jobMeta.getPrivateDatabases() != null ) {
-      // save all private database names http://jira.pentaho.com/browse/PPP-3413
```
pdesai16, line 450: We should make sure we are not regressing on the https://hv-eng.atlassian.net/browse/PPP-3413

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/TransDelegate.java
```diff
@@ -258,30 +258,6 @@ public void dataNodeToElement( final DataNode rootNode, final RepositoryElementI
     throws KettleException {
     TransMeta transMeta = (TransMeta) element;
 
-    Set<String> privateDatabases = null;
-    // read the private databases
-    DataNode privateDbsNode = rootNode.getNode( NODE_TRANS_PRIVATE_DATABASES );
-    // if we have node than we use one of two new formats. The older format that took
-    // too long to save, uses a separate node for each database name, the new format
-    // puts all the database names in the PROP_TRANS_PRIVATE_DATABASE_NAMES property.
-    // BACKLOG-6635
```
pdesai16, line 267: We should test our changes with BACKLOG-6635 to make sure we are not regressing

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/TransDelegate.java
```diff
@@ -607,13 +583,6 @@ public DataNode elementToDataNode( final RepositoryElementInterface element ) th
 
     DataNode rootNode = new DataNode( NODE_TRANS );
 
-    if ( transMeta.getPrivateDatabases() != null ) {
-      // save all private transformations database name http://jira.pentaho.com/browse/PPP-3405
```
pdesai16, line 611: Same comment as above. We need to test PPP-3405 to make sure we are not regressing

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/TransDelegate.java
abryant-hv: That issue doesn't exist in our jira. 

The entire concept of "private" databases isn't relevant anymore. There are only "in use" databases.

plugins/pur/core/src/main/java/org/pentaho/di/repository/pur/TransDelegate.java
abryant-hv: Same as above. Plus, what gets exported is being defined in BACKLOG-42840 and similar stories.

PR #9834, opened 2025-02-24
[BACKLOG-43080] - support case changes on SO import

abryant-hv:
- change SharedObjectsIO.findSharedObjectIgnoreCase to take a Set instead of a Map so it can be used in more cases.
- Support case renames in BaseSharedObjectManager
- correct RepositorySharedObjectsIO handling of renames (DON'T delete first)
- make sure import identifies name changes as changes.


pdesai16 (approved):

tgf (approved):
Wondering if it wouldn't be better to switch to maps with lowercase keys. The maps are now only serving getSharedObjects(type), where they never seem to be needed as maps anyway.

abryant-hv:
> Wondering if it wouldn't be better to switch to maps with lowercase keys. The maps are now only serving getSharedObjects(type), where they never seem to be needed as maps anyway.

Is `toLower()` consistent with `equalsIgnoreCase()` across locales? 
I'm also not sure, but we might need the real name to call down to the lower layers, in some cases. We could also get that from the existing object, I guess.

PR #9835, opened 2025-02-24
[BACKLOG-43259] File Open/Save dialog fails to open

pdesai16:
- If the .spoonrc file is referring to repolastdir and reposlastfile that does not exist in repository, the FileOpenSave dialog throws an IllegalAccessError which was not caught. Updated the catch to also  catch IllegalAccessError

### DIFF
--- a/plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/providers/recents/RecentFileProvider.java
+++ b/plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/providers/recents/RecentFileProvider.java
@@ -193,7 +193,7 @@ public class RecentFileProvider extends BaseFileProvider<RecentFile> {
           objectID = spoonInstance.rep.getJobId( lastUsedFile.getFilename(),
             spoonInstance.rep.findDirectory( lastUsedFile.getDirectory() ) );
         }
-      } catch ( KettleException e ) {
+      } catch ( KettleException | IllegalAccessError e ) {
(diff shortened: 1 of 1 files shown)


plugins/file-open-save-new/core/src/main/java/org/pentaho/di/plugins/fileopensave/providers/recents/RecentFileProvider.java
```diff
@@ -193,7 +193,7 @@ private void getLastUsedFile( List<LastUsedFile> lastUsedFiles, Spoon spoonInsta
           objectID = spoonInstance.rep.getJobId( lastUsedFile.getFilename(),
             spoonInstance.rep.findDirectory( lastUsedFile.getDirectory() ) );
         }
-      } catch ( KettleException e ) {
+      } catch ( Throwable e ) {
```
abryant-hv, line 196: Catching Throwable is a bad idea because it can also catch Errors, which should really never be caught. 

At the very least, this should change to something like `catch( KettleException | RuntimeException e )` (or IllegalArgumentException or InvocationTargetException, etc.) so we don't catch Errors. 

Better yet, though, would be to have the Repository code catch the RuntimeException and convert it to a checked (probably KettleException) at that layer.

abryant-hv (changes requested):

bconlon-hv (commented):
Update the commit message - it no longer describes the change

abryant-hv (approved):

bconlon-hv (approved):

PR #9861, opened 2025-03-13
[BACKLOG-43080] - propagate objectid to imported partition schemas

abryant-hv:


bconlon-hv (approved):

engine/src/main/java/org/pentaho/di/repository/RepositoryImporter.java
```diff
@@ -668,6 +677,7 @@ protected boolean importTransformation( Node transnode, RepositoryImportFeedback
       transMeta.setRepositoryDirectory( targetDirectory );
       patchTransSteps( transMeta );
       patchDatabaseConnections( transMeta );
+      patchPartitionSchemas( transMeta );
```
pdesai16, line 680: Why was this not needed earlier?

bconlon-hv: import.sh hits an NPE without it.  See https://hv-eng.atlassian.net/browse/BACKLOG-43562

abryant-hv: The root cause goes back to our SharedObjectsIO, and following better code practices. It used to be that all the references to these shared object java objects all referred to the same object... shared state. We instead now follow best-practices with defensive copies, etc. This leaves a gap where old code assumed that writes to one instance of an object (effectively) updated that object wherever it was used. 

In the case of DatabaseConnections, every step, plus the Transformations and Jobs themselves, have a reference to a DatabaseMeta that they loaded at load time. When other changes to those objects happen, we have to explicitly propagate those changes to those objects, hence patchDatabaseConnections, and the SharedObjectUtil method that it calls. 

In the case of import, we're specifically worried about ObjectIds. In the process of copying from Local to Global, each shared object gets an ObjectId. patchDatabaseConnections was propagating those ObjectIds back into the steps. This lets the code that writes the steps to the repository correctly link to the database in the repository. 

Bryan found a similar case for PartitionSchemas. Basically the same problem. I hadn't hit it because I didn't test thoroughly enough. The solution is the same: propagate changes back to the objects embedded in the steps. If we end up needing this in more places, the code could also move to SharedObjectUtil, but we may not need it.

pdesai16: So this case only handles partitionSchema. Do other sharedObjects like clusterSchema needs to be handled the same way

pdesai16 (approved):

PR #9866, opened 2025-03-17
[BACKLOG-43421] Rename "Global" to "System" and "Local" to "File"

pdesai16:
These changes update the level name suffix as it appears in right hand navigation tree in Spoon.
- Added method in Bowl API to return the level name
- LeveledTreeNode uses the numeric display order for sorting instead of level name


core/src/main/java/org/pentaho/di/core/bowl/BaseBowl.java
```diff
@@ -61,5 +61,10 @@ public void addParentBowl( Bowl parent ) {
   public synchronized void clearManagers() {
     managerInstances.clear();
   }
+
+  @Override
+  public String getLevelDisplayName() {
+    return "";
+  }
```
abryant-hv, line 68: Why have this? It would probably be better to force concrete instances to pick something.

core/src/main/java/org/pentaho/di/core/bowl/DefaultBowl.java
```diff
@@ -31,6 +33,8 @@
  *
  */
 public class DefaultBowl extends BaseBowl {
+  private static Class<?> PKG = Const.class; // for i18n purposes, needed by Translator2!!
```
abryant-hv, line 36: Shouldn't this be DefaultBowl.class?

plugins/connections/ui/src/main/resources/org/pentaho/di/connections/ui/tree/messages/messages_en_US.properties
```diff
@@ -3,9 +3,9 @@ VFSConnectionsTree.Title=VFS Connections
 VFSConnectionPopupMenuExtension.MenuItem.New=New...
 VFSConnectionPopupMenuExtension.MenuItem.Edit=Edit...
 VFSConnectionPopupMenuExtension.MenuItem.Delete=Delete
-VFSConnectionPopupMenuExtension.MenuItem.CopyToGlobal=Copy to Global
+VFSConnectionPopupMenuExtension.MenuItem.CopyToGlobal=Copy to {0}
 VFSConnectionPopupMenuExtension.MenuItem.CopyToProject=Copy to Project
```
abryant-hv, line 7: "Project" is still hard-coded here. Shouldn't this use the Bowl LevelDisplayName?

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/ui/RunConfigurationPopupMenuExtension.java
```diff
@@ -130,7 +131,8 @@ private Menu createItemPopupMenu( Tree tree ) {
       } );
 
       MenuItem copyMenuItem = new MenuItem( itemMenu, SWT.NONE );
-      copyMenuItem.setText( BaseMessages.getString( PKG, "RunConfigurationPopupMenuExtension.MenuItem.CopyToGlobal" ) );
+      copyMenuItem.setText( BaseMessages.getString( PKG, "RunConfigurationPopupMenuExtension.MenuItem.CopyTo",
+              spoonSupplier.get().getGlobalManagementBowl().getLevelDisplayName() ) );
```
abryant-hv, line 135: This is great, but why not do the same at lines 150 and 161 for Projects?

plugins/engine-configuration/impl/src/main/resources/org/pentaho/di/engine/ui/messages/messages_en_US.properties
```diff
@@ -9,12 +9,12 @@ RunConfigurationDialog.Group.Settings=Settings
 RunConfigurationPopupMenuExtension.MenuItem.New=New...
 RunConfigurationPopupMenuExtension.MenuItem.Edit=Edit...
 RunConfigurationPopupMenuExtension.MenuItem.Delete=Delete
-RunConfigurationPopupMenuExtension.MenuItem.CopyToGlobal=Copy to Global
+RunConfigurationPopupMenuExtension.MenuItem.CopyTo=Copy to {0}
 RunConfigurationPopupMenuExtension.MenuItem.CopyToProject=Copy to Project
```
abryant-hv, line 13: Still hard-coded to Project.

ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -3301,12 +3298,16 @@ private void setDatabaseMenuItems( XulDomContainer mainSpoonContainer, SpoonTree
...
+      if ( leveledSelection.getLevel() == LeveledTreeNode.LEVEL.FILE ) {
         if ( getBowl() == getGlobalManagementBowl() ) {
           moveGlobalItem.setVisible( true );
+          moveGlobalItem.setLabel( BaseMessages.getString( PKG, "Spoon.Menu.Popup.MoveTo",
+                  globalManagementBowl.getLevelDisplayName() ) );
           moveProjectItem.setVisible( false );
         } else {
           moveGlobalItem.setVisible( true );
+          moveGlobalItem.setLabel( BaseMessages.getString( PKG, "Spoon.Menu.Popup.MoveTo",
+                                                                    globalManagementBowl.getLevelDisplayName() ) );
```
abryant-hv, line 3310: weird indent

core/src/main/java/org/pentaho/di/core/bowl/Bowl.java
```diff
@@ -88,4 +88,9 @@ public interface Bowl {
    */
   VariableSpace getADefaultVariableSpace();
 
+  /**
+   * Return the display name of the level.
```
abryant-hv, line 92: We should encourage the implementation to use i18n

ui/src/main/resources/org/pentaho/di/ui/spoon/messages/messages_en_US.properties
```diff
@@ -246,12 +243,15 @@ ExportCmdLineShFiles.ConfirmOverwrite=Confirm?
 TransGraph.PopupMenu.Clustering=&Clusters...
 Spoon.Menu.Transformation.Verify=&Verify...
 Spoon.Log.ElementHasNoParent=Element has no parent
+Spoon.Menu.Popup.CONNECTIONS.DuplicateInProject=Duplicate in Project...
+Spoon.Menu.Popup.DuplicateIn=Duplicate in {0}...
```
abryant-hv, line 247: Why still have the Project here? just use the one below it, and get the name from the Bowl.

abryant-hv (changes requested):

core/src/main/java/org/pentaho/di/core/bowl/BaseBowl.java
pdesai16: I wasn't sure if all implementation of Bowl will have a level name that gets displayed on the tree. Therefore added a method in BaseBowl

core/src/main/java/org/pentaho/di/core/bowl/DefaultBowl.java
pdesai16: Bowl package does not have any resource file so I didn't want to create a resource file with one entry. I reused the one at the parent level

plugins/connections/ui/src/main/resources/org/pentaho/di/connections/ui/tree/messages/messages_en_US.properties
pdesai16: My impression was only the Global needs to be changed as it is dynamic. The Project related menu items are not changing as left it as is

abryant-hv: This PR adds some places that use the Level Display Name from the Project bowl. It would be good to just plumb that through everywhere. 

A straw man idea would be that we could change the term "Project" to something else by only changing the plugin. Not that we *would* do this, but it's a litmus test.

ui/src/main/java/org/pentaho/di/ui/core/widget/tree/LeveledTreeNode.java
```diff
@@ -21,29 +21,34 @@
...
+
+    public int getDisplayOrder() {
+      return displayOrder;
     }
   }
 
   private static final String NAME_KEY = "name";
   private static final String LEVEL_KEY = "level";
+  public static final String LEVEL_FILE_DISPLAY_NAME = "File";
+  public static final String LEVEL_DEFAULT_DISPLAY_NAME = "Default";
```
abryant-hv, line 45: Oh, and this too. Both "File" and "Default" need to be internationalized.

core/src/main/java/org/pentaho/di/core/bowl/BaseBowl.java
abryant-hv: I guess this is fine. You could also have just added it as a default method in the interface to start with.

plugins/connections/ui/src/main/resources/org/pentaho/di/connections/ui/tree/messages/messages_en_US.properties
pdesai16: This can be implemented but was not covered in the acceptance criteria for this story

abryant-hv (approved):

PR #9868, opened 2025-03-18
Db fixes

varuntangirala:


core/src/main/java/org/pentaho/di/core/util/Utils.java
```diff
@@ -299,4 +302,20 @@ public static short[][] normalizeArrays( int normalizeToLength, short[]... array
     return rtn;
   }
 
+  public static String setStringValueFromMap( Map<String, String> map, String key ) {
+    if ( Objects.nonNull( map.get( key ) ) ) {
```
sVarma1992, line 306: This code can be simplified using getOrDefault

sVarma1992 (approved):

PR #9889, opened 2025-03-24
[BACKLOG-43471] - refresh db connections when connecting/disconnectin…

abryant-hv:
…g from the repository.

Allows execution to use the correct environment when an open file is run after connect/disconnect

### DIFF
--- a/engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
+++ b/engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
@@ -431,6 +431,8 @@ public abstract class AbstractMeta implements ChangedFlagInterface, UndoInterfac
   public void setBowl( Bowl bowl ) {
     this.bowl = Objects.requireNonNull( bowl );
     initializeNonLocalSharedObjects();
+    // now that the bowl has changed, make sure we have up-to-date DatabaseMetas.
+    allDatabasesUpdated();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
```diff
@@ -431,6 +431,10 @@ public Bowl getBowl() {
   public void setBowl( Bowl bowl ) {
     this.bowl = Objects.requireNonNull( bowl );
     initializeNonLocalSharedObjects();
+    // now that the bowl has changed, make sure we have up-to-date DatabaseMetas.
+    for ( String name : getUsedDatabaseConnectionNames() ) {
```
pdesai16, line 435: The getUsedDatabaseConnectionNames() is getting he dbnames from steps. but the databaseUpdated() is again going thru the steps.getUsedDatabaseConnectionNames() and looping thru dbs. It looks like we are doing it twice

abryant-hv: True. This could be more efficient. I'll add a new method for that.

abryant-hv: Refactored.

engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
```diff
@@ -1295,7 +1308,13 @@ public void databaseUpdated( String name ) {
 
   // helper method for databasesUpdated()
   protected void updateFields( DatabaseMeta existing, Optional<DatabaseMeta> newDb ) {
-    newDb.ifPresentOrElse( db -> existing.replaceMeta( db ), () -> existing.replaceMeta( new DatabaseMeta() ) );
+    newDb.ifPresentOrElse( db -> existing.replaceMeta( db ),
+      // remember the name.
+      () -> {
+        DatabaseMeta newMeta = new DatabaseMeta();
+        newMeta.setName( name );
```
tgf, line 1315: is this the right name?

abryant-hv: Fixed.

tgf (approved):

pdesai16 (approved):

PR #9961, opened 2025-04-17
[BACKLOG-42786] Update the RunConfigurationManager to accept a regist…

singletonc:
…ration of a new configuration type from a plugin

- Refactored engine-configuration plugin to have separate api and impl modules so that the api jar can be dropped into kettle/lib, thus allowing other plugins (like spark-execution) to use it -- reorganized dependencies with this in mind
- Created Factory and FactoryManager classes to: -- allow for plugins to register their Provider types via Factories (e.g. Spark/AEL) through ServicePlugin framework -- address future changes for Project Profile in which there will be one RunConfigurationManager and metastore per project
- Removed defaultRunConfigurationProvider parameter and logic from RunConfigurationManager as this should be handled in the constructor logic and should not ever be null
- Created new test class for new class RunConfigurationProviderFactoryManager
- Moved default constructor logic from Providers up to RunConfigurationManager() so that we will have the metastoreSupplier to pass to RunConfigurationProviderFactoryManager.getProviders()
- Updated and fixed tests as best I could, struggling with the RunConfigurationProviderFactoryManagerImpl.testPluginServiceLoaded(), it works when I debug, but not when I just run it... leaving TODO for now to get draft in for review


plugins/engine-configuration/api/src/main/java/org/pentaho/di/engine/configuration/api/MetaStoreRunConfigurationFactory.java
```diff
@@ -30,10 +32,22 @@
  */
 public abstract class MetaStoreRunConfigurationFactory implements RunConfigurationFactory {
 
+  Logger logger = LoggerFactory.getLogger( MetaStoreRunConfigurationFactory.class );
+
   protected CheckedMetaStoreSupplier metastoreSupplier;
 
   public MetaStoreRunConfigurationFactory( CheckedMetaStoreSupplier metastoreSupplier ) {
-    this.metastoreSupplier = metastoreSupplier;
+    if( metastoreSupplier == null){
```
abryant-hv, line 40: "if(" -> "if ("

plugins/engine-configuration/api/src/main/java/org/pentaho/di/engine/configuration/api/MetaStoreRunConfigurationFactory.java
```diff
@@ -30,10 +32,22 @@
...
+      try {
+        Collection<MetastoreLocator> metastoreLocators = PluginServiceLoader.loadServices( MetastoreLocator.class );
+        MetastoreLocator metastoreLocator = metastoreLocators.stream().findFirst().get();
+        this.metastoreSupplier = metastoreLocator::getMetastore;
+      } catch ( Exception e ) {
+        logger.warn( "Error getting MetastoreLocator", e );
+      }
```
abryant-hv, line 47: This should just be a call to MetastoreConst.getDefaultMetastoreSupplier() or something similar. We don't want the logic for loading MetastoreLocator plugins, or caching them correctly, spread all over the place. 

This also isn't doing any caching at all, which means this will iterate all the plugins every time it is called. Expensive(ish).

PR #9966, opened 2025-04-23
[BACKLOG-42786] Update the RunConfigurationManager to accept a registration of a new configuration type from a plugin

singletonc:
- Refactored engine-configuration plugin to have separate api and impl modules so that the api jar can be dropped into kettle/lib, thus allowing other plugins (like spark-execution) to use it
-- reorganized dependencies with this in mind
- Created Factory and FactoryManager classes to:
-- allow for plugins to register their Provider types via Factories (e.g. Spark/AEL) through ServicePlugin framework
-- address future changes for Project Profile in which there will be one RunConfigurationManager and metastore per project
- Removed defaultRunConfigurationProvider parameter and logic from RunConfigurationManager as this should be handled in the constructor logic and should not ever be null
- Created new test class for new class RunConfigurationProviderFactoryManager
- Moved null-constructor, null-checkedMetastoreSupplier logic from Providers abstract parent class MetaStoreRunConfigurationFactory so that logic doesn't have to be copied to each RunConfigurationProvider
- Created/updated unit tests

Includes:
[BACKLOG-42787] - Move all the AEL code added in pentaho-kettle to a new enterprise plugin

[BACKLOG-42787]: https://hv-eng.atlassian.net/browse/BACKLOG-42787?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

and 

[BACKLOG-42785](https://hv-eng.atlassian.net/browse/BACKLOG-42785) - Allow PDI core system to register a new execution engine through a plugin

- Created Factory and FactoryManager classes to:
-- allow plugins to register TransFactory for respective Run Configuration and create either a Trans Object(For default Run Config) or a TransWebSocketEngineAdapter object (for Spark Run Config)
- Refactored the creation of Trans objects in JobEntryTrans and TransGraph using the TransFactory
- Created/updated unit tests
- Create new test class for DefaultTransFactoryManager


plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImpl.java
```diff
@@ -0,0 +1,79 @@
...
+ * Since RunConfigurationProviders are tied to metastores, and metastores are tied to Projects, we need to be able to
+ * create (via Factories) instances of these Providers for each RunConfigurationManager instance associated with each
+ * Project.
+ */
+public class RunConfigurationProviderFactoryManagerImpl implements RunConfigurationProviderFactoryManager {
+  private static RunConfigurationProviderFactoryManagerImpl instance;
+
+  private ArrayList<RunConfigurationProviderFactory> factories;
+
+  Logger logger = LoggerFactory.getLogger( RunConfigurationProviderFactoryManagerImpl.class );
```
abryant-hv, line 43: private static?

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImpl.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public static RunConfigurationProviderFactoryManagerImpl getInstance() {
+    if ( null == instance ) {
+      instance = new RunConfigurationProviderFactoryManagerImpl();
+    }
+    return instance;
+  }
```
abryant-hv, line 50: needs to be synchronized.

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImpl.java
```diff
@@ -0,0 +1,79 @@
...
+
+  public RunConfigurationProviderFactoryManagerImpl() {
```
abryant-hv, line 52: Do we want this public? Making it private will enforce that the class is a singleton.

abryant-hv (approved):

engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
```diff
@@ -403,6 +406,14 @@ public void setBowl( Bowl bowl ) {
     this.bowl = bowl;
   }
 
+  public TransSupplier getTransSupplier() {
```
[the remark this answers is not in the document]
singletonc: That is a good point. I didn't make that change, and the original author was perhaps still working on it, but is out this week, so I am committing it as I need to put my code on top. I can look deeper into options to refactor this.

Right now, I think it is because execute in the RunConfigurationExecutor implementations have an AbstractMeta [in the method signature](https://github.com/pentaho/pentaho-kettle/blob/df160505c880a24e6f1cfeda115fad41a86ae113/plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/api/RunConfigurationExecutor.java#L26)

I will see what I can do to improve this...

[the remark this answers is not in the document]
singletonc: we are already using instanceof in those overrides, I will add
`    if( meta instanceof TransMeta) {
      ((TransMeta) meta).setTransSupplier( new DefaultTransSupplier() );
    }`

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImpl.java
singletonc: I am using it to get the test class to work properly, I can at least make it protected, lets see if I can make it private... I could add a protected "clear()" method so that it can be reinitialized in the test class?

abryant-hv: I've seen both used. If you have a test that doesn't *directly* call this, but instead calls something that uses the singleton, clear() can be more useful.

plugins/engine-configuration/api/src/main/java/org/pentaho/di/engine/configuration/api/MetaStoreRunConfigurationFactory.java
```diff
@@ -30,10 +32,22 @@
  */
 public abstract class MetaStoreRunConfigurationFactory implements RunConfigurationFactory {
 
+  Logger logger = LoggerFactory.getLogger( MetaStoreRunConfigurationFactory.class );
+
   protected CheckedMetaStoreSupplier metastoreSupplier;
 
   public MetaStoreRunConfigurationFactory( CheckedMetaStoreSupplier metastoreSupplier ) {
-    this.metastoreSupplier = metastoreSupplier;
+    if( metastoreSupplier == null){
```
NJtwentyone, line 40: why are we handling a null metastoreSupplier here?

is this related to `RunConfigurationProviderFactoryManagerImpl.java`
```
 public List<RunConfigurationProvider> generateProviders() {
    return generateProviders( null );
  }
```
And then why we only selecting the first one returned from a list and then we assume there is at least one and access a potential null obhect.

plugins/engine-configuration/impl/src/test/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public class RunConfigurationProviderFactoryManagerImplTest {
+
+  @Test
+  public void testDefaultProvidedMetastore() {
+    MemoryMetaStore memoryMetaStore = new MemoryMetaStore();
+    CheckedMetaStoreSupplier metastoreSupplier = () -> memoryMetaStore;
```
NJtwentyone, line 35: Not a blocker, but think MemoryMetaStorejava is going away or not really used. But for a unit test, you should try and not rely on a class not under test to provide functionality.

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImpl.java
NJtwentyone: if the goal is to create a singleton then the constructor(s) have to private - https://www.geeksforgeeks.org/private-constructors-and-singleton-classes-in-java/

NJtwentyone: to piggy back off Alan, you want to manipulate the factory variable

- make the a getter/setter proctected, package-private
- @VisibleForTesting to a getter/setter
- - https://github.com/pentaho/pentaho-kettle/blob/ac91908ea2ecab2b2f911c75d13f30303b740e76/engine/src/main/java/org/pentaho/di/kitchen/KitchenCommandExecutor.java#L389

plugins/engine-configuration/impl/src/test/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+    List<RunConfigurationProvider> providers =
+      new RunConfigurationProviderFactoryManagerImpl().generateProviders( metastoreSupplier );
+
+    Assert.assertEquals( 1, providers.size() );
```
NJtwentyone, line 39: if you are just trying to test the logic for `RunConfigurationProviderFactoryManagerImpl#generateProviders(checkedMetaStoreSupplier )` then you can create a getter/setter function for factory variable - https://github.com/pentaho/pentaho-kettle/pull/9966/files#r2056832326

and test that generic FactoryManagers work.

For unit tests, the goal is to test only the class under Test, RunConfigurationProviderFactoryManagerImpl.java and not neccesarily DefaultRunConfigurationProviderFactory#getProvider returns a correct result

plugins/engine-configuration/impl/src/test/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImplTest.java
```diff
@@ -0,0 +1,66 @@
...
+    try ( MockedStatic<PluginServiceLoader> pluginServiceLoaderMockedStatic = Mockito.mockStatic(
+      PluginServiceLoader.class ) ) {
+      pluginServiceLoaderMockedStatic.when(
+        () -> PluginServiceLoader.loadServices( RunConfigurationProviderFactory.class ) ).thenReturn( factories );
+
+      List<RunConfigurationProvider> providers =
+        new RunConfigurationProviderFactoryManagerImpl().generateProviders( metastoreSupplier );
+
+      Assert.assertEquals( 2, providers.size() );
+      Assert.assertEquals( "Pentaho", providers.get( 0 ).getType() );
```
NJtwentyone, line 60: This is creating a brittle test.

I would create Mock class for RunConfigurationProviderFactory.class and return a type "GenericProvider1" and test that. The order shouldn't matter, wether it's first or last or somewhere in the middle of the return List<RunConfigurationProvider>

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImpl.java
```diff
@@ -0,0 +1,79 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * The RunConfigurationProviderFactoryManager is used to manage the registration of RunConfiguration types and
+ * generate their RunConfigurationProviders for RunConfigurationManager instances associated with Projects
+ * <p>
+ * Each RunConfiguration type (e.g. Pentaho, Spark) has a RunConfigurationProvider that allows users to create
```
NJtwentyone, line 32: I wouldn't explicitly mention "Spark", this is CE code and spark is not a concept or plugin in this domain. Who knows what the future could hold, "Spark" might not even be an option.

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImpl.java
```diff
@@ -0,0 +1,79 @@
...
+
+    try {
+      factories.addAll( PluginServiceLoader.loadServices( RunConfigurationProviderFactory.class ) );
+    } catch ( KettlePluginException e ) {
+      logger.warn( "Error loading plugin RunConfigurationProviderFactory(s)", e );
+    }
+  }
+
+  public List<RunConfigurationProvider> generateProviders() {
+    return generateProviders( null );
```
NJtwentyone, line 64: why are passing null into RunConfigurationProviderFactoryManagerImp ?#generateProviders ?

is it related to this - https://github.com/pentaho/pentaho-kettle/pull/9966/files#r2056812071

engine/src/main/java/org/pentaho/di/base/AbstractMeta.java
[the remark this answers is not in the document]
NJtwentyone: TransSupplier/java, is just a stop-gap for our feature branch

@abryant-hv  we are working on using a factory to return the Trans.java apart of spike - https://hv-eng.atlassian.net/browse/BACKLOG-42709. I can't find the story that will implement it but

plugins/engine-configuration/api/src/main/java/org/pentaho/di/engine/configuration/api/MetaStoreRunConfigurationFactory.java
singletonc: I had moved the logic from [DefaultRunConfigurationProvider()](https://github.com/pentaho/pentaho-kettle/pull/9966/files#diff-7f75f5396784eea66235a48945302208615d1cce3d6682ca14d13354dbd0919bL60) and [SparkRunConfigurationProvider()](https://github.com/pentaho/spark-execution/pull/2/commits/8120ba6bd7d01e6012743f21e17007a4b786acf8#diff-e4d7d4ec0bea4a5fc4c055e38ce3fb0f010a75f6ddc57f3b5a6b8f5bcd7d9492L55) up to RunConfigurationProviderFactoryManagerImpl, but then decided to move it to their parent class here. I can move it back to RunConfigurationProviderFactoryManagerImpl if that makes more sense.

plugins/engine-configuration/impl/src/test/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImplTest.java
singletonc: I will be updating this to [MetastoreConst as Alan pointed out](https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2056705325)

I will update this comment with the link to the change...

plugins/engine-configuration/impl/src/main/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImpl.java
singletonc: yes, this logic is all related and comes from preexisting logic that I have moved around and hadn't modified until Alan's suggestion. See the comment thread you linked for more information.

plugins/engine-configuration/impl/src/test/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImplTest.java
singletonc: I will centralize related discussion [in this thread](https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2056812071)

plugins/engine-configuration/api/src/main/java/org/pentaho/di/engine/configuration/api/MetaStoreRunConfigurationFactory.java
```diff
@@ -30,10 +32,22 @@
...
+      try {
+        Collection<MetastoreLocator> metastoreLocators = PluginServiceLoader.loadServices( MetastoreLocator.class );
+        MetastoreLocator metastoreLocator = metastoreLocators.stream().findFirst().get();
+        this.metastoreSupplier = metastoreLocator::getMetastore;
+      } catch ( Exception e ) {
+        logger.warn( "Error getting MetastoreLocator", e );
+      }
+    } else {
+      this.metastoreSupplier = metastoreSupplier;
```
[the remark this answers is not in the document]
singletonc: I will be updating this to `if ( metastoreSupplier == null ) { this.metastoreSupplier = MetaStoreConst::getDefaultMetastore;` 

Is that ok? There is getDefaultMetastoreSupplier(), but then it seems I have to cast the result to CheckedMetaStoreSupplier, vs just using the method reference operator

plugins/engine-configuration/impl/src/test/java/org/pentaho/di/engine/configuration/impl/RunConfigurationProviderFactoryManagerImplTest.java
singletonc: ok, ty, I wanted to get into the ServiceProvider piece of this, seems I over-complicated!

plugins/engine-configuration/api/src/main/java/org/pentaho/di/engine/configuration/api/MetaStoreRunConfigurationFactory.java
singletonc: So, through multiple fixes based on the review comments, I could change this to:
`return generateProviders( MetaStoreConst::getDefaultMetastore );`

In fact... that's a one liner, I can just do it. I had, as stated, moved that logic here because it's the abstract parent of the Providers, and the logic had been in the child classes, but, it seems that wasn't really the best place for it to begin with, so, I'll make that change now...

I am making all of these changes in grouped commits, I can squash it all after reviews are in if we don't want to "Squash and merge" on GitHub (Duarte had mentioned some reason as to why this can be a problem, I don't recall)

NJtwentyone: My stance, the RunConfigurationManger and related classes should never be managing the MetaStore it should be passed to us and calling the supplier #get().

`MetaStoreConst::getDefaultMetastore` is a better approach than passing` null`.

I'd have to look at the code and flow of logic. to have a better recommendation than `MetaStoreConst::getDefaultMetastore`

NJtwentyone: @singletonc  is there ever a situation where you won't have access to bowl or metastore either (abstractMeta, jobMeta, TransMeta) ?

I haven't done an exhaustive looks, But is the useds of RunConfigurationManager#getInstance()

- RunConfigurationImportExtensionPoint.java,  
 - you could access the RunConfigurationManager#getInstance(bowl) in side #callExtensionPoint(logChannelInterface, o)
- RunConfigurationSaveExtensionPoint.java
  - same thing as above
- RunConfigurationDelegate.java, (I assume UI selection)
 - #edit  calls  RunConfigurationManager logic
 - but another function #updateLoadedJobs has `spoonSupplier.get().getLoadedJobs()` , I assume we could get the bowl/metastore from this call or if there is a function related to "current focus job/trans"

singletonc: In conversation with @rmansoor, he thinks that if:
- what we have currently *works* with the code we currently have
- there is _not_ something drastically wrong with this approach

Then we ought to proceed with the current changes (that I have updated to this point based on review feedback) and we can refactor further as we approach the point at which we will be merging spark-execution into master. That then also answers the question we discussed today @NJtwentyone: our team will be doing this refactoring as project profile will be merging before us, we will rebase on them.

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.function.Supplier;
+
+public class DefaultTransManager implements TransMananger {
+
+  private static final DefaultTransManager instance = new DefaultTransManager();
+
+  Map<String, TransFactory> mapConfigurationTypeTransFactory;
+
+
+  public DefaultTransManager(){
```
NJtwentyone, line 17: Sonar will find flag the implementation of this singleton `DefaultTransManager,` in short it needs to be thread safe.

https://rules.sonarsource.com/java/tag/design/RSPEC-6548/

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
```diff
@@ -0,0 +1,43 @@
+package org.pentaho.di.trans;
+
+import org.pentaho.di.core.exception.KettleException;
+import org.pentaho.di.core.logging.LogChannelInterface;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Supplier;
+
+public class DefaultTransManager implements TransMananger {
```
NJtwentyone, line 10: javadoc class comment for DefaultTransManager

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  public static synchronized DefaultTransManager getInstance() {
+    if ( null == instance ) {
+      instance = new DefaultTransManager();
+    }
+    return instance;
+  }
+
+  @Override
```
NJtwentyone, line 28: javadoc comments for DefaultTransManager#createTrans and other functions

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
```diff
@@ -0,0 +1,43 @@
...
+  public Trans createTrans( TransMeta transMeta, LogChannelInterface log, String runConfiguration,
+    Supplier<Trans> fallbackSupplier ) throws KettleException {
+    TransFactory transFactory = getTransFactory( runConfiguration );
+    return transFactory.create( transMeta, log, fallbackSupplier );
```
NJtwentyone, line 32: We are having am offline discussion on use use of `fallbackSupplier.` IMHO i think the use of `fallbacksupplier` is unnecessary design approach, back when the previous team did have ael as true plugin and ael was packaged along with the monolith pdi application.

We should assume that all the engine plugins have been registered and return null or throw an exception.

If we want to have some type of "default", then it should be hard coded in the  DefaultTransManager construction Or (brainstorming) engine plugins can signify that they will be the default, then you have to have sort of "tiebreaker" or fail it two plugins make themselves the "default"

engine/src/main/java/org/pentaho/di/trans/TransFactory.java
```diff
@@ -0,0 +1,11 @@
+package org.pentaho.di.trans;
+
+import org.pentaho.di.core.exception.KettleException;
+import org.pentaho.di.core.logging.LogChannelInterface;
+
+import java.util.function.Supplier;
+
+public interface TransFactory {
```
NJtwentyone, line 8: javadoc class comment TransFactory

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
```diff
@@ -0,0 +1,43 @@
+package org.pentaho.di.trans;
```
NJtwentyone, line 1: All new files do need a source code header per Ramaiz - https://hitachivantara-eng.slack.com/archives/C0819BSJK/p1738251537401139?thread_ts=1736883936.812999&cid=C0819BSJK

the thing after the header is there, devops will manage them

engine/src/main/java/org/pentaho/di/trans/TransMananger.java
```diff
@@ -0,0 +1,15 @@
+package org.pentaho.di.trans;
+
+import org.pentaho.di.core.exception.KettleException;
+import org.pentaho.di.core.logging.LogChannelInterface;
+
+import java.util.function.Supplier;
+
+public interface TransMananger  {
```
NJtwentyone, line 8: javadoc class comment for TransManager

engine/src/main/java/org/pentaho/di/trans/TransMananger.java
```diff
@@ -0,0 +1,15 @@
...
+  Trans createTrans( TransMeta transMeta, LogChannelInterface log, String runConfiguration,
```
NJtwentyone, line 9: javadoc class comment for TransManager#create and all functions

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public static synchronized DefaultTransManager getInstance() {
+    if ( null == instance ) {
+      instance = new DefaultTransManager();
+    }
+    return instance;
+  }
+
+  @Override
+  public Trans createTrans( TransMeta transMeta, LogChannelInterface log, String runConfiguration,
```
NJtwentyone, line 29: no unit tests for DefaultTransManager.java and other classes that have implementations.

engine/src/main/java/org/pentaho/di/job/entries/trans/JobEntryTrans.java
```diff
@@ -1124,8 +1124,8 @@ public Result execute( Result result, int nr ) throws KettleException {
           //trans = new Trans( transMeta, this );
           final TransMeta meta = transMeta;
 
-          trans = transMeta.getTransSupplier().get( transMeta, log, () -> new Trans( meta, this ) );
-
+          trans = DefaultTransManager.getInstance().createTrans( transMeta, log,
```
NJtwentyone, line 1127: I think we are missing the point of having a factory pattern.

The point is to have the instantiation "new Trans(...)" and "new XYZTrans(...)" handled inside the factory and you just pass in only the arguments and/or the "selector" ie run configuration. 

Below are some clean examples:
- https://refactoring.guru/design-patterns/factory-method
- https://refactoring.guru/design-patterns/factory-method/java/example#example-0

The current proposed code changes to use the new factory pattern

```DefaultTransManager.getInstance().createTrans( transMeta, log,
executionConfiguration.getRunConfiguration(), () -> new Trans( meta, this ) );```
and
```DefaultTransManager.getInstance().createTrans( transMeta, log,
            executionConfiguration.getRunConfiguration(), this::createLegacyTrans );```
doesn't have a default Trans Factory which calls the pentaho engine  code such as new Trans which is hidden in the "fallbackSupplier". I voiced my concerns about that too - https://github.com/pentaho/pentaho-kettle/pull/9966/commits/50e4e551c96a5479288ba79e1ebb32d86212d526#r2069456995

We need to support to support the multiple Trans constructors

- `Trans( TransMeta transMeta )`
- `Trans( TransMeta transMeta, LoggingObjectInterface parent )`
- `Trans( Parent parent, Repository rep, String name, String dirname, String filename )`
- `Trans( Parent parent, Repository rep, String name, String dirname, String filename, TransMeta parentTransMeta )`

I think I have them all listed. I think for spark-execution scenario we are immediate supporting, we only need to support two.

You can address this two ways

1.  Overload the TransManager interface
DefaultTransManager.getInstance().createTrans(executionConfiguration.getRunConfiguration(), transMeta);
DefaultTransManager.getInstance().createTrans(executionConfiguration.getRunConfiguration(), transMeta, log);
etc...
2. TransManager#getFactory method + overload TransFactory interface DefaultTransManager.getInstance().getFactory(executionConfiguration.getRunConfiguration()).createTrans(transMeta);
DefaultTransManager.getInstance().getFactory(executionConfiguration.getRunConfiguration()).createTrans(transMeta, log)

You already have TransManager#getTransFactory and I think option No. 2 is a better approach, more flexible and extensible. 

But you need to solve the problem of registering the "default TransFactory" that actually calls `new Trans(...)`

File: ui/src/main/java/org/pentaho/di/ui/spoon/trans/TransGraph.java [hunk 22]
```diff
@@ -3861,8 +3861,8 @@ public synchronized void start( TransExecutionConfiguration executionConfigurati
           // memory
           // To be able to completely test this, we need to run it as we would normally do in pan
           //
-          trans = transMeta.getTransSupplier().get( transMeta, log, this::createLegacyTrans );
-
+          trans = DefaultTransManager.getInstance().createTrans( transMeta, log,
+            executionConfiguration.getRunConfiguration(), this::createLegacyTrans );
```
NJtwentyone, line 3865: see comment https://github.com/pentaho/pentaho-kettle/pull/9966/commits/50e4e551c96a5479288ba79e1ebb32d86212d526#r2069538322 for use of `this::createLegacyTrans`

for context `TransGraph#createLegacyTrans`

```
  private Trans createLegacyTrans() {
    try {
      return new Trans( transMeta, spoon.rep, transMeta.getName(),
        transMeta.getRepositoryDirectory().getPath(),
        transMeta.getFilename(), transMeta);
    } catch ( KettleException e ) {
      throw new RuntimeException( e );
    }
  }
```

NJtwentyone (changes requested):
@singletonc PR was good to go.
Then @soagarwal1 added his changes which were apart of separate jira issue.
So these request changes are strictly for @soagarwal1. 

_**I suggest we keep the two efforts separate in separate peer reviews.**_

The major items are:

- Singleton thread safe issue - https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2068880243
- Factory approach - https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2069538322
- Use of fallbacksupplier - https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2069456995
- Tests for new code (depends on above) - https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2069472116
- Headers - https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2069459663
- javadoc comments

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/trans/TransFactory.java
```diff
@@ -0,0 +1,11 @@
...
+
+  Trans create( TransMeta transMeta, LogChannelInterface log, Supplier<Trans> fallbackSupplier ) throws KettleException;
```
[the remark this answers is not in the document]
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/trans/TransMananger.java
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/trans/TransMananger.java
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/job/entries/trans/JobEntryTrans.java
soagarwal1: addressed

ui/src/main/java/org/pentaho/di/ui/spoon/trans/TransGraph.java
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
```diff
@@ -0,0 +1,59 @@
...
+package org.pentaho.di.trans;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * The DefaultTransManager is used to manage the registration of TransFactory and
+ * generate the Trans Object for respective RunConfiguration Type
+ */
+public class DefaultTransManager implements TransMananger {
```
NJtwentyone, line 22: mentioned this in the comment - https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2069538322

"TransManager would be more aptly named TransFactoryManger since you have getTransFactory. (follows with other pentaho similar factories)"

I would suggest to rename is to DefaultTransManager --> DefaultTransFactoryManager

- inline with the new code, you are only registering and return TransFactory and not creating any Trans objects
- inline with same functionality as plugins/engine-configuration/api/src/main/java/org/pentaho/di/engine/configuration/api/RunConfigurationProvider**FactoryManager**.java

engine/src/main/java/org/pentaho/di/trans/TransFactoryImpl.java
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Creates a Trans Object for the registered Factory type
+   * @param transMeta
+   * @param repository
+   * @return Trans
+   */
+  @Override
+  public Trans create(TransMeta transMeta, Repository repository )
+    throws KettleException {
+    return new Trans( transMeta, repository, transMeta.getName(), transMeta.getRepositoryDirectory().getPath(),
```
NJtwentyone, line 30: you aren't checking for null object arguments before calling functions on them ie `transMeta.getName()` could be null.

There are several options
[option 1] 
So you can either check for null objects here in this class #`createTrans` and throw an exceptions with an appropriate error message.

- transMeta
- transMeta.getRepositoryDirectory()

with option1 you are forcing a constraint that the other constructor values have to come from transMeta which the Trans constructor does not enforce since it has a constructor with different arguments. 

[options 2] 
Or you can defer to the Trans class and pass in arguments "as is" and let Trans logic handle null pointers and not have to worry about checking arguments since it would be a pass through. So the create(...) would mirror the Trans constructors noted in https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2069538322

I would suggest to go with option 2

engine/src/main/java/org/pentaho/di/trans/DefaultTransManager.java
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/trans/TransFactoryImpl.java
soagarwal1: addressed

NJtwentyone (changes requested):
A couple of major issues and some others.
- compile error/syntax
- not calling type for registering factory

engine/src/main/java/org/pentaho/di/trans/TransFactoryImpl.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Use of this software is governed by the Business Source License included
+ * in the LICENSE.TXT file.
+ *
+ * Change Date: 2029-07-20
+ ******************************************************************************/
+
+package org.pentaho.di.trans;
+
+import org.pentaho.di.core.exception.KettleException;
```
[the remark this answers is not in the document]
soagarwal1: This class will create a Trans object which is the Parent and the TransWebSocket object is the child class which is in spark-execution plugin, we cannot use that class in our test.

I dont see a point of testing for Trans object being created, because even for Spark flow, Trans object will be created as it is the parent.

engine/src/main/java/org/pentaho/di/trans/TransFactoryImpl.java
```diff
@@ -0,0 +1,45 @@
...
+import org.pentaho.di.core.logging.LoggingObjectInterface;
+import org.pentaho.di.repository.Repository;
+
+public class TransFactoryImpl implements TransFactory {
```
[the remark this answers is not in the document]
soagarwal1: addressed

ui/src/main/java/org/pentaho/di/ui/spoon/trans/TransGraph.java
```diff
@@ -3861,8 +3861,8 @@ public synchronized void start( TransExecutionConfiguration executionConfigurati
           // memory
           // To be able to completely test this, we need to run it as we would normally do in pan
           //
-          trans = transMeta.getTransSupplier().get( transMeta, log, this::createLegacyTrans );
-
+          trans = DefaultTransFactoryManager.getInstance().getTransFactory( executionConfiguration.getRunConfiguration() )
```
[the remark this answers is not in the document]
soagarwal1: addressed, might have been removed while doing the commit

engine/src/test/java/org/pentaho/di/trans/DefaultTransFactoryManagerTest.java
```diff
@@ -0,0 +1,32 @@
...
+    TransFactory transFactory = DefaultTransFactoryManager.getInstance().getTransFactory( "random" );
+    Assert.assertTrue( transFactory instanceof TransFactoryImpl );
+  }
+
+  @Test
+  public void registerFactoryTest(){
+    DefaultTransFactoryManager.getInstance().registerFactory( "TestDefault", new TransFactoryImpl() );
+    TransFactory transFactory = DefaultTransFactoryManager.getInstance().getTransFactory( "TestDefault" );
+    Assert.assertTrue( transFactory instanceof TransFactoryImpl );
+  }
```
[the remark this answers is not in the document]
soagarwal1: addressed

ui/src/main/java/org/pentaho/di/ui/spoon/trans/TransGraph.java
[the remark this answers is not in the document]
soagarwal1: addressed as per slack comments, made changes to param name and javadoc, changed it to runConfigurationName.
https://hitachivantara-eng.slack.com/archives/C01HJ874L8Z/p1746513066182599?thread_ts=1746453062.791679&cid=C01HJ874L8Z

engine/src/main/java/org/pentaho/di/trans/TransFactoryImpl.java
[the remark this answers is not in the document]
soagarwal1: created a unit test class, although the test does not make sense. It will always we true.

ui/src/main/java/org/pentaho/di/ui/spoon/trans/TransGraph.java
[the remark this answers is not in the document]
NJtwentyone: So the interface also needs to be update to reflected change in argument

currently interface

-  TransFactoryMananger#registerFactory( String runConfigurationType, TransFactory transFactory );
-  TransFactoryMananger#getTransFactory( String runConfigurationType );

  
  versus implementing class

-   DefaultTransFactoryManager#registerFactory( String runConfigurationName, TransFactory transFactory ) 
-   DefaultTransFactoryManager#getTransFactory( String runConfigurationName )

NJtwentyone (changes requested):
need to update interface TransFactory with the new arguement name and javadoc comments.
- https://github.com/pentaho/pentaho-kettle/pull/9966/commits/40ecba38cfa758c439a4a86180f798821d8b757f#r2075689128

ui/src/main/java/org/pentaho/di/ui/spoon/trans/TransGraph.java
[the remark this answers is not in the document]
soagarwal1: addressed

engine/src/main/java/org/pentaho/di/trans/DefaultTransFactoryManager.java
```diff
@@ -0,0 +1,59 @@
...
+   * @param runConfigurationName
+   * @param transFactory
+   */
+  @Override
+  public void registerFactory( String runConfigurationName, TransFactory transFactory ) {
+    mapConfigurationNameTransFactory.put( runConfigurationName, transFactory );
+  }
+
+  /**
+   * Fetches a TransFactory or Returns default TransFactoryImpl using the provided Run Configuration Name
```
NJtwentyone, line 51: javadoc comment references old class name  "TransFactoryImpl" instead of new name
You should use @see @link so the javadoc will have a hyperlink or tooltip to the actual class - https://www.baeldung.com/javadoc-see-vs-link#2-displaying-see-and-linktag
and any rename changes or refactor, an IDE will automatically pick it up.

engine/src/main/java/org/pentaho/di/trans/TransFactoryImpl.java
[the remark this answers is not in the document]
NJtwentyone: in the test class your should only be calling `DefaultTransFactory#create` and not using the `DefaultManger#getTransFactory`, we are not testing the class` DefaultTransFactoryManager`

NJtwentyone (changes requested):
unit test case for DefaultTransFactory#create needs a better focus and minor javadoc
- https://github.com/pentaho/pentaho-kettle/pull/9966#discussion_r2077622383

engine/src/main/java/org/pentaho/di/trans/DefaultTransFactoryManager.java
soagarwal1: addressed

engine/src/test/java/org/pentaho/di/trans/DefaultTransFactoryTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.pentaho.di.core.logging.LoggingObjectInterface;
+import org.pentaho.di.core.util.Assert;
+import org.pentaho.di.core.variables.Variables;
+
+public class DefaultTransFactoryTest {
+
+
+  @Test
+  public void testCreate() throws KettleException {
+    TransMeta meta =
```
NJtwentyone, line 27: Try using a mock next time, so complex class constructors won't fail your test.

```TransMeta meta = Mockito.mock(TransMeta.class)```

This will work, but next time consider using a mock due to out TransMeta Class might not be a simple class. And if the Trans or any other object has actually complex logic during constructor that require configuration like filesystem configuration or IO then your test could fail.

engine/src/main/java/org/pentaho/di/trans/TransFactoryImpl.java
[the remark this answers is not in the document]
NJtwentyone: test class was updated and addressed.

engine/src/test/java/org/pentaho/di/trans/DefaultTransFactoryTest.java
soagarwal1: I initially used mock. Then it started failing on a few other items used in the method call. And had to declare a lot of other mocks for that. I found this example used in on of the other test cases and used it in a similar way

NJtwentyone (approved):
My approval is conditional on that commits get cleaned up @singletonc @soagarwal1  before merge.

There are PR commit messages such as

- "Addressing PR Comments [to be squashed]:"
- "PR Review: "

PR #9998, opened 2025-05-27
[BACKLOG-42974] Double-click on the VFSConnection or Run Configuratio…

pdesai16:
…n will open New Run Configurations or VFS Connections Dialog

### DIFF
--- a/ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
+++ b/ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
@@ -89,6 +89,7 @@ import org.pentaho.di.cluster.ClusterSchemaManagementInterface;
 import org.pentaho.di.cluster.SlaveServer;
 import org.pentaho.di.cluster.SlaveServerManagementInterface;
 import org.pentaho.di.connections.ConnectionManager;
+import org.pentaho.di.connections.vfs.VFSConnectionDetails;
 import org.pentaho.di.core.AddUndoPositionInterface;
 import org.pentaho.di.core.Const;
 import org.pentaho.di.core.DBCache;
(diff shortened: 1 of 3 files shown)


ui/src/main/java/org/pentaho/di/ui/spoon/Spoon.java
```diff
@@ -7038,6 +7042,18 @@ private List<PartitionSchema> pickupPartitionSchemas( TransMeta transMeta ) thro
     }
   }
 
+  /**
+   * Handles the double click on the VFS connection and RunConfiguration on the tree
+   * @param selection
+   */
+  void createNewSelectionTreeExtension( Object selection ) {
```
bconlon-hv, line 7049: Are we guaranteed that the selection is always going to be either a run configuration or vfs connection?  What if it's not? Feels like we should rename the method to reflect that and do some type checking on that incoming selection object.

abryant-hv: Spoon does not know what the object would be. Each TreeProvider can add their own object types. Hadoop, for example, would add its own type.

bconlon-hv: Ok, so in that case, the javadoc and especially the error message that we throw shouldn't specifically mention VFS connections and run configurations.  The method name is good.

pdesai16: We can't do the type checking on the selection object as the implementation class for VFSConnection and RunConfiguration are in the plugins and we don't want to create a dependency between the Spoon's ui package and plugins. 
The type checking is done in the plugin implementation of the extension point

bconlon-hv (approved):

rmansoor (commented):
ok

PR #10017, opened 2025-06-16
[BACKLOG-44134] - Fix Avro Step wiring/registration issue

soagarwal1:


NJtwentyone:
Suggest to close the PR, the current code changes were apart of a POC and is not implemented correctly. Any extension of ClassLoader should properly implement  loading of classes and resources.


The discussion around Kettle using a parent-first delegation will be captured in:
https://hv-eng.atlassian.net/browse/BACKLOG-44472

A think a better approach would be to just apply as a temp workaround
`@ParentFirst( patterns = { ".*" } ) `
to the various plugins that are immediate impacted like avro, parquet etc...
This will essentially give you parent-first delegation.

[https://github.com/pentaho/pentaho-cassandra-plugin/blob/000d0330652f06944d1373fad[…]ho/di/trans/steps/cassandrasstableoutput/SSTableOutputMeta.java](https://github.com/pentaho/pentaho-cassandra-plugin/blob/000d0330652f06944d1373fad0f5a2a58ca9422a/core/src/main/java/org/pentaho/di/trans/steps/cassandrasstableoutput/SSTableOutputMeta.java#L58)

nawaz34-hitachivantara (approved):

